PR #124, opened 2026-04-11
fix(bot): harden against prompt injection via file manipulation

camlloyd:
## Problem

The bot is vulnerable to persistent prompt injection through two vectors:

1. **Log token exposure**: PTB's `_get_encoded_url()` percent-encodes
   `:` → `%3A` in file download URLs, so the bot token appeared as
   `bot123456%3AABC` in httpx GET logs — bypassing `_TokenRedactFilter`
   which only checked for the raw token.

2. **Protected file overwrite**: The `write_file` and `save_file` tools
   had no restrictions on which files could be written. A user could
   trick the LLM into overwriting `SOUL.md`, `CLAUDE.md`,
   `requirements.txt`, or bot scripts — achieving persistent system
   prompt poisoning or code execution.

## Fixes

**Token redaction** (`_TokenRedactFilter`):
- Pre-computes `urllib.parse.quote(token, safe="")` and applies both
  raw and percent-encoded replacements unconditionally on every log
  record. Removes the `in` guard (str.replace is a no-op on no match).
- Moves `urllib.parse` to top-level imports.

**Protected file writes** (`execute_write_file` + `execute_save_file`):
- Module-level `_PROTECTED_NAMES` frozenset shared by both tools,
  checked case-insensitively after filename sanitization.
- `write_file` destination clamped to `DATA_DIR` — structural allowlist
  beats any blocklist alone.
- `save_file` now has the same protected-name check as `write_file`.
- `_sanitize_filename` strips leading/trailing whitespace to close
  trailing-space bypass on case-insensitive filesystems (macOS HFS+).

**System prompt**:
- `ROLE_GUARDRAILS` explicitly forbids reading or modifying system
  config files even if the user claims to be an administrator. Note:
  this is model-dependent — the tool layer provides the hard guarantee.

### DIFF
--- a/bot/roboterri.py
+++ b/bot/roboterri.py
@@ -27,6 +27,7 @@ import shutil
 import sys
 import tempfile
 import time
+import urllib.parse
 from datetime import datetime
 from pathlib import Path
 
@@ -93,8 +94,7 @@ class _TokenRedactFilter(logging.Filter):
         self._token = token
         # PTB's _get_encoded_url() percent-encodes ':' → '%3A' in download
         # URLs, so store both forms and always replace both.
-        import urllib.parse as _up
-        self._token_encoded = _up.quote(token, safe="")
+        self._token_encoded = urllib.parse.quote(token, safe="")
 
     def filter(self, record: logging.LogRecord) -> bool:
         try:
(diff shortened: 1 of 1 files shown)


manuelcorpas (approved):
Audited by ClawBio pr-audit. SAFE.

Excellent security hardening:
- `_PROTECTED_NAMES` frozenset blocks writes to SOUL.md, CLAUDE.md, .env, and other config files
- Token redaction now covers percent-encoded form too
- SYSTEM FILE POLICY added to ROLE_GUARDRAILS
- `execute_write_file` destination clamped to DATA_DIR
- Audit logging on blocked attempts

Minor nit: lines 79/96 use f-string in `logger.warning()` — `%s` formatting is marginally safer practice for logging, but not blocking since `filename` is already sanitised.

PR #146, opened 2026-04-21
fix(repro): portable reproducibility bundles — closes #143

danilomonge:
## What this PR does

This PR introduces a **shared reproducibility infrastructure** for ClawBio and integrates it across multiple skills.

After a skill completes, updated skills now write a **reproducibility bundle** into:

    <output_dir>/reproducibility/

This bundle provides a consistent way to re-run analyses and verify their outputs, improving reproducibility across runs and environments.

The PR:

- Establishes a **central, reusable API** for reproducibility
- Standardizes how replay commands are generated
- Integrates the system across multiple skills
- Improves reliability of output verification via checksums
- Fixes issues discovered during end-to-end validation

---

## How it works

After execution, updated skills generate a reproducibility bundle containing:

| File | Purpose |
|------|--------|
| commands.sh | Script to re-run the analysis |
| environment.yml | Conda-compatible environment definition |
| checksums.sha256 | Hashes for verifying deterministic outputs |

---

### Replay script (commands.sh)

The generated script reconstructs the original command in a structured and consistent way.

It defines:

    SCRIPT_DIR=...
    OUTPUT_DIR=...
    CLAWBIO_ROOT=...

and executes the corresponding skill using these variables.

Paths inside the command are rendered relative to logical anchors:

- Repository paths → $CLAWBIO_ROOT/...
- Output paths → $OUTPUT_DIR/...

This avoids embedding hardcoded paths directly into the command and improves portability.

---

### Structured command generation

This PR replaces manual string-based command construction with a structured approach using:

- ReproPath → represents a path with context (repo, output, or external)
- ReproCommand → represents the command and its arguments

This ensures:

- consistent command reconstruction
- reduced duplication across skills
- easier extension and maintenance

---

### Checksums

The PR standardizes output verification:

- Only deterministic outputs are fingerprinted
- Files that may vary across runs (e.g. reports, logs, metadata) are excluded

This significantly improves the reliability of checksum validation when re-running analyses.

---

## New shared API — clawbio/common/reproducibility.py

| Symbol | Description |
|--------|------------|
| ReproPath | Encapsulates a path and its reproducibility context |
| ReproCommand | Structured representation of a replay command |
| write_portable_commands_sh | Generates the replay script |
| write_environment_yml | Writes environment specification |
| write_checksums | Writes SHA-256 hashes for outputs |

---

## Skills updated

The new reproducibility system has been integrated into multiple skills, including:

[...]

PR #167, opened 2026-04-22
Fix bio-orchestrator routing for pure scRNA embedding queries

### LINKED ISSUE #165: `bio-orchestrator` misroutes "latent embedding" queries to `scrna-orchestrator` instead of `scrna-embedding` due to longest-match keyword precedence
## Summary

`detect_skill_with_hint_from_query()` in `skills/bio-orchestrator/orchestrator.py` uses a longest-match strategy over `KEYWORD_MAP` as its final routing stage. For the query `Build a latent embedding for this single-cell dataset`, the keyword `single-cell` (11 characters, mapped to `scrna-orchestrator`) has a longer character count than `embedding` (9) or `latent` (6), both of which are mapped to `scrna-embedding`. The longest-match rule therefore routes the query to `scrna-orchestrator`, contradicting the semantic intent of the query, the design of the early-return guards, and the expectation of the existing regression test.

## Evidence

`skills/bio-orchestrator/orchestrator.py`, `KEYWORD_MAP` (lines approximately 68-93):

```python
"latent":       "scrna-embedding",    # 6 chars
"embedding":    "scrna-embedding",    # 9 chars
...
"single-cell":  "scrna-orchestrator", # 11 chars <- wins by character count
```

Routing function (`detect_skill_with_hint_from_query`, lines 321-345):

```python
wants_embedding = any(term in query_lower for term in SCRNA_EMBEDDING_TERMS)
wants_downstream = any(term in query_lower for term in SCRNA_DOWNSTREAM_TERMS)
has_latent_artifact = any(term in query_lower for term in SCRNA_LATENT_ARTIFACT_TERMS)

if has_latent_artifact and wants_downstream:
    return ("scrna-orchestrator", ...)
if wants_embedding and wants_downstream:
    return ("scrna-embedding", ...)
# Falls through to longest-match KEYWORD_MAP scan
best_skill, best_len = None, 0
for keyword, skill in KEYWORD_MAP.items():
    if keyword in query_lower and len(keyword) > best_len:
        best_skill = skill
        best_len = len(keyword)
```

For `build a latent embedding for this single-cell dataset`:

| Signal | Value |
| --- | --- |
| `wants_embedding` | `True` (`latent` and `embedding` match) |
| `wants_downstream` | `False` (no `marker`, `cluster`, `annotation`, etc.) |
| `has_latent_artifact` | `False` (no `integrated h5ad`, `after scvi`) |

Neither early-return guard fires. The longest-match scan selects `single-cell` (11 chars) -> `scrna-orchestrator`.

Failing test (`skills/scrna-embedding/tests/test_scrna_embedding.py`, line 649):

```python
assert module.detect_skill_from_query(
    "Build a latent embedding for this single-cell dataset"
) == "scrna-embedding"
# AssertionError: 'scrna-orchestrator' != 'scrna-embedding'
```

## Root Cause

The early-return guards correctly encode the semantic intent: `wants_embedding` and not `wants_downstream` should route to `scrna-embedding`. However, the code structure requires both `wants_embedding` and `wants_downstream` to be `True` to activate the `scrna-embedding` early return. When only `wants_embedding` is `True`, the pure embedding case falls through to the keyword map, where the generic `single-cell` token outcompetes the more specific `latent` and `embedding` tokens by raw length.

## Proposed Fix

Add a third early-return guard for the pure embedding case:

```python
if wants_embedding and not wants_downstream and not has_latent_artifact:
    return (
        "scrna-embedding",
        "Detected an embedding-focused scRNA workflow. "
        "Run `scrna-embedding` to produce `integrated.h5ad` with a "
        "scVI/scANVI latent space. For downstream clustering and annotation, "
        "run `scrna-orchestrator` with `--use-rep X_scvi` afterwards.",
    )
```

This fires before the `KEYWORD_MAP` scan and correctly handles queries that express a clear embedding intent without downstream signals.

Additionally, extend the routing test suite with ambiguous cases to prevent regressions, for example queries containing `single-cell` plus `latent` but no downstream terms, and queries containing `single-cell` alone.

danilomonge:
## Problem

Issue #165 found that `detect_skill_with_hint_from_query()` in the bio-orchestrator was misrouting pure embedding queries (e.g. `"Build a latent embedding for this single-cell dataset"`) to `scrna-orchestrator` instead of `scrna-embedding`.

The root cause is a collision between the longest-match keyword strategy in `KEYWORD_MAP` and the early-return guard structure: the keyword `single-cell` (11 chars) outcompeted both `embedding` (9 chars) and `latent` (6 chars) by raw character length, even though those shorter keywords express the user's actual intent.

The early-return guards encode the correct semantics, but the pure-embedding case (`wants_embedding = True`, `wants_downstream = False`) was missing a guard of its own. Only the mixed case (`wants_embedding AND wants_downstream`) had an early return, so pure embedding queries fell through to the keyword scan and were misrouted.

## Fix

- Add a third early-return guard: if `wants_embedding` is true and neither `wants_downstream` nor `has_latent_artifact` are set, route directly to `scrna-embedding` before the `KEYWORD_MAP` scan runs.
- Preserve existing downstream latent-analysis routing to `scrna-orchestrator` (latent artifact + downstream signals).
- Add regression tests for pure embedding queries and generic `single-cell` queries to prevent future longest-match surprises.

## Testing

- `pytest skills/bio-orchestrator/tests/test_orchestrator.py -q -k 'DetectSkillWithHint or detect_skill_from_query'`
- `pytest skills/scrna-embedding/tests/test_scrna_embedding.py -q`

Closes #165

### DIFF
--- a/skills/bio-orchestrator/orchestrator.py
+++ b/skills/bio-orchestrator/orchestrator.py
@@ -331,13 +331,23 @@ def detect_skill_with_hint_from_query(query: str) -> tuple[str | None, str]:
             "with `--use-rep X_scvi` on `integrated.h5ad` to run clustering, annotation, "
             "and contrastive markers after scVI.",
         )
-    if wants_embedding and wants_downstream:
-        return (
-            "scrna-embedding",
-            "Detected a two-step advanced scRNA workflow. First run `scrna-embedding` to "
-            "produce `integrated.h5ad`, then run `scrna-orchestrator` with "
-            "`--use-rep X_scvi` for downstream clustering, annotation, and contrastive markers.",
-        )
+
+    if wants_embedding:
+        if wants_downstream:
+            return (
+                "scrna-embedding",
+                "Detected a two-step advanced scRNA workflow. First run `scrna-embedding` to "
+                "produce `integrated.h5ad`, then run `scrna-orchestrator` with "
+                "`--use-rep X_scvi` for downstream clustering, annotation, and contrastive markers.",
+            )
+        if not has_latent_artifact:
+            return (
+                "scrna-embedding",
+                "Detected an embedding-focused scRNA workflow. Use `scrna-embedding` to "
+                "produce `integrated.h5ad` with a scVI/scANVI latent space before "
+                "running downstream clustering or annotation.",
+            )
+
     # Prefer longest keyword match to avoid ambiguity (e.g. "variant annotation"
     # should match vcf-annotator, not equity-scorer via "variant" substring)
     best_skill = None
--- a/skills/bio-orchestrator/orchestrator.py
+++ b/skills/bio-orchestrator/orchestrator.py
@@ -324,6 +324,8 @@ def detect_skill_with_hint_from_query(query: str) -> tuple[str | None, str]:
     wants_downstream = any(term in query_lower for term in SCRNA_DOWNSTREAM_TERMS)
     has_latent_artifact = any(term in query_lower for term in SCRNA_LATENT_ARTIFACT_TERMS)
 
+    # Chain-aware scRNA routing favors explicit embedding requests unless the
+    # user is clearly asking for downstream analysis on an existing latent artifact.
     if has_latent_artifact and wants_downstream:
         return (
             "scrna-orchestrator",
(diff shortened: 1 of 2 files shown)


danilomonge:
Incorporated a follow-up to tighten the routing change:
- handle explicit embedding queries that mention `integrated.h5ad` by keeping them on `scrna-embedding` when the user is not asking for downstream analysis
- flatten the scRNA routing precedence so the chain-aware cases are easier to read and maintain
- add regression coverage for the latent-artifact embedding case and a downstream-only orchestrator case
- strengthen the new hint assertions so they check the intended phrases directly

Validation:
- `pytest -q skills/bio-orchestrator/tests/test_orchestrator.py -k "DetectSkillWithHint or detect_skill_from_query"`
- `pytest -q skills/scrna-embedding/tests/test_scrna_embedding.py -k "orchestrator_routes_embedding_keywords_to_new_skill"`

manuelcorpas (approved):

Adds two early-return guards in `detect_skill_with_hint_from_query` for pure scRNA embedding queries, returning the `scrna-embedding` skill string with descriptive hints. No shell calls, no subprocess, no safety check bypass; pure routing logic. New tests cover both branches.

Note: PR #146 contains a divergent variant of this same routing change. Whichever merges last wins; recommend merging #167 first then asking #146 to rebase.

PR #182, opened 2026-04-22
fix(cli): replace flat skill-flag registry with parse_known_args passthrough

### LINKED ISSUE #155: clawbio.py crashes at startup due to duplicate --search argument registration
## Summary

The main CLI entrypoint (`clawbio.py`) registers `--search` twice on the same `argparse` subparser (`run_parser`). Python's `argparse` does not silently ignore duplicate long options: it raises `argparse.ArgumentError` when the parser is constructed, which happens before any command is parsed. This makes every invocation of `clawbio.py` fail unconditionally.

## Affected Version

Confirmed on HEAD (`ea1f8ca`, 22 Apr 2026).

## Evidence

`clawbio.py`, line 1176:

```python
run_parser.add_argument("--search", default=None,
    help="Live Bioconductor package search query for bioc skill")
```

`clawbio.py`, line 1190:

```python
run_parser.add_argument("--search", default=None,
    help="Search query (galaxy/flow skills)")
```

Both calls target the same `run_parser` object. `argparse` raises on the second registration:

```text
argparse.ArgumentError: argument --search: conflicting option string(s): --search
```

## Minimal Reproduction

```bash
git clone https://github.com/ClawBio/ClawBio.git && cd ClawBio
python clawbio.py list
python clawbio.py upload --help
python clawbio.py run pharmgx --demo
```

All three commands crash before doing useful work.

## Downstream Failures

Any test that invokes `clawbio.py` as a subprocess will receive a non-zero exit code caused solely by this parser error, masking the actual skill behaviour.

Known examples include subprocess-based smoke tests in:

- `bioconductor-bridge`
- `bigquery-public`
- `diff-visualizer`
- `flow-bio`
- `illumina-bridge`

## Expected Behaviour

The parser constructs without errors; `clawbio.py list`, `upload --help`, and `run <skill>` all execute normally.

## Proposed Fix

Rename one of the two flags to a skill-specific name, for example `--bioc-search` for the Bioconductor flag and keep `--search` for the flow/galaxy flag, or vice versa, and update the corresponding skill dispatch code.

Alternatively, move skill-specific flags into their own subparsers or a shared flag namespace.

A regression test at the subprocess level (`python clawbio.py list` -> exit code `0`) should be added to prevent recurrence.

## Severity

Critical. The documented quickstart (`python clawbio.py run pharmgx --demo`) is broken on every supported platform as of HEAD.

danilomonge:
## Problem

Issue #155 identified a critical startup crash: `clawbio.py` registered `--search` twice on the same `run_parser` — once for the Bioconductor skill and once for the flow/galaxy skills. Python's `argparse` raises `ArgumentError` when a duplicate flag is registered, which happens at parser construction time — before any command is parsed. Every invocation of the CLI (`list`, `run`, `upload --help`) failed unconditionally.

The root cause is architectural: `clawbio.py run <skill>` is a subprocess wrapper. A wrapper must not pre-register the internal flags of the programs it wraps. This fix corrects that design.

### `clawbio.py`
- Replace `parser.parse_args()` with `parser.parse_known_args()`. Unrecognised flags are captured in `extra` and forwarded verbatim to the skill subprocess.
- Remove all ~60 skill-specific `add_argument` registrations from `run_parser`.
- Remove the ~220-line manual dispatch block that reconstructed `extra_args` from individual `getattr` calls.
- Replace defensive `getattr(args, "profile_path", None)` with direct `args.profile_path` (the attribute is unconditionally registered on `run_parser`).
- **Net diff: −435 lines, +2 lines.** Only six flags remain on `run_parser`: `skill`, `--demo`, `--input`, `--output`, `--profile`, `--timeout`.

### `skills/flow-bio/flow_bio.py`
- The old dispatch layer silently renamed `--pipeline-detail → --pipeline`, `--sample-detail → --sample`, and `--execution-detail → --execution` before forwarding. With verbatim passthrough, those renames disappear.
- Add the `*-detail` forms as argparse aliases directly on the skill parser (e.g. `add_argument("--pipeline", "--pipeline-detail", ...)`), consistent with the documented public interface in `SKILL.md` (lines 141–143).
- The skill is the correct owner of its flag interface; the wrapper is not.

### `tests/test_cli_smoke.py` *(new file)*
- Four regression tests guarding parser construction, `--help` on all subcommands, and unknown-flag passthrough.

## Impact

| | Before | After |
|---|---|---|
| Flags registered in wrapper | ~60 | 6 |
| Startup crash on duplicate flags | Yes | Impossible by design |
| New skills may reuse any flag name | No | Yes |
| Regression tests | 0 | 4 |

The entire class of "duplicate flag" bugs is eliminated: since `clawbio.py` no longer declares skill flags, no future skill addition can trigger a conflict at the wrapper level.

## Testing

```bash
python -m pytest tests/test_cli_smoke.py       # 4/4 pass
python -m pytest skills/flow-bio/tests/        # 25/25 pass
python clawbio.py list                         # exits 0
python clawbio.py run flow-bio --pipeline-detail <id>   # alias works
python clawbio.py run flow-bio --pipeline <id>          # short form works
```

Closes #155


manuelcorpas (approved):

`parse_known_args()` collects unknown flags into `extra` and forwards verbatim to the skill subprocess via the existing list-args invocation. No `shell=True` introduced; the existing argv-list pattern is preserved, so CLI passthrough does not open a shell injection vector. Net -434 LOC in `clawbio.py` reduces the central flag-registry attack surface. Aliases added to flow-bio's own argparse (correct owner). New smoke tests included.

PR #185, opened 2026-04-23
fix(clinical-trial-finder): add sys.path guard before bare local imports

### LINKED ISSUE #158: `clinical-trial-finder` performs a bare module-level `from api import` that breaks in any multi-skill Python process
## Summary

`skills/clinical-trial-finder/clinical_trial_finder.py` imports its API layer via an unqualified, module-level `from api import ...`. Because eight other skills also expose a file named `api.py` at their skill root, this import resolves non-deterministically depending on `sys.path` ordering. In any process where another skill's directory is on `sys.path` before `clinical-trial-finder/`, the import silently resolves to the wrong `api` module, raising `ImportError` or producing incorrect behaviour at runtime.

## Evidence

`skills/clinical-trial-finder/clinical_trial_finder.py`, line 14:

```python
from api import fetch_trials, parse_input
```

This is a top-level module import, not deferred and not inside a function or `if __name__ == "__main__"` block.

Other skill roots that also contain `api.py`:

- `skills/nutrigx_advisor/api.py`
- `skills/hla-typing/api.py`
- `skills/gwas-prs/api.py`
- `skills/pharmgx-reporter/api.py`
- `skills/clinpgx/api.py`
- `skills/data-extractor/api.py`
- `skills/genome-compare/api.py`
- `skills/equity-scorer/api.py`

`clinical-trial-finder/api.py` exports `fetch_trials` and `parse_input`. None of the other `api.py` files export these symbols. If any other skill directory precedes `skills/clinical-trial-finder` on `sys.path`, the import will fail with:

```text
ImportError: cannot import name 'fetch_trials' from 'api'
```

Or, if the other `api.py` happens to export a symbol of the same name with different semantics, the failure will be silent.

## Why This Is Structural

The `clinical-trial-finder` skill docstring correctly documents that `api.py` is the API client layer. The problem is not its content but that it is referenced by an unqualified name that collides with every other skill that follows the same `api.py` naming convention.

## Proposed Fix

Replace the bare import with a qualified relative import:

```python
# Before
from api import fetch_trials, parse_input

# After (if skill is a package with __init__.py)
from .api import fetch_trials, parse_input

# Or with explicit path manipulation already present in other skills
from clinical_trial_finder_api import fetch_trials, parse_input
```

Alternatively, rename `api.py` to `ctg_client.py` (ClinicalTrials.gov client) to both clarify intent and eliminate the collision.

Update `skills/clinical-trial-finder/tests/` imports accordingly and add a test that imports `clinical_trial_finder` after having another skill's directory on `sys.path` to confirm deterministic resolution.

danilomonge:
## Problem

Issue #158 identified that `clinical_trial_finder.py` imported its three local modules (`api`, `constants`, `writers`) via bare, unqualified names with no `sys.path` guard. Eight other skills also expose a file named `api.py` at their skill root. In any multi-skill Python process — such as a `pytest` run from the repository root — `sys.path` ordering is non-deterministic, so `from api import fetch_trials, parse_input` could resolve to the wrong `api.py`, raising `ImportError` or silently binding the wrong symbols.

## Change

`skills/clinical-trial-finder/clinical_trial_finder.py` — 5 lines added between the stdlib imports and the local imports:

```python
import sys

_SKILL_DIR = Path(__file__).resolve().parent
if str(_SKILL_DIR) not in sys.path:
    sys.path.insert(0, str(_SKILL_DIR))
```

This guarantees that `api`, `constants`, and `writers` always resolve to the skill's own modules regardless of `sys.path` ordering in the calling process. The guard prevents duplicate insertion on repeated imports.

Follows the pattern established in `skills/clinical-variant-reporter/clinical_variant_reporter.py`.

## Testing

```bash
# Full test suite
pytest skills/clinical-trial-finder/tests/test_clinical_trial_finder.py  # 141/141 pass

# Direct CLI
python skills/clinical-trial-finder/clinical_trial_finder.py --help  # exits 0

# Hostile sys.path — three other skills' api.py inserted before clinical-trial-finder
# fetch_trials still resolves to clinical-trial-finder/api.py ✓

# Guard idempotency — importing the module 3 times inserts _SKILL_DIR exactly once ✓
```

Closes #158

### DIFF
--- a/skills/clinical-trial-finder/clinical_trial_finder.py
+++ b/skills/clinical-trial-finder/clinical_trial_finder.py
@@ -9,7 +9,12 @@ Entry point and CLI only.  Business logic lives in:
 """
 
 import argparse
+import sys
 from pathlib import Path
+
+_SKILL_DIR = Path(__file__).resolve().parent
+if str(_SKILL_DIR) not in sys.path:
+    sys.path.insert(0, str(_SKILL_DIR))
 
 from api import fetch_trials, parse_input
 from constants import ALL_STATUSES, DEFAULT_PAGE_SIZE, DEMO_DATA


manuelcorpas (approved):

`_SKILL_DIR = Path(__file__).resolve().parent` and `sys.path.insert(0, str(_SKILL_DIR))` is correctly the skill's own directory; does not reach outside the package. Idempotent guard prevents duplicate insertion. Pattern matches the existing `clinical-variant-reporter` convention.

PR #187, opened 2026-04-23
fix: declare missing skill-level dependencies for affinity-proteomics and variant-annotation

### LINKED ISSUE #159: `statsmodels` is undeclared in `requirements.txt`; `differential_abundance()` in `affinity-proteomics` fails silently in the base environment
## Summary

The `affinity-proteomics` skill's `differential_abundance()` function uses `statsmodels` for multiple-testing correction (Benjamini-Hochberg). This dependency is not listed in the repository's `requirements.txt`, nor in any skill-level requirements file. In the base environment, calling `differential_abundance()` raises `ModuleNotFoundError` at the point of invocation, silently disabling a core part of the analysis pipeline.

## Evidence

`skills/affinity-proteomics/affinity_proteomics.py`, line 238:

```python
def differential_abundance(
    data: ProteomicsData,
    group_col: str,
    contrast: tuple[str, str],
    ...
) -> list[DiffAbundanceResult]:
    from scipy import stats
    from statsmodels.stats.multitest import multipletests
```

The import is correctly deferred inside the function body, so the module itself imports without error. However, any code path that calls `differential_abundance()`, including the primary demo pipeline, fails with:

```text
ModuleNotFoundError: No module named 'statsmodels'
```

`requirements.txt`: `statsmodels` is absent.

`skills/affinity-proteomics/requirements.txt`: does not exist.

## Impact

`differential_abundance()` is invoked by `run_pipeline()`, which is the canonical entry point for both the `--demo` mode and real Olink/SomaScan analyses. Any end-to-end test or user workflow that reaches this function will fail in a freshly installed environment.

`scipy` (also used in the same function) is similarly absent from `requirements.txt`, though it is a transitive dependency of other packages and may be present incidentally.

## Proposed Fix

1. Add `statsmodels>=0.14` to `requirements.txt`, or to `skills/affinity-proteomics/requirements.txt` if per-skill dependency files are adopted.
2. If `statsmodels` is intentionally optional for lightweight deployments, replace the bare deferred import with a conditional and a clear user-facing error:

```python
try:
    from statsmodels.stats.multitest import multipletests
except ImportError as exc:
    raise ImportError(
        "statsmodels is required for multiple-testing correction. "
        "Install it with: pip install statsmodels"
    ) from exc
```

3. Add a test that calls `differential_abundance()` end-to-end under the documented environment to prevent silent regressions.

### LINKED ISSUE #160: `pysam` is undeclared in `requirements.txt`; VCF parsing in `variant-annotation` is unavailable in the base environment
## Summary

The `variant-annotation` skill requires `pysam` for VCF parsing. The dependency is correctly guarded at import time, so the module loads without error when `pysam` is absent, but `pysam` is not listed in `requirements.txt` or any skill-level requirements file. In a clean install, every code path that invokes `parse_vcf()` raises `ImportError`, making the primary use case of the skill non-functional.

## Evidence

`skills/variant-annotation/variant_annotation.py`, lines 21-24:

```python
try:
    pysam = importlib.import_module("pysam")
except ImportError:  # pragma: no cover - handled at runtime in main()
    pysam = None
```

Lines 271-278 (`parse_vcf`):

```python
def parse_vcf(input_path: Path) -> list[dict[str, Any]]:
    """Parse a single-sample or multi-sample VCF via pysam."""
    if pysam is None:
        raise ImportError(
            "pysam is required for VCF parsing. "
            "Install it with `pip install pysam`."
        )
    with pysam.VariantFile(str(input_path)) as vcf:
```

The guard pattern is correctly implemented: the module itself is importable, and the error message is actionable. However, because `pysam` is absent from `requirements.txt`, the error surfaces for all users who follow the standard installation instructions (`pip install -r requirements.txt`).

Tests that fail in the base environment (`pysam` absent):

- `test_parse_vcf_reads_demo_variants`
- `test_batch_variants_and_region_string`
- `test_run_pipeline_writes_expected_files`

`requirements.txt`: `pysam` is absent.

`skills/variant-annotation/requirements.txt`: does not exist.

## Note on `pysam` Platform Availability

`pysam` is not available on Windows and requires a C build toolchain. If Windows support is a goal, a pure-Python fallback such as `cyvcf2` or manual VCF parsing should be considered alongside the dependency declaration.

## Proposed Fix

1. Add `pysam>=0.22` to `requirements.txt` or to a skill-level `skills/variant-annotation/requirements.txt` with a note about platform constraints.
2. If `pysam` is treated as an optional extra, document the installation step explicitly in `skills/variant-annotation/SKILL.md` and use `pytest.importorskip("pysam")` in affected tests so they skip cleanly rather than fail on unsupported platforms.

danilomonge:
## Problem

Issues #159 and #160 found that two skills had undeclared runtime dependencies. Both skills already handle the missing-dependency case correctly in code — imports are deferred and guarded — but without declared requirements, users following standard install instructions (`pip install -r requirements.txt`) would hit `ModuleNotFoundError` at runtime with no guidance on how to fix it.

### `skills/affinity-proteomics/requirements.txt` *(new file)*

`differential_abundance()` imports `statsmodels` and `scipy` at call time (correctly deferred inside the function body). Neither was declared in any requirements file. Both are declared together since they are used together in the same function.

```
statsmodels>=0.14
scipy>=1.10
```

### `skills/variant-annotation/requirements.txt` *(new file)*

`pysam` is already guarded with `try/except ImportError` at module level, and `parse_vcf()` raises a clear `ImportError` with installation instructions when `pysam` is absent. The dependency was simply undeclared. The platform constraint (Linux/macOS only — htslib C extensions) is documented inline.

```
# Linux / macOS only — pysam requires htslib C extensions
pysam>=0.22
```

## Testing

No code was changed. Existing test behaviour is unchanged:

```bash
# affinity-proteomics: 13 passed, 6 skipped (somadata optional) with statsmodels installed
pytest skills/affinity-proteomics/tests/

# variant-annotation: passes with pysam installed; gracefully raises ImportError without it
pytest skills/variant-annotation/tests/
```

Closes #159
Closes #160

### DIFF
--- a/skills/affinity-proteomics/requirements.txt
+++ b/skills/affinity-proteomics/requirements.txt
@@ -1,0 +1,2 @@
+statsmodels>=0.14
+scipy>=1.10
--- a/skills/variant-annotation/requirements.txt
+++ b/skills/variant-annotation/requirements.txt
@@ -1,0 +1,2 @@
+# Linux / macOS only — pysam requires htslib C extensions
+pysam>=0.22


manuelcorpas (approved):

Adds `statsmodels>=0.14`, `scipy>=1.10`, `pysam>=0.22` — all reputable PyPI packages, no typo-squats. Reasonable lower bounds, no upper pins (acceptable for skill-local requirements).

PR #209, opened 2026-04-26
fix: isolate skill-private namespaces to avoid import collisions

### LINKED ISSUE #157: Shared `core` top-level namespace across four skills causes `ModuleNotFoundError` in aggregated test runs
## Summary

Four independent skills each contain a local package named `core/`:

| Skill | Path |
| --- | --- |
| `fine-mapping` | `skills/fine-mapping/core/` |
| `struct-predictor` | `skills/struct-predictor/core/` |
| `data-extractor` | `skills/data-extractor/core/` |
| `gwas-lookup` | `skills/gwas-lookup/core/` |

All four skills add their own directory to `sys.path` at module level and then import from `core` using unqualified absolute imports. Because `sys.modules` is a process-global dictionary, the first skill directory added to `sys.path` that gets registered as `core` in `sys.modules` will shadow all others. Any subsequently imported skill that refers to `core` will receive the wrong package.

## Evidence

`skills/fine-mapping/fine_mapping.py`, lines 47-51:

```python
if str(_SKILL_DIR) not in sys.path:
    sys.path.insert(0, str(_SKILL_DIR))

from core.io import load_sumstats, load_ld
from core.abf import compute_abf
from core.susie import run_susie
from core.credible_sets import build_credible_sets_susie, build_credible_set_abf
from core.report import generate_markdown, write_tables, generate_figures, ...
```

`skills/struct-predictor/struct_predictor.py`, lines 29-32:

```python
if str(_SKILL_DIR) not in sys.path:
    sys.path.insert(0, str(_SKILL_DIR))

from core.io import validate_and_prepare
from core.predict import run_boltz
from core.confidence import extract_confidence
from core.report import generate_report
```

`fine-mapping/core` and `struct-predictor/core` expose entirely different modules (`abf`, `susie`, `credible_sets` vs. `io`, `predict`, `confidence`, `report`). Once one is registered under the key `core` in `sys.modules`, the other raises `ModuleNotFoundError`.

`data-extractor/web/server.py` and `gwas-lookup` tests exhibit the same pattern, multiplying collision surfaces.

## Failure Scenario (Minimal)

```python
import sys
sys.path.insert(0, "skills/struct-predictor")
import struct_predictor  # registers core -> struct-predictor/core

sys.path.insert(0, "skills/fine-mapping")
import fine_mapping
# ModuleNotFoundError: No module named 'core.abf'
# (because sys.modules['core'] already points to struct-predictor/core)
```

## Why `--import-mode=importlib` Does Not Fully Mitigate This

`pytest.ini` sets `--import-mode=importlib`, which isolates test file imports. However, the `sys.path.insert` calls execute at module load time inside the skill scripts themselves, not inside test files. Once any skill script is imported, it registers `core` globally. Isolation is therefore incomplete in multi-skill test sessions and in any production process that runs more than one skill.

## Proposed Fix

Rename each `core/` to a skill-specific package:

- `skills/fine-mapping/fine_mapping_core/`
- `skills/struct-predictor/struct_predictor_core/`
- `skills/data-extractor/data_extractor_core/`
- `skills/gwas-lookup/gwas_lookup_core/`

Or, make each skill a proper Python package with `__init__.py` at its root and use relative imports (`from .core.io import ...`). Either approach eliminates the unqualified global namespace.

Add a cross-skill aggregated test that imports both `fine_mapping` and `struct_predictor` in the same process and asserts that no `ModuleNotFoundError` is raised.

elbotmasbot:
## Summary
- rename skill-private `core` packages so they no longer collide in `sys.modules`
- isolate the remaining top-level `api` collisions between `gwas-lookup` and `data-extractor`
- add a shared regression test for multi-skill imports in the same Python process
- update fine-mapping benchmark imports and stabilize `genome-compare` tests to use `tmp_path`

## Why
Several skills exposed top-level `core` packages and imported them after mutating `sys.path`. In a shared Python process, the first imported skill could capture `sys.modules['core']` and break later imports. While validating the fix, the same pattern also surfaced for top-level `api` modules.

This PR makes those helper modules skill-private and adds regression coverage so multiple skills can be imported safely in one interpreter.

## Verification
- `python -m pytest skills/fine-mapping/tests/test_fine_mapping.py skills/struct-predictor/tests/test_struct_predictor.py skills/gwas-lookup/tests/test_gwas_lookup.py skills/data-extractor/tests/test_extractor.py tests/test_skill_namespace_collisions.py -q`
- `python -m pytest tests/benchmark/test_finemapping_benchmark.py -q`
- `python -m pytest skills/genome-compare/tests/test_genome_compare.py -q`
- `python -m pytest -q`
- manual smoke tests for `fine-mapping`, `gwas-lookup`, and `genome-compare` demos, plus same-process module imports for the affected skills

Fixes #157


manuelcorpas (approved):

Mass rename of skill-internal `core/` and `api/` packages to skill-prefixed names (`fine_mapping_core`, `gwas_lookup_core`, `gwas_lookup_api`, `data_extractor_core`, `data_extractor_api`, `struct_predictor_core`). Pure Python rename plus matching import updates plus a new `tests/test_skill_namespace_collisions.py`. Also stabilises `genome-compare` tests by switching from hardcoded `/tmp/test_*` paths to pytest `tmp_path` fixtures (good hygiene fix). Removes `tests/benchmark` from `pytest.ini` testpaths and adds top-level `tests`. The sys.modules collision rationale is real.

PR #240, opened 2026-05-09
feat: support structured suggested actions

smoe:
Can we discuss a bit if this is wanted or if this should somehow working differently? 

## Summary

Any skill producing some results is likely to have an idea what is special in its findings. And depending on what was found special, a particular action may be desirable to follow up with. ClawBio of course will have extra thoughts on its own, and so will the user, but the expert skill should be granted an option to make a proposal on what should be next. The .json returned shall therefore be structured as

 - chat_summary_lines: concise text written by the skill for chat
 - preferred_artifacts: files the UI should prioritize displaying
 - suggested_actions: deterministic follow-up requests to offer
 - report_md: full markdown report embedded in result.json

This came up for the skill (GENtle) I was preparing. Immediate use could be for the introduction of a new skills, like a table of content with more detailed follow ups for individual topics. It would feel a bit like hyperlinks, I guess.

## Skill affected

None. It is all optional.

## Checklist

- [X] Tests included and passing (`pytest -v`)
- [ ] Demo data included for reviewers to test
- [X] No patient/sensitive data committed
- [X] Follows [CONTRIBUTING.md](../CONTRIBUTING.md) conventions

I kept the diff minimal, but that Demo I should provide, had used it only for GENtle. I'll add a patch for that, still need to make up my mind on what skill to amend.

## Summary by Claude - reviewing code generated by Codex

Adds a small, skill-agnostic framework that lets any ClawBio skill suggest deterministic follow-up actions, and lets both chat adapters (Telegram + Discord) offer, confirm, and execute exactly those actions.

A skill opts in by emitting one or more entries in `result.json`:

````json
{
  "chat_summary_lines": ["Concise text written by the skill for chat."],
  "preferred_artifacts": ["report.md", "result.json"],
  "suggested_actions": [
    {
      "action_id": "show-demo-report",
      "label": "Show demo report",
      "request": {
        "schema": "example.skill_request.v1",
        "mode": "report",
        "report_id": "demo"
      },
      "requires_confirmation": false
    }
  ]
}
````

The runner surfaces those fields on `run_skill()`'s return value. The chat adapter renders the offer as a numbered choice list, stores it per chat with a 30-minute TTL, parses the next user reply (index, label, `yes`, `cancel`, fuzzy match) and — on confirmation — executes only the stored nested `request`, written to a temp file and run through `clawbio.py run <skill> --input`. Any `shell_line` field is intentionally stored but never executed.

## Why

So far each adapter invented continuation prose after a tool result, which is unsafe (LLM-authored shell commands), inconsistent across Telegram and Discord, and gives skills no way to drive the next step deterministically. This change moves authorship of "what should we offer next" to the skill, in structured JSON, and centralises rendering and execution in one shared library.

## What's in this PR

- `bot/action_offers.py` — shared library: extract, render, parse, TTL, execute. Pure, no I/O outside the runner callback.
- `clawbio.py` — `_promote_structured_result_fields()` reads `result.json` on success and attaches `chat_summary_lines`, `preferred_artifacts`, `suggested_actions`, and `report_md` to the runner result.
- `bot/roboterri.py` (Telegram) and `bot/roboterri_discord.py` (Discord) — same wiring on both: per-chat pending offer state, confirmation state machine, audit events, and a generic `request` parameter on the `clawbio` tool schema.
- Side-fix in the Discord adapter: `_pending_text` and `_pending_media` are now keyed per channel (previously a single global list/dict keyed on `0`), so concurrent channels no longer cross-contaminate.
- `tests/test_action_offers.py` (10 tests) and `tests/test_clawbio_runner_actions.py` (1 runner integration test) using a placeholder `example.skill_request.v1` schema.

## What's not in this PR

- No skill is added or modified. No changes to `skills/`, `skills/catalog.json`, `CLAUDE.md`, `pytest.ini`, or the catalog generator.

## Security boundary

- The chat adapter only runs the nested structured `request` payload; the optional `shell_line` field is recorded for skill authors' reference but is never executed.
- Pending offers expire after 30 minutes; ambiguous, single-word, or unrecognised replies are rejected rather than guessed.
- Existing per-skill flag allowlist on `clawbio.py run` continues to apply.

## Test plan

- [ ] `pytest tests/test_action_offers.py tests/test_clawbio_runner_actions.py`
- [ ] Manual: run a skill that emits `suggested_actions[]`, confirm the offer renders, choose `1`, confirm the nested request is executed and the follow-up reply is rendered correctly.
- [ ] Manual on Discord: verify two concurrent channels do not see each other's pending text/media (regression check for the per-channel keying fix).

### DIFF
--- a/bot/action_offers.py
+++ b/bot/action_offers.py
@@ -1,0 +1,343 @@
+"""Shared helpers for structured skill-suggested follow-up actions.
+
+Skills may return `suggested_actions[]` in result.json. The chat adapters use
+this module to render those actions, remember them briefly, resolve natural
+follow-up replies, and execute only the stored nested request payload.
+"""
+
+from __future__ import annotations
+
+import difflib
+import json
+import re
+import tempfile
+import time
+from datetime import datetime
+from pathlib import Path
+from typing import Any, Callable
+
+DEFAULT_PENDING_ACTION_TTL_SECONDS = 30 * 60
+
+# These are stand-alone confirmation/navigation phrases. They are intentionally
+# separate from numbered choices so "yes" can stay ambiguous when several
+# pending actions exist.
+_AFFIRM_ONLY = {
+    "yes",
+    "y",
+    "yeah",
+    "yep",
+    "ok",
+    "okay",
+    "sure",
+    "please do",
+    "go ahead",
+    "do it",
+    "run it",
+    "please show me",
+    "show me",
+}
+
+_CANCEL_WORDS = {
+    "cancel",
+    "no",
+    "nope",
+    "never mind",
+    "nevermind",
+    "stop",
+    "abort",
+}
+
+_INDEX_CONFIRM_PREFIXES = (
+    "yes",
+    "run",
+    "do",
+    "execute",
+    "show",
+    "please run",
+    "please do",
+    "please show me",
+    "show me",
+)
+_INDEX_CONFIRM_PATTERN = re.compile(
+    r"\s*(?:" + "|".join(re.escape(prefix) for prefix in _INDEX_CONFIRM_PREFIXES) + r")?\s*(\d+)\s*",
+    flags=re.IGNORECASE,
+)
+
+
+def _normalize_text(text: str) -> str:
+    text = text.strip().lower()
+    return re.sub(r"\s+", " ", text)
+
+
+def _token_count(text: str) -> int:
+    return len([part for part in text.split(" ") if part])
+
+
+def looks_like_action_followup(reply_text: str) -> bool:
+    """Return True for short replies that only make sense after an offer."""
+    normalized = _normalize_text(reply_text)
+    if normalized in _AFFIRM_ONLY or normalized in _CANCEL_WORDS:
+        return True
+    return _INDEX_CONFIRM_PATTERN.fullmatch(reply_text.strip()) is not None
+
+
+def extract_action_offer(result: dict[str, Any]) -> list[dict[str, Any]] | None:
+    """Return executable suggested actions from a skill result.
+
+    A displayed offer must be executable later. For that reason a valid action
+    needs a human label, a stable id, and a structured nested request. The chat
+    adapters deliberately ignore shell-only or prose-only suggestions.
+    """
+    actions = result.get("suggested_actions")
+    if not isinstance(actions, list):
+        return None
+    cleaned: list[dict[str, Any]] = []
+    for action in actions:
+        if not isinstance(action, dict):
+            continue
+        action_id = action.get("action_id")
+        label = action.get("label")
+        request = action.get("request")
+        if not isinstance(action_id, str) or not action_id.strip():
+            continue
+        if not isinstance(label, str) or not label.strip():
+            continue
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by smoe** [comment]
Hello, I had Codex prepare a minimal example and then asked Claude about its opinion who then had extra ideas how to shrink it further. The comment below is a back and forth between the two agents:

Here’s the same comment with the outer fence removed so the JSON block renders correctly:

A quick note on the `affinity-proteomics` example in this PR, since it is the first skill that actually uses the new suggested-actions contract.

A ClawBio skill normally produces a report bundle. With this PR, a skill can also include a small list of structured next steps in `result.json`. The chat adapters read those `suggested_actions`, show them as numbered choices, remember the stored request briefly, and run the selected request through the normal ClawBio path when the user replies with `1`, the action label, or a confirmation.

The `affinity-proteomics` demo keeps this deliberately small. After the main analysis, it offers one follow-up action: **Top Proteins**. The action request is not prose and not a shell command; it is typed JSON:

```json
{
  "schema": "affinity_proteomics.action_request.v1",
  "action": "top-proteins",
  "n": 5,
  "total_proteins_tested": 40,
  "significant_proteins": 5,
  "proteins": [
    {
      "protein_id": "OID00001",
      "gene": "IL6",
      "log2fc": 1.23,
      "padj": "1.00e-04"
    }
  ]
}
```

When selected, the skill handles that request in a small follow-up mode and writes a focused `report.md` / `result.json` for the top-proteins view.

This shape is intentionally modest:

- One action is enough to demonstrate the contract without turning the example into a second feature.
- The request carries structured data and parameters, so the skill still owns how the follow-up is rendered.
- `SKILL.md` now documents the exact suggested-action shape for future skill authors.
- The early action-request dispatch in `main()` is labelled, so readers can see where the follow-up path enters before the normal full pipeline.

So the example is less about affinity proteomics specifically and more about showing the intended pattern: a skill can expose a safe, deterministic next step that chat can offer without inventing commands or parsing bot prose.

manuelcorpas (approved):
## Audit verdict: SAFE — APPROVE

Adds structured suggested-action surface to bot adapters (Telegram + Discord) and clawbio.py runner.

**Security design is sound:**

- Stored `request` payloads are JSON dicts written to a `tempfile.TemporaryDirectory` and passed as `--input` to the existing `clawbio.py run` CLI
- Source comment is explicit: `shell_line is never executed from this path`
- Users can only select an offer the skill itself emitted; raw user text cannot synthesise a new executable request
- `subprocess.run([sys.executable, str(CLAWBIO_PY), ...])` is list-form (no `shell=True`), with timeout and cwd set
- `is_admin` boundaries are untouched (only appear in diff context lines)
- 30-min TTL on `_pending_actions` keyed by chat/channel id, with cancel/expiry/ambiguous paths handled
- Discord global → per-channel dict refactor is correct (channels no longer cross-contaminate)

Scope discipline is clean: bot adapters, one shared helper `bot/action_offers.py`, `clawbio.py` runner support, one demonstrator skill (`affinity-proteomics`) updated to emit the new `suggested_actions` field, two new test files.

---

PR #270, opened 2026-05-25
fix: Circumventing skill-lint YAML parsing deficiency

smoe:
## Summary

The true bug is with the regex that is used in scripts/lint_skills.py which cannot cope with the newline. As an immediate minimal fix I think this is fine. 

The SKILL.md YAML was valid, but its quoted multiline input description contained a blank line. In environments without PyYAML, `scripts/lint_skills.py` falls back to a simple regex parser, which stopped reading the `metadata:` block before it reached `metadata.openclaw`. CI therefore reported `No metadata.openclaw block`.

This PR keeps the same description text but writes it as a single-line scalar, so both real YAML parsers and the zero-dependency fallback parser see the existing `openclaw` block.

## Skill affected

de-summary

### DIFF
--- a/skills/de-summary/SKILL.md
+++ b/skills/de-summary/SKILL.md
@@ -19,10 +19,7 @@ metadata:
     - csv
     - tsv
     required: true
-    description: 'DESeq2, edgeR, or limma output table with columns: gene_id or gene_name, log2FoldChange, pvalue, padj (adjusted
-      p-value). Optional columns: baseMean, lfcSE, stat.
-
-      '
+    description: "DESeq2, edgeR, or limma output table with columns: gene_id or gene_name, log2FoldChange, pvalue, padj (adjusted p-value). Optional columns: baseMean, lfcSE, stat."
   outputs:
   - name: report
     type: file


manuelcorpas (commented):
**Automated audit verdict: NEEDS REVIEW** (0 critical, 7 warnings, 0 info)

1-line YAML parse fix in `skills/de-summary/SKILL.md`.

All 7 warnings are template-conformance gaps (`## Trigger`, `## Scope`, `## Workflow`, `## Example Output`, `## Gotchas`, `## Safety`, `## Agent Boundary`) on the pre-existing SKILL.md. They are not regressions caused by this PR; they pre-date it. The audit just reports current file state.

Suggest merging this fix on its own, then opening a follow-up to backfill the canonical sections in `skills/de-summary/SKILL.md` per `templates/SKILL-TEMPLATE.md`.

_Automated audit by ClawBio pr-audit._

manuelcorpas (approved):
**Automated audit verdict: SAFE**

Reformats the `de-summary` SKILL.md `description` from a multiline folded YAML scalar (with a trailing blank line) to a single quoted string. This fixes the skill-lint failure where the malformed YAML prevented the whole metadata block from parsing, surfacing as "No metadata.openclaw block". No prompt injection, hidden unicode, or secret exposure. Minimal, correct.

Approving and merging.

PR #289, opened 2026-06-11
Audit and harden nfcore-rnaseq-wrapper for nf-core/rnaseq 3.26.0

danilomonge:
## Overview

This PR delivers a thorough audit and hardening pass of the **`nfcore-rnaseq-wrapper`** skill — the wrapper that runs the upstream **nf-core/rnaseq 3.26.0** pipeline (FASTQ/BAM → count matrices).

Every behaviour was checked against the official nf-core/rnaseq 3.26.0 sources (`nextflow_schema.json`, `assets/schema_input.json`, and the usage/output docs) and validated with tests. The goal is to make the skill **correct, robust, portable, and reproducible** on any compatible OS, with no gap between what the documentation promises and what the code does.

All changes are contained in `skills/nfcore-rnaseq-wrapper/` — no other skills, shared modules, or routing are touched, and the branch is based on the current `main` so it merges cleanly.

## Why

The skill drives a real Nextflow pipeline with many parameters, reference combinations, and execution backends. Small mismatches — a wrong default, an unbounded value, a path that breaks on another machine — tend to surface late and confusingly inside Nextflow. This pass closes those gaps so problems are caught early with clear, actionable messages, and so a run can be reproduced on a different machine.

### Parameters & input validation (aligned to the 3.26.0 schema)
- Every CLI flag, enum, default, type, regex pattern, and numeric bound now matches `nextflow_schema.json` and `assets/schema_input.json`.
- **`--pseudo-aligner-kmer-size`** is now bounded to an odd integer in `1..31` (the Salmon/Kallisto index limit). It was the only numeric option without a guard, so an invalid value previously failed deep inside the pipeline instead of at preflight.
- **`--gtf` + `--gff` together** now follows upstream behaviour: the GTF is used and the GFF is dropped with a warning, instead of being rejected.
- A **warning** is emitted when an index flag the chosen aligner/pseudo-aligner cannot consume is supplied, so a prebuilt index is never silently ignored.

### Robustness & error handling
- An **incomplete previous run** (a `result.json` with no manifest) is detected and reported with a precise, actionable message, instead of the generic "use `--resume`" hint that cannot work in that state.
- `remap_paths.py` reports **replay-readiness consistently** across re-runs — the exit code reflects whether the bundle is ready, not whether the current invocation happened to change anything.
- Output-directory writability is validated up front; input readability is intentionally deferred to Nextflow (documented, since the container often reads as a different user); provenance never captures secrets.

### Portability & the reproducibility bundle
- The reproducibility bundle is now **portable across machines and folders**:
  - `commands.sh` uses a portable interpreter (`${PYTHON:-python3}`), so replay works on systems that ship only `python3`.
  - `commands.sh` records only **non-default** flags (consistent with `params.yaml`).
  - `report.md` includes the **skill version**.
  - Checksums use **relative paths**, so the bundle verifies correctly after being moved.
- Commands are built as **argument lists** (no shell-string assumptions); reference and output paths normalise to POSIX separators.
- On macOS/Docker, the per-process memory ceiling is **capped to the live VM size**, so a container is never OOM-killed by requesting more memory than the VM has.

## Testing

**441 tests pass**, covering:
- valid and invalid inputs, samplesheet rules, and reference combinations,
- command construction and version pinning,
- parameter / enum / bound validation,
- error handling and partial-failure reporting,
- the reproducibility bundle and cross-module drift guards,
- `clawbio run rnaseq-pipeline` flag forwarding,
- and a regression test for every fix above.

## Scope

All changes live under `skills/nfcore-rnaseq-wrapper/`. No other skills or shared modules are modified.

### DIFF
--- a/skills/nfcore-rnaseq-wrapper/README.md
+++ b/skills/nfcore-rnaseq-wrapper/README.md
@@ -6,7 +6,7 @@ ClawBio wrapper for running `nf-core/rnaseq` bulk RNA-seq preprocessing from FAS
 
 - Upstream bulk RNA-seq preprocessing via Nextflow.
 - Supported aligners: `star_salmon`, `star_rsem`, `hisat2`, `bowtie2_salmon`.
-- Validated samplesheet input with absolute normalized FASTQ/BAM paths.
+- Validated samplesheet input with normalized local FASTQ/BAM paths or preserved remote URIs.
 - Reproducibility bundle, provenance bundle, `report.md`, and `result.json`.
 - Canonical merged count matrix detection for downstream `rnaseq`.
 
--- a/skills/nfcore-rnaseq-wrapper/SKILL.md
+++ b/skills/nfcore-rnaseq-wrapper/SKILL.md
@@ -20,11 +20,11 @@ metadata:
       format:
         - csv
       description: >
-        nf-core/rnaseq samplesheet. Required columns: sample, strandedness. FASTQ mode
-        adds fastq_1 (and optional fastq_2). BAM reprocessing mode adds genome_bam and/or
-        transcriptome_bam plus percent_mapped. Optional metadata columns: seq_platform,
-        seq_center. A row may not mix FASTQ and BAM inputs.
-      required: true
+        nf-core/rnaseq samplesheet. Required columns: sample, fastq_1, strandedness.
+        FASTQ mode may add fastq_2. BAM reprocessing mode preserves the original FASTQ
+        columns and adds genome_bam and/or transcriptome_bam plus percent_mapped; use
+        it only with --skip-alignment. Optional metadata columns: seq_platform, seq_center.
+      required: false  # required for real runs; not for --demo or self-contained nf-core test profiles (the only universally required CLI arg is --output)
   outputs:
     - name: report
       type: file
@@ -37,7 +37,7 @@ metadata:
         - json
       description: Structured result payload with detected count matrices and provenance
   dependencies:
-    python: ">=3.11"
+    python: ">=3.10"
     packages: []
   demo_data:
     - path: demo/README.md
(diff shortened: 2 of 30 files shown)


danilomonge:
**Note on CI** — the single failing check (`Run Skill Builder tests`) is not related to this PR.

`skill-builder/tests/test_skill_builder.py::TestPatchClawbioPy::test_patch_on_real_clawbio_py` also fails on a clean `main`. Its `patch_clawbio_py()` looks for the `SKILLS = {…}` dict inside `clawbio.py`, but `clawbio.py` is now a thin shim and the registry lives in `clawbio/cli.py`, so the patcher can't find an insertion point and returns `False`.

This PR is confined to `skills/nfcore-rnaseq-wrapper/` and does not touch `clawbio` or `skill-builder`. (The CI workflow's per-skill steps don't currently include the `nfcore-rnaseq-wrapper` tests, so this skill's 441 tests are validated locally rather than in this run.)

Suggested separate fix: point `skill-builder`'s `patch_clawbio_py()` / its `SKILLS` AST lookup at `clawbio/cli.py`.

manuelcorpas (approved):
## pr-audit: SAFE ✅

Audited the nfcore-rnaseq-wrapper hardening (+3,372 / -360, 30 files).

**Security:** clean. No secrets, no `eval`/`exec`/`os.system`/`shell=True`, no hidden unicode, no unexpected outbound network calls. The single `subprocess.run` (the Nextflow invocation) uses no shell and no string interpolation.

**SKILL.md conformance:** PASS. All required sections present (Trigger, Scope, Workflow, Example Output, Gotchas, Safety, Agent Boundary); `version`/`author` set under `metadata`; both Trigger lists with routing; Example Output code block; safety disclaimer; `--demo` flag; 354 lines.

**Tests:** 16 test files (~3,000 lines) across preflight, params, samplesheet, executor, outputs, provenance, drift guards, and reporting.

**Scope:** in scope (transcriptomics / bulk RNA-seq).

**Notes (non-blocking):** three small helper modules (`_isolated_imports.py`, `errors.py`, `schemas.py`) have no dedicated `test_<module>.py` and are covered indirectly. No `clawbio_bench` harness exists for pipeline wrappers, so the scientific-correctness bench was not applicable.

Thorough, well-tested hardening work. Approving.

---

PR #295, opened 2026-06-17
Fix nf-core wrapper test harness after the clawbio CLI package split

danilomonge:
## Summary

On a clean checkout, the test suites for two of the nf-core wrappers fail — 12 tests in total (4 in `nfcore-rnaseq-wrapper`, 8 in `nfcore-scrnaseq-wrapper`). The failures all trace back to the `clawbio.py` → `clawbio/` package split: the runner registry (`SKILLS`) and `__version__` now live in `clawbio/cli.py`, but the two wrappers' test harnesses still referenced the pre-split locations.

The production code is already correct — every flag the tests assert on is present in the real allow-lists. Only the tests' import/access paths were stale, so this PR is **test-infrastructure only and changes no runtime behaviour**.

### `nfcore-rnaseq-wrapper` (4 failures)

The test `conftest.py` pre-stubs the `clawbio` package — a deliberate optimisation that keeps the test session light by avoiding a full import of the CLI engine. That stub no longer exposed `__version__`. Because `clawbio/cli.py` begins with `from clawbio import __version__`, importing `clawbio.cli` to read `clawbio.cli.SKILLS` raised:

```
ImportError: cannot import name '__version__' from 'clawbio'
```

**Fix:** the stub now reads the real version straight from `clawbio/__init__.py` (parsed with `ast`, without executing the heavy `__init__`) and exposes it, so the stubbed value can never drift from the package version.

### `nfcore-scrnaseq-wrapper` (8 failures)

Two tests loaded the top-level `clawbio.py` shim by path and then read `clawbio.SKILLS` / monkeypatched `clawbio.run_skill`. After the split, both `SKILLS` and `run_skill` live in `clawbio.cli`; `clawbio.py` is a thin shim that only delegates to it. As a result:

- reading `SKILLS` off the shim raised `AttributeError: module 'clawbio_script' has no attribute 'SKILLS'`, and
- patching `run_skill` on the shim had no effect on the binding `main()` resolves internally, so the call was not intercepted and exited with "No input provided".

**Fix:** the tests now target `clawbio.cli` directly — the canonical home of the runner registry and callables. This both restores `SKILLS` access and makes the monkeypatch reach the function `main()` actually calls.

## Verification

Per-wrapper suites, run the way CI runs them (one wrapper at a time):

| Wrapper | Result |
|---|---|
| `nfcore-rnaseq-wrapper` | 441 passed |
| `nfcore-sarek-wrapper` | 321 passed (already green) |
| `nfcore-scrnaseq-wrapper` | 387 passed, 2 skipped |

(The 2 skips are intentional — they require a sibling `nf-core/scrnaseq` checkout to cross-check against.)

All three `--demo` pipelines were also run end-to-end and launch/execute correctly under Nextflow (`nf-core/rnaseq` 3.26.0, `nf-core/sarek` 3.8.1, `nf-core/scrnaseq`): containers run, processes complete, and the reproducibility bundle is written.

## Scope

Three files, all under the wrappers' `tests/` directories — no production module is modified:

- `skills/nfcore-rnaseq-wrapper/tests/conftest.py`
- `skills/nfcore-scrnaseq-wrapper/tests/test_audit_fixes.py`
- `skills/nfcore-scrnaseq-wrapper/tests/test_audit_round2.py`

### DIFF
--- a/skills/nfcore-rnaseq-wrapper/tests/conftest.py
+++ b/skills/nfcore-rnaseq-wrapper/tests/conftest.py
@@ -1,9 +1,23 @@
-"""
-conftest.py — pre-stub clawbio package to avoid loading runner.py
-(runner.py loads clawbio.py which uses `str | None` syntax, Python 3.10+ only)
+"""conftest.py — pre-stub the ``clawbio`` package for this skill's tests.
+
+The stub maps ``clawbio`` / ``clawbio.common`` onto the real package directories
+via ``__path__`` and eagerly loads only the leaf helper modules the wrapper needs
+(``clawbio.common.checksums`` and friends). This keeps the test session light:
+importing the real ``clawbio`` package would execute ``clawbio/__init__.py`` ->
+``from .runner import ...`` -> ``from clawbio.cli import ...``, pulling the whole
+CLI engine into every test.
+
+The runner-allowlist tests do need ``clawbio.cli`` (for ``clawbio.cli.SKILLS``).
+Loading that module runs ``clawbio/cli.py``, whose first line is
+``from clawbio import __version__``. Because we deliberately skip executing
+``clawbio/__init__.py``, the stub must expose ``__version__`` itself, otherwise
+that import fails with ``ImportError: cannot import name '__version__'``. We read
+the value straight from ``clawbio/__init__.py`` so the stub never drifts from the
+real package version.
 """
 from __future__ import annotations
 
+import ast
 import importlib.util
 import sys
 import types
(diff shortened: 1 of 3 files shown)


manuelcorpas (approved):
**Audit verdict: SAFE, approving**

Clean, correctly-targeted fix. After the engine moved from the top-level `clawbio.py` script into `clawbio/cli.py`, the path-loaded tests were hitting the shim; repointing them at `clawbio.cli` is right. Reading `__version__` via `ast.literal_eval` (non-executing) is the safe approach, no assertions were removed or tests skipped, and the `test_audit_round2` monkeypatch change actually closes a binding gap where patches hit the wrong module. Tests still verify real behaviour.

Minor (non-blocking): the `except OSError/ValueError: return "0.0.0"` fallbacks swallow a malformed `__init__.py` silently rather than failing loudly.


---

PR #296, opened 2026-06-18
Organ Aging Studio - interpretable Goeminne proteomic clocks

alinjfz:
## Summary

This PR adds `organ-aging-studio`, an inspectable wrapper around the published Goeminne et al. (2025) organ-specific proteomic aging clocks.

The goal is to keep the science honest while making the output more useful for demos and review:

- `proteomics-clock` remains the batch-oriented prediction path.
- `organ-aging-studio` exposes ranked protein contributions and filter controls so reviewers can see which proteins drive each organ age.
- The branch has been rebased onto the current `ClawBio/ClawBio:main`, so it now compares cleanly against the latest upstream base.

## What this includes

- `skills/organ-aging-studio/SKILL.md` with scope, workflow, gotchas, and safety guidance
- `skills/organ-aging-studio/organ_aging_studio.py` for interpretable organ-age breakdowns
- `skills/organ-aging-studio/tests/test_organ_aging_studio.py` covering the demo flow and the CLI guard

## Review-related fixes

- Added a friendly error when the CLI is run without `--demo` or `--input`.
- Added an input sanity section so suspiciously scaled inputs do not silently look like valid Olink NPX.
- Marked filtered runs (`--top-n`, `--min-abs-coef`) as partial-model predictions instead of the fully validated published clock.
- Clarified that the reported delta is raw predicted-minus-chronological age, not residualized acceleration.
- Added checksum verification for cached coefficient downloads.
- Removed an unused `matplotlib` dependency from the studio requirements.

## Demo / outputs

```bash
python skills/organ-aging-studio/organ_aging_studio.py \
  --demo --output /tmp/organ-aging-studio \
  --organs Heart,Brain,Immune,Organismal \
  --sample-id DEMO_000 --top-n 10
```

Expected outputs:

- `report.md`
- `result.json`
- `tables/protein_contributions.csv`
- `commands.sh`

## Verification

- `python3 -m pytest skills/proteomics-clock/tests/test_proteomics_clock.py skills/organ-aging-studio/tests/test_organ_aging_studio.py -q`
- `16 passed`

## Notes

- `pyaging_data/` is a local untracked artifact directory and is intentionally not part of this PR.
- The branch was synced by rebasing onto current `origin/main` and force-pushing the feature branch to the fork.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ skills/pharmgx-reporter/report/
 skills/genome-compare/data/manuel_corpas_23andme.txt
 skills/pr-audit/
 data/
+pyaging_data/
 !skills/methylation-clock/data/
 !skills/methylation-clock/data/GSE139307_small.csv.gz
 !skills/rare-disease-rnaseq/data/
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,6 @@ skills/pharmgx-reporter/report/
 skills/genome-compare/data/manuel_corpas_23andme.txt
 skills/pr-audit/
 data/
-pyaging_data/
 !skills/methylation-clock/data/
 !skills/methylation-clock/data/GSE139307_small.csv.gz
 !skills/rare-disease-rnaseq/data/
--- a/skills/organ-aging-studio/SKILL.md
+++ b/skills/organ-aging-studio/SKILL.md
@@ -42,7 +42,6 @@ metadata:
     packages:
       - pandas>=2.0
       - numpy>=1.24
-      - matplotlib>=3.7
       - requests>=2.28
   demo_data:
     - path: ../proteomics-clock/data/demo_olink_npx.csv.gz
@@ -104,6 +103,7 @@ predicted_age = intercept + Σ (protein_NPX × coefficient)
 | Hard to explain to clinicians / judges | `report.md` + `protein_contributions.csv` + JSON for agents |
 
 Built on the same pinned [organAging](https://github.com/ludgergoeminne/organAging) coefficients as `proteomics-clock`. **No invented weights.**
+Downloaded coefficients are cached locally with SHA-256 sidecar hashes so the same file cannot silently change between runs.
 
 ## Core Capabilities
 
@@ -111,6 +111,18 @@ Built on the same pinned [organAging](https://github.com/ludgergoeminne/organAgi
 2. **Gen1 / Gen2** — chronological age models or mortality hazard → years (Gompertz)
 3. **Protein filters** — `--top-n`, `--min-abs-coef`, single `--sample-id`
 4. **Structured outputs** — Markdown report, JSON, contribution table, replay `commands.sh`
+
+## Scope
+
+One skill, one task. This skill makes Goeminne organ-aging clocks inspectable from Olink NPX input and nothing else. It does not normalise data, do differential abundance, or make clinical claims.
+
+## Workflow
+
+1. **Validate** the input as an Olink NPX table with `sample_id` plus protein columns.
+2. **Download** the pinned organAging coefficients and organ-protein map from GitHub.
+3. **Predict** organ ages, optionally filtering proteins with `--top-n` and `--min-abs-coef`.
+4. **Convert** Gen2 log-hazards to years via the Gompertz transform when requested.
+5. **Write** `report.md`, `result.json`, `protein_contributions.csv`, and a replayable `commands.sh`.
 
 ## Input Formats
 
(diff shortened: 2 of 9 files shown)


manuelcorpas (commented):
**Audit verdict: needs review**

Nice work, and thank you for sourcing the coefficients honestly from the commit-pinned organAging repository rather than inventing weights. No malware or secrets. A few things before merge:

SKILL.md conformance:
- Missing `## Scope`, `## Workflow` (numbered), and `## Gotchas` (>= 3 failure modes) sections. Please see `templates/SKILL-TEMPLATE.md`.

Code:
- `--input` argument is dereferenced before the guard, so running with neither `--demo` nor `--input` raises a TypeError instead of a friendly error. Reorder the check.
- `matplotlib` is declared but never imported. Drop it.

Scientific calibration (please at least document):
- No NPX standardisation before applying weights; the published clocks were trained on a specific normalisation, so raw NPX will miscalibrate the age.
- The reported delta is a raw predicted-minus-chronological gap, not age-residualised acceleration, so it is biased with age. The example "Brain -24 yr" row illustrates this and should carry a caveat or be residualised.
- No checksum on the downloaded coefficient files (URL is commit-pinned, which is good; a SHA-256 check would close the loop).


---

manuelcorpas (commented):
## Audit verdict: NEEDS REVIEW (two doc fixes, one check before merge)

Scientifically honest wrapper around the merged proteomics-clock skill: coefficients are real and SHA-pinned (`ORGANAGING_COMMIT` -> Goeminne organAging GTEx CSVs), no fabricated ages, the Goeminne et al. 2025 disclaimer is rendered, and the decomposition logic is unit-tested with a hand-computed expected value. Local-first holds (only coefficients are fetched, Olink NPX never leaves the machine). Three things to address.

**1. NPX normalisation caveat not propagated.** The studio multiplies raw input columns by NPX-trained coefficients with no input-unit validation. The parent skill carries the critical gotcha ("Olink NPX is already log2-scale, do NOT log-transform; non-Olink data needs rescaling via Table S3 SDs") but this SKILL.md omits it. A user feeding SomaLogic / re-log-transformed data silently gets a wrong organ age. Please copy the units gotcha across and add an input sanity check.

**2. Filtered predictions presented as the validated clock.** With `--top-n` / `--min-abs-coef` active, the report still prints a confident signed `Delta` (e.g. "Brain -24 yr") even though most of the model has been dropped. Please label filtered/partial predictions as illustrative, not the validated clock.

**3. Please confirm `--fold` semantics.** `download_coefficients` indexes `rows[fold-1]` of the coefficient CSV, treating each row as a CV fold. Whether row N actually corresponds to fold N (vs a header/aggregate row) is unverified, and no test checks the real coefficients against a known expected age.

Also add a `## Gotchas` section (currently missing) so the units and filter caveats have a home.

manuelcorpas (commented):
**ClawBio domain audit — SAFE (merge-ready after minor polish)**

The key failure mode for an aging-clock PR (placeholder coefficients presented as the published clock) does not occur here. Coefficients are downloaded from the real Goeminne organAging repo at a pinned commit (`ORGANAGING_COMMIT = 5147b03...`, live-verified, genuine organ-specific proteins), the DOI `10.1016/j.cmet.2024.10.005` resolves to the correct Cell Metabolism paper, the demo data is transparently synthetic, biological-age claims carry disclaimers and an agent boundary, and the prediction maths is unit-tested against exact values. Local-first; the only network access is to the pinned public coefficient files.

Two cosmetic fixes before merge:
- Align the SKILL.md section headings with the repo convention (`## Scope`, `## Workflow`, `## Example Output`, `## Gotchas`; `## Safety` rather than `## Safety Rules`). The content already exists under other headings.
- Drop the unused `matplotlib>=3.7` dependency from requirements.txt and the SKILL.md frontmatter (the script does no plotting).

No scientific or safety changes required. Recommend merge once the above are addressed.

manuelcorpas (commented):
## PR Audit: Comment (non-blocking)

Solid contribution — no fabricated science here. The citation checks out (PMID 39488213, Goeminne et al. *Cell Metab* 2025), coefficients are delegated to the real `organAging` repo rather than invented, there is no `np.random`, the demo NPX is honestly labelled synthetic, and the core prediction formula is unit-tested against a known expected value. A couple of things to tighten before this is unconditionally safe:

- **Upstream coefficient pinning (please confirm):** all scientific correctness rests on the sibling `proteomics-clock` skill actually pinning a specific `organAging` commit/release rather than a moving `main`. This diff contains none of that logic, so it can't be verified from the PR alone. Please confirm the pin out-of-band.
- **Silent missing-protein dropping:** `filter_protein_coefs` keeps only proteins present in the sample and still emits `predicted_age_years` from the partial subset. On a sparse Olink panel this yields a biased partial age with no warning. Recommend a coverage-based abstain/flag threshold (you already report `n_proteins_used/n_proteins_in_model`, which is the right signal to gate on). It correctly does **not** zero-impute, which is good.
- **NPX normalisation:** confirm whether the upstream coefficients expect z-scored/standardised NPX; raw-NPX input against standardised coefficients would produce wrong ages.
- **Minor:** add a `## Gotchas` section (currently missing — conformance gap); drop the unused `matplotlib`/`requests` dependencies from `requirements.txt`.

_Automated audit by ClawBio pr-audit. Manual review is still recommended._

alinjfz:
I rebased this branch onto the current `ClawBio/ClawBio:main` and force-pushed the synced history, so the PR now compares cleanly against the latest upstream base.

I also updated the PR description to reflect the audit feedback more clearly. The main fixes are:

- The CLI now fails with a friendly parser error if neither `--demo` nor `--input` is provided.
- The report now includes an input sanity section so suspiciously scaled inputs are called out instead of being treated as valid Olink NPX without warning.
- `organ-aging-studio` now labels `--top-n` / `--min-abs-coef` runs as partial-model predictions rather than the fully validated published clock.
- The reported delta is now explicitly described as the raw predicted-minus-chronological gap, not residualized acceleration.
- Cached coefficient downloads now include SHA-256 verification.
- The `SKILL.md` documentation was brought into template conformance, including `Scope`, `Workflow`, and `Gotchas`.
- The unused `matplotlib` dependency was removed from the studio requirements.

I also verified the focused test suite again:
`python3 -m pytest skills/proteomics-clock/tests/test_proteomics_clock.py skills/organ-aging-studio/tests/test_organ_aging_studio.py -q`
and it passed (`16 passed`).

manuelcorpas (approved):
## PR Audit: SAFE (approve with two polish items)

Deep domain review independently verified the provenance and prediction path, and it holds up well. Coefficients load from the real `ludgergoeminne/organAging` repo at pinned commit `5147b03` (verified reachable, coefficient file paths resolve), the prediction is the published linear formula `intercept + sum(npx*coef)`, there is no `np.random` or synthesised weight anywhere in the prediction path, and the arithmetic is unit-tested. Strong guardrails (Agent Boundary, Safety, embedded disclaimer).

### Polish before merge (non-blocking)

1. **Partial-model predictions keep the full-panel intercept.** With `--top-n`/`--min-abs-coef`, `compute_organ_age` drops proteins but retains `intercept` fitted for the full panel, biasing the result toward the mean training age, then pushes the truncated sum through `mortality_to_years`. It is disclosed via `is_partial_prediction`, but recommend suppressing or relabelling the hard year value when partial, rather than presenting it as an age.
2. **Demo table shows a dramatic gap.** SKILL.md shows "Brain ~42 yr vs 66 yr, -24 yr" on synthetic NPX. It is caveated, but a skimming reader could misread it as a real finding; consider relabelling the numbers as illustrative only.

Minor: `raw_delta_years` duplicates `age_delta_years`, and `n_proteins_available` equals `n_proteins_used`; harmless but mislabelled.

PR #302, opened 2026-06-20
fix(cvr): use versioned Ensembl transcript accessions

camlloyd:
## Summary

- Add `transcript_version=1` param to VEP REST API call so transcript IDs include version suffixes (e.g. `ENST00000357654.9` instead of `ENST00000357654`)
- Update demo evidence cache with current Ensembl transcript versions
- Add test asserting all demo transcripts use versioned accessions per HGVS v21.1

## Skill affected

clinical-variant-reporter

## Checklist

## Test output

```
37 passed in 0.49s
```


manuelcorpas (commented):
**Audit verdict: needs review (low risk).**

Confirmed safe in substance: matching is purely coordinate-based (`chrom:pos:ref:alt`), and `transcript` is a display-only field in the report, so adding the `.version` suffix cannot drop or mis-assign a clinical annotation. No security concerns, and the demo cache update is internally consistent (same transcript, same version across rows).

One gap before merge: the only live-mode behaviour change, `transcript_version: 1` on the VEP REST call, has no test. The new test runs in demo mode and asserts the versioned-format property, so the suite would stay green even if the VEP param were removed or broken. Please add a test that exercises the live path (or mocks the VEP response) to lock in the param. Minor: the hardcoded Ensembl version numbers in the demo cache are display-only but will drift across Ensembl releases.


---

manuelcorpas (approved):
## Audit verdict: APPROVE

Mechanically correct and, unusually for a one-line live-path change, properly tested. Spot-checked 5 of the 12 accessions against live Ensembl GRCh38, all MANE Select with correct versions (BRCA1 .9, TP53 .9, BRCA2 .8, HFE .10, LDLR .11). `transcript_version: 1` is the correct VEP REST usage, and `TestVepLivePath.test_transcript_version_param_sent` would fail if the param were dropped, so this is not demo-mode-only green. No allele-frequency or effect-size fields touched; pinning the version does not shift any reported coordinate or consequence.

One pre-existing item to fix or file separately (not introduced by this PR): the LDLR demo record `ENST00000252444.11` carries `c.10580G>A` / `p.Arg3527Gln`, but that transcript has a ~2,838 nt CDS (946 aa), so the coordinate is impossible, and it is the Ensembl Canonical, not the MANE Select (ENST00000558518) for LDLR. Versioning a clinically-wrong transcript hard-codes the error. Please correct the record or open a follow-up so "versioned" is not read as "validated".

PR #310, opened 2026-06-30
Harden the three nf-core pipeline wrappers and the ClawBio launcher: forwarding, output contract, remote/reference handling, failure diagnostics

danilomonge:
## Summary

This PR fixes repository-level inconsistencies in the three nf-core pipeline
wrappers (`nfcore-rnaseq`, `nfcore-sarek`, `nfcore-scrnaseq`) and the ClawBio
launcher (`clawbio/cli.py`) that govern how options are forwarded, how outputs are
laid out, how remote and custom-reference inputs are handled, and how failures are
explained. The goal is that the three wrappers behave consistently with each other
and precisely follow the official nf-core documentation for their pinned versions
(rnaseq 3.26.0, sarek 3.8.1, scrnaseq 4.1.0).

Every change is covered by tests (red → green). Suites stay green in isolation:
`clawbio/tests` 103, `nfcore-rnaseq-wrapper` 469, `nfcore-sarek-wrapper` 365,
`nfcore-scrnaseq-wrapper` 412 (+2 skipped). All three wrappers were also run
end-to-end on official nf-core test data (see **Validation**).

The sections below are grouped by theme rather than by the order issues were found.

---

## 1. Launcher: option forwarding and flag spellings

- **Nextflow config files were dropped depending on spelling.** Each wrapper accepts
  `-c`, `--config`, and `--nextflow-config` as aliases of one repeatable option, but
  the launcher declared two separate argparse options and each pipeline forwarded
  only one, so a valid config was silently discarded for the "wrong" spelling. The
  two options are merged into one and every entry is forwarded as
  `--nextflow-config` for all three pipelines.

- **nf-core-native (snake_case) flag spellings are now accepted.** nf-core params are
  snake_case (`--skip_tools`, `--fasta_fai`, …) while the wrappers expose hyphenated
  flags; the launcher's INT-001 allowlist matched exact (hyphenated) tokens, so a
  command copied from the upstream docs had the token silently dropped. The launcher
  now canonicalises `_`↔`-` when matching the allowlist for the three pipeline skills
  and forwards the wrapper's canonical hyphen spelling; the sarek parser also
  registers the native `--<param>` alias for direct invocation. Scoped to the
  pipeline skills — other skills keep exact-match filtering, and the normalisation
  can never admit a flag that is not already on the allowlist.

- **Missing / mis-classified allowlist entries.** `rnaseq` `--timeout-hours` and
  `--allow-pipeline-version-override` were absent from the allowlist; `sarek`
  `--allow-remote-inputs` / `--allow-pipeline-version-override` (both `store_true`)
  were listed only among value-taking flags (so the filter could consume the next
  token). Fixed, with a launcher↔wrapper allowlist-consistency test added per suite.

## 2. Unified output contract across the three wrappers

- **Output layout standardised.** `nfcore-sarek` wrote `report.md`, `result.json`,
  and logs under `reproducibility/`; the other two write them at the output root with
  logs in `<output>/logs/`. Sarek now matches: `report.md`/`result.json` at the root,
  logs in `<output>/logs/`, both excluded from `checksums.sha256`; `reproducibility/`
  keeps the portable replay + provenance bundle. A consumer now finds
  `<output>/result.json` for any of the three pipelines (on failure too).

- **Output-directory guard made consistent.** Writing outputs inside the ClawBio
  source tree is rejected with a dedicated `OUTPUT_DIR_INSIDE_REPO` code across all
  three (was missing on scrnaseq and mislabelled `OUTPUT_DIR_NOT_WRITABLE` on rnaseq).

- **`result.json` shares a minimal contract.** All three now carry `status`
  (`"ok"`/`"error"`) and `ok` (bool) on success and failure. For rnaseq/scrnaseq this
  is an opt-in `status`/`ok` on the shared `write_result_json` helper, so the ~40
  other skills that use it are unaffected.

- **Sarek report `Profile` / `Java` / `Nextflow` are populated** (were `-`): preflight
  now returns the detected Java/Nextflow versions and the composed profile is threaded
  into the report and `result.json`, matching the sibling wrappers.

- **Version strings keep their zero-padded components (sarek + rnaseq).** The version
  shown in `report.md`/`result.json` was reconstructed from the integer comparison
  tuple, so `int("04") = 4` turned `26.04.3` into `26.4.3` — not a real Nextflow
  release, and inconsistent with `manifest.json`, which kept the correct string. Both
  wrappers now record the version exactly as reported for display and use the tuple
  only for the minimum-version comparison. scrnaseq already did this.

- **`--demo` reports effective samples and tools.** Under the hermetic upstream
  `-profile test`, the local samplesheet is empty and no tools are requested, so
  reports showed `Samples: 0` / `Tools: (none)`. The count now falls back to the
  samples detected in the outputs, and sarek's tools fall back to those observed in
  the parsed variant-calling/annotation outputs; scrnaseq's count no longer returns a
  literal `0` instead of the detected samples. All derived from actual outputs, never
  assumed.

## 3. Remote input and reference/genome handling (local-first, URI-safe)

- **`--allow-remote-inputs` opt-in, with precise semantics.** Remote samplesheet
  inputs and reference paths (`s3://`, `gs://`, `https://`, `ftp://`, …) are rejected
  at preflight (`REMOTE_INPUT_NOT_ALLOWED`) unless the flag is passed, in which case a
  runtime warning names every path fetched over the network. Shared across the three
  wrappers. SKILL.md now states explicitly what the flag does and does not do: it
  relaxes only the wrapper's own local-first preflight check; the wrapper does **not**
  download anything; remote URIs are written into the normalized samplesheet/`params.yaml`
  verbatim and staged **natively by Nextflow**, so they require outbound network and are
  incompatible with `NXF_OFFLINE` (under which nf-core's own nf-schema file-existence
  check still runs and fails on the remote paths).

- **Remote reference URIs are preserved in `params.yaml`.** With `--allow-remote-inputs`,
  scrnaseq resolved a remote `--fasta`/`--gtf`/index URI as a local path, collapsing
  the scheme (`https://host/x` → `<cwd>/https:/host/x`) and failing nf-core's parameter
  validation. Reference fields now preserve URIs verbatim (`_posix_or_uri`) and only
  resolve local paths. The same latent bug was fixed on sarek's MultiQC config/logo
  fields for parity; rnaseq already handled it correctly.

- **Custom `--fasta` without `--genome` disables iGenomes automatically (sarek).**
  nf-core/sarek 3.8.1 defaults `genome = 'GATK.GRCh38'`, so a custom-reference run that
  supplied `--fasta` but left `--genome` unset still loaded the full GATK.GRCh38
  iGenomes config and aborted while validating its ~20 `s3://ngi-igenomes/...` paths as
  local files — a wall of "does not exist" errors that never mentioned
  `--igenomes-ignore`. The wrapper now sets `igenomes_ignore=true` in that case (with a
  WARNING recorded in the report), matching the nf-core custom-reference guidance and
  the equivalent behaviour already in scrnaseq. An explicit `--genome` alongside
  `--fasta` (documented partial override) and `--demo`/`-profile test` runs are left
  untouched.

- **The BQSR preflight error explains that baserecalibrator runs by default (sarek).**
  When a start step needs `--dbsnp`/`--known_indels` and neither is available, the
  `MISSING_REFERENCE` message now adds that baserecalibrator runs by default in the
  Sarek mapping workflow, independently of the requested variant-calling `--tools` — so
  a user who only asked for, say, `--tools haplotypecaller` understands why the
  known-sites requirement (and the fix: `--skip-tools baserecalibrator`, or supplying
  the resources) applies. The enforcement itself is unchanged.

## 4. Actionable failure diagnostics on `EXECUTION_FAILED`

On a non-zero exit the executors scan the captured Nextflow logs and append an
actionable hint when a known signature is present — always diagnosed from the actual
error text (never predicting pipeline internals), so a hint only appears when
relevant. Shared verbatim across the three wrappers unless noted:

- **Host smaller than an nf-core default request** (`Process requirement exceeds
  available memory`) → cap resources with a `-c` config using `process.resourceLimits`.
- **Unreachable network** (`Network is unreachable` / Java connection exception),
  common on IPv6-only / NAT64 hosts → verify DNS/HTTPS and set
  `NXF_OPTS='-Djava.net.preferIPv6Addresses=true'`. The wrappers never override a
  user-set `NXF_OPTS`.
- **Config-parse failure** (`Unable to parse config file` / `ConfigParseException`),
  seen on Nextflow 26.x when the remote `nfcore_custom.config` cannot be fetched → set
  `NXF_OFFLINE=true` for a fully local run, or confirm outbound HTTPS/DNS.
- **macOS + Docker `/tmp`**: Colima does not share `/tmp` into its VM, so a `/tmp`
  work-dir hard-fails with `.command.run: No such file or directory`; preflight warns
  and the executor repeats the actionable hint (move `--output` under `$HOME`).
- **rnaseq all-skipped run** (`EXPECTED_OUTPUTS_NOT_FOUND`): nf-core drops samples
  below `min_trimmed_reads` (default 10000) before quantification, so a tiny-input run
  completes with no count matrix → the fix now points at `--min-trimmed-reads`.
- **scrnaseq CellBender** (`CELLBENDER_REMOVEBACKGROUND` errors on very small/test
  datasets) → CellBender is optional; re-run with `--skip-cellbender`.
- **`--demo` under `NXF_OFFLINE`**: demo runs the upstream remote `-profile test`, so
  preflight fails fast with `DEMO_REQUIRES_NETWORK` instead of a cryptic
  nf-schema `does not exist`. Docs clarify `--demo` only *downloads* public test data,
  so it is compatible with the local-first guarantee.

## 5. scrnaseq `protocol` samplesheet column

nf-core/scrnaseq 4.1.0's `schema_input.json` does not define a `protocol` column, but
the pipeline's own example `assets/samplesheet.csv` ships one, so a user copying it
was warned "unrecognised column: protocol". `protocol` is now recognised and preserved.
Per the usage docs, protocol is set via the `--protocol` flag (a global parameter, not
a per-sample column), so the wrapper still requires `--protocol` and its preflight
error now says so explicitly — no undocumented per-sample-protocol behaviour is
introduced.

## 6. Reviewed and documented (no behaviour change)

- **Sarek custom-reference indices are built automatically.** A custom local `--fasta`
  has its `.fai` index and `.dict` sequence dictionary built automatically by Sarek
  ("fasta is the only required input"); the wrapper passes the FASTA through without
  requiring pre-built indices. Together with the BQSR known-sites enforcement (already
  early at preflight, now with the clearer message in §3), this matches nf-core/sarek
  3.8.1 exactly. Documented as gotchas.
- **Benign nf-schema warning.** sarek 3.8.1's schema plugin mislabels argv[0]
  (`nextflow`) as a positional argument; the wrapper's command is well-formed and
  identical in shape to the sibling wrappers. Documented so users are not alarmed.

## Validation

All three wrappers were run end-to-end on official nf-core test samplesheets and
references (both local files and public remote inputs), and all three completed
successfully — rnaseq (STAR+Salmon count matrices), sarek (Strelka germline + somatic
VCFs), scrnaseq (STARsolo/simpleaf h5ad/SCE/Seurat). The diagnostics in §4 and the
fixes in §2/§3/§5 came directly from these runs. Failures encountered during
validation that are **not** wrapper defects were triaged and left unchanged: process
memory defaults exceeding host RAM (nf-core exposes no `--max_memory` CLI param — the
`-c resourceLimits` path the hint recommends is the faithful one); the `NXF_OFFLINE`
↔ remote-input coupling (offline disables remote staging by design); tiny toy
references unfit for simpleaf/Salmon; the `testdata.nf-core.sarek` test genome's own
annotation config; and upstream Nextflow-26 warnings (`prov.enabled`, flowcell-ID
extraction).

## Documentation conformance

The wrappers were cross-checked against the official nf-core docs (parameters / usage /
output) for the pinned versions and match: version pins are exact (3.26.0 / 3.8.1 /
4.1.0); defaults either match the docs (`min_trimmed_reads` 10000, sarek `aligner`
bwa-mem and the boolean flags off, scrnaseq `aligner` simpleaf / `star_feature` Gene)
or are deliberately deferred to upstream ("omit = trust upstream default", so a future
upstream change is never silently overridden); allowed-value sets are faithful
(scrnaseq validates `--protocol` against the pinned `protocols.json` and covers all six
aligners); and the behaviours the new hints/gotchas/fixes describe all match the pinned
documentation — including, confirmed for this round against the sarek 3.8.1 parameters
and usage pages: `genome` defaults to `GATK.GRCh38` so a custom `--fasta` requires
`igenomes_ignore`; baserecalibrator runs by default in the mapping workflow and needs
known sites; a custom FASTA is the only required custom-reference input; plus (from
earlier rounds) CellBender on by default, `min_trimmed_reads` sample-skipping, and
protocol via `--protocol`.

## Testing

- New `clawbio/tests` for config forwarding (every spelling × pipeline) and
  snake_case flag spellings.
- Per-wrapper allowlist-consistency and regression tests for each newly forwarded flag.
- Output-contract tests: root layout, `OUTPUT_DIR_INSIDE_REPO`, `result.json`
  `status`/`ok`, sarek runtime metadata, demo samples/tools.
- Remote-URI preservation tests (scrnaseq references, sarek MultiQC) and local-path
  resolution.
- Executor failure-hint tests across all three wrappers (memory, network, `/tmp`,
  config-parse, CellBender) plus rnaseq skipped-samples and the `--demo` offline guard.
- scrnaseq `protocol` column recognition.
- sarek custom-`--fasta` auto-`igenomes_ignore` (and the untouched explicit-`--genome`,
  no-`--fasta`, and `--demo` cases), and the BQSR-runs-by-default error message.
- Zero-padded Nextflow/Java version preservation (sarek + rnaseq).


manuelcorpas (approved):
## PR Audit: Approve

This is a genuine correctness-and-consistency fix, and the subprocess/flag handling is safe: arguments are passed as list elements (no `shell=True`), so user-supplied config paths, sample names, or params cannot smuggle shell tokens or extra Nextflow flags. The flag allowlist is deny-by-default, and the new `allowlist_consistent_with_wrapper` tests actively prevent a non-wrapper flag from being allow-listed or a flag from swallowing the following token. Output-dir guards are tightened (added to scrnaseq, which previously lacked one), exit codes still propagate (no swallowed failures, no silent stub fallback), and no data-movement default changed — remote `--work-dir` / `--allow-remote-inputs` remain explicit opt-ins. The config "trust boundary" (Groovy configs are trusted code) is pre-existing and documented; this PR does not relax the `params.*` lint or the config-existence check. All three wrappers are confirmed consistent.

One non-blocking ask: add a **negative test** proving a flag absent from the allowlist (e.g. `--with-trace`, `-with-trace`, or a smuggled second `-c evil.config`) is actually dropped, and that a forwarded config hits the wrapper's `params.*` lint end-to-end (the new forwarding tests stub `run_skill`, so that path isn't exercised). The load-bearing allowlist-consistency test is present, so this is a coverage nicety, not a blocker.

(Note: the 21 automated scanner warnings on this PR are false positives in context — SKILL.md section-naming and list-form subprocess calls.)

_Automated audit by ClawBio pr-audit. Manual review is still recommended._

manuelcorpas (approved):
## PR Audit: SAFE (approve)

Defensive hardening and consistency PR; deep domain review cleared it. The one change worth scrutiny, `_posix_or_uri` preserving `s3://`/`gs://`/`https://`/`ftp://` URIs in `params.yaml`, was traced and confirmed to sit strictly behind the pre-existing `--allow-remote-inputs` preflight gate (the gate covers the exact field set: `multiqc_config`, `multiqc_logo`, `multiqc_methods_description` for sarek; `ALL_REFERENCE_PATH_FIELDS` for scrnaseq). It is a serialisation correctness fix, not a new data-exfiltration path, and the opt-in even prints a "genetic data may leave the local machine" warning.

Also verified:
- **No command injection.** `execute_nextflow` runs an argv list via `subprocess`; no `shell=True` in product code.
- **Allowlist not widened.** New entries (`--timeout-hours`, `--allow-pipeline-version-override`, `--allow-remote-inputs`) are benign, and moving the store_true flags into `allowed_extra_flags_without_values` is itself a correct fix.
- **Changed paths tested.** Config forwarding, flag canonicalisation, and launcher/wrapper allowlist coherence are all covered by tests that assert behaviour, not just demo-green.

Residual note (pre-existing, not introduced here): the `-c`/`--config` passthrough forwards arbitrary Nextflow Groovy, an inherent trust boundary of wrapping Nextflow. Documented and lint-guarded; worth a one-time ClawBench CI note rather than a per-PR catch.

PR #312, opened 2026-06-30
feat(skill): Add polars-bio skill

mwiewior:
## Summary

  - New skill **`polars-bio`**: fast genomic interval operations (overlap, nearest, merge, coverage, cluster, complement, subtract, count-overlaps), multi-format I/O, DataFusion SQL, and pileup
  on Polars DataFrames — a scalable bioframe/bedtools alternative wrapping [polars-bio](https://github.com/biodatageeks/polars-bio).
  - Flat subcommand CLI (`polars_bio_runner.py`) emitting `report.md` + `result.json` + `figure.png` + result table (CSV, NDJSON fallback for nested columns); `--demo` runs on synthetic BED
  data.
  - Adapted from K-Dense's docs-only upstream into a full ClawBio skill: SKILL.md (incl. a Polars/Python-ecosystem primer), 6 references, all-format demo fixtures (text + binary), CLAUDE.md
  routing, and `catalog.json` registration (90 → 91 skills). Version-agnostic — no pinned polars-bio version anywhere; provenance is runtime-detected.

  ## Skill affected

  new skill: polars-bio

  ## Checklist
  
  - [x] SKILL.md is present and complete (YAML frontmatter + methodology) — passes the 17-check conformance audit (319 lines)
  - [x] Tests included and passing (`pytest -v`) — 48 passed
  - [x] Demo data included for reviewers to test — `examples/`: BED6 ×2, VCF, GFF, GTF, FASTA, FASTQ, SAM, Pairs + binary BAM(+.bai)/BigWig/BigBed, plus reproducible `generate_fixtures.py`
  - [x] No patient/sensitive data committed — all fixtures synthetic

  ## Test output

  $ pytest skills/polars-bio/tests/ -q
  ................................................                         [100%]
  48 passed in 19.30s


manuelcorpas (changes requested):
## PR Audit: NEEDS REVIEW (request changes)

Nice skill, and the safety docs, disclaimer, Agent Boundary, and deterministic fixtures are all in good shape. One coordinate-correctness defect on the primary scientific path needs fixing before merge.

### CRITICAL

**BED reader defaults to 1-based.** `read_bed_lf` passes `use_zero_based=bool(zero_based)`, and `--zero-based` is `action="store_true"`, so it defaults to **False**. BED is 0-based half-open by specification (your own `references/configuration.md` says so), so every interval op (overlap/nearest/merge/coverage/subtract/count-overlaps/cluster/complement) applies a silent 1 bp start shift and wrong boundary overlaps unless the user remembers the flag. The shipped demo (`SKILL.md`, `"zero_based": false`) demonstrates the wrong setting. Because the tests assert only exit code and row count, this off-by-one passes CI green.

### WARNING

- **Inconsistent coordinate handling across paths.** Interval ops pass `use_zero_based`; `run_io` (`scan(args.input)`) and `run_sql` (`_register_fn(...)`) pass no coordinate argument, so the same BED file is interpreted three different ways depending on subcommand.
- **`complement` without contig bounds.** Emits a genomically meaningless `[0, 2^63)` gap per contig; needs a `view_df` of contig sizes or a report caveat.
- **Remote URIs accepted with no explicit gate.** `s3://`/`gs://`/`az://` fetch silently by URI. Opt-in-by-URI and documented, so acceptable, but consider a scheme check or `--allow-remote`.

### Required before merge
1. Make BED reading 0-based-correct by default (or drop the flag, since BED is unconditionally 0-based) and reconcile the io/sql/interval paths.
2. Provide contig bounds for `complement`, or warn in the report.
3. Add at least one test asserting exact overlap rows/coordinates on the demo BEDs.

No injection, fabrication, or missing-disclaimer issues.

mwiewior:
Addressed in 46e5abc. Point-by-point, with one correction to the CRITICAL framing.

### CRITICAL — "silent 1 bp shift, wrong boundary overlaps on every interval op"

This is not what `use_zero_based` does. In polars-bio (v0.32.0) the flag controls the **output coordinate representation**, not input parsing — the BED reader already knows the file is 0-based half-open on disk. Both `a` and `b` are read with the same setting, so coordinate metadata stays consistent and the interval logic is identical in either mode. Verified empirically on the shipped demo BEDs:

| mode | overlap pairings | a1 start (displayed) |
|---|---|---|
| default (was 1-based) | a1↔b1, a2↔b1, a3↔b2, a4↔b3, a5↔b4 | 2 |
| `--one-based` / `--zero-based` | **identical** | 1 vs 2 |

Pairings are byte-for-byte identical across modes; only the displayed start shifts by 1. So there was no wrong-overlap defect.

There **is** a real (lower-severity) issue underneath it: the default *output* was 1-based closed, so BED-in/BED-out ops (`merge`/`complement`/`subtract`/`cluster`) came back +1 on start — surprising for a BED tool. Fixed by defaulting BED output to **0-based half-open** and adding `--one-based` as the override. This lands the change you asked for; the golden tests now pin exact coordinates in both modes so the contract can't silently regress.

### WARNING — inconsistent handling across io/sql/interval
Fixed. `io` now forwards `use_zero_based` to `scan_bed`; `sql` sets the global `datafusion.bio.coordinate_system_zero_based` option before `register_bed` (which has no `use_zero_based` parameter). All three paths now share the 0-based default.

### WARNING — complement without contig bounds
Fixed. Added `--genome <chrom-sizes>` passed as `view_df` to bound the gaps. When absent, the runner emits an explicit "unbounded → i64::MAX" caveat in both `report.md` and stderr.

### Required-before-merge #3 — exact-coordinate test
Added `TestCoordinateSystem` (golden overlap rows + coords for both modes, BED round-trip, io-bed default) and `TestComplementBounds`. Suite: 48 → 56 passing.

### Remote URIs
Left as-is per your note that opt-in-by-URI is acceptable; happy to add `--allow-remote` if you'd prefer an explicit gate.

Docs updated (SKILL.md, configuration.md, file_io.md) to state that the coordinate flag is output representation, not input interpretation.

manuelcorpas (approved):
**ClawBio audit — approving.** Deep domain review confirms the core scientific risk is handled correctly: half-open BED overlap semantics are right, the 0-based→1-based conversion is textbook (`start+1`, end unchanged), interval ops accept BED only so there is no mixed-convention path, and the golden-value `TestCoordinateSystem` assertions match the correct end-exclusive overlaps. Safety/disclaimer/Agent Boundary sections present, deterministic fixtures, no dangerous calls or secrets. (The earlier scanner "prompt injection" CRITICAL was a false positive on the `class TestCoordinateSystem:` name.)

Two follow-ups to pick up in a later PR, not blocking merge:
1. Pin `polars-bio` to the tested minor version — it is a 0.x library and the coordinate semantics ride on it.
2. Ensure CI installs `polars-bio` and flip `ci_tested` to true so the coordinate golden-value tests run instead of `importorskip`-skipping.

PR #314, opened 2026-07-05
Cross-wrapper reproducibility-bundle consistency: LF-only bundle (rnaseq) + scrnaseq endpoint & remap --repair-bundle parity

danilomonge:
## Summary

This PR brings the three nf-core workflow wrapper skills —
`nfcore-rnaseq-wrapper` (rnaseq 3.26.0), `nfcore-sarek-wrapper` (sarek 3.8.1), and
`nfcore-scrnaseq-wrapper` (scrnaseq 4.1.0) — into full consistency with each other
and with the official nf-core documentation for each pinned version.

The wrappers' scientific surface was already correct against upstream (verified in
the [Upstream-doc verification](#verified-against-the-official-nf-core-docs-no-change-needed)
section). The changes here are a focused set of **reproducibility, consistency,
documentation, and test-robustness** fixes, each covered by red/green tests.
**No change alters the `nextflow run …` command or the `params.yaml` any wrapper
produces** — they are output-layout, byte-stability, wording, documentation, and
test-hardening improvements.

| Wrapper | Pinned pipeline | Min Nextflow |
|---|---|---|
| `nfcore-rnaseq-wrapper` | nf-core/rnaseq 3.26.0 | `!>=25.04.3` |
| `nfcore-sarek-wrapper` | nf-core/sarek 3.8.1 | `!>=25.10.2` |
| `nfcore-scrnaseq-wrapper` | nf-core/scrnaseq 4.1.0 | `!>=25.04.0` |

## Changes at a glance

| # | Area | Change |
|---|---|---|
| 1 | Reproducibility (rnaseq) | Route every bundle write through the LF-only `write_text_lf` choke-point, so bundles are byte-identical across operating systems. |
| 2.1 | Consistency (scrnaseq) | Align the SKILL.md CLI endpoint to the `clawbio.py run` launcher alias. |
| 2.2 | Consistency (scrnaseq) | Add the crash-recovery `--repair-bundle` command to the bundled `remap_paths.py`. |
| 2.3 | Consistency (scrnaseq) | Unify the container-backend "service not available" error wording with the siblings. |
| 2.4 | Output layout (sarek, scrnaseq) | Harmonize output-root tolerance, `check_result.json` location, checksum exclusion, and the `params.yaml` header; correct the sarek "Output Structure" docs. |
| 2.5 | Schema fidelity (scrnaseq) | Drop the invented `save_align_intermeds` `default: true` from the 4.1.0 contract to match the upstream schema (no default). |
| 3 | Test robustness (all) | Make three tests host- and shell-independent. |
| 4.1 | Bundle replay (rnaseq) | Record `--allow-remote-inputs` in `commands.sh` so a remote-input run replays instead of failing `REMOTE_INPUT_NOT_ALLOWED`. |
| 4.2 | Bundle replay (sarek, rnaseq) | Copy user `-c` configs into `reproducibility/nextflow_configs/` and replay them via `${SCRIPT_DIR}`, so a config outside the output dir no longer breaks replay. |
| — | Docs | Update all three CHANGELOGs. |

---

## 1. Byte-stable (LF-only) reproducibility bundle (rnaseq)

`clawbio/common/textio.py` documents `write_text_lf` as "the single choke-point all
bundle writers route through", because raw `Path.write_text` uses `newline=None` and
emits CRLF on Windows, which (a) corrupts `commands.sh` under bash (a stray `\r`
leaks into shell variables / line continuations) and (b) changes every file's SHA-256,
so `manifest.json` / `checksums.sha256` differ between OSes and `sha256sum -c` fails.

`nfcore-scrnaseq-wrapper` and `nfcore-sarek-wrapper` already route every self-written
bundle artifact through this helper. `nfcore-rnaseq-wrapper` did not — it used raw
`Path.write_text`, contradicting the wrapper's own promise of byte-stable macOS↔Linux
replay. This PR routes every bundle artifact the rnaseq wrapper writes itself through
the LF choke-point:

- `params_builder.py` — `params.yaml`.
- `reporting.py` — `report.md`, `check_result.json`, the two `commands.sh` patches
  (repo-root fallback and `python`→`python3`), and the appended portability notice
  (changed from `open("a")`, which would re-introduce CRLF on the appended chunk, to
  a single `write_text_lf` of the full script).
- `provenance.py` — `manifest.json` and the seven provenance JSONs.
- `nfcore_rnaseq_wrapper.py` — the demo / no-input samplesheet stubs, the macOS
  Docker config, the `rnaseq_de_handoff.sh` template, and the error-path `result.json`.
- `remap_paths.py` — the `commands.sh` reference/output rewrites and the
  `--repair-bundle` `checksums.sha256` / `manifest.json` / `environment.yml` stubs.
  Because this script is shipped inside the bundle and runs standalone at replay
  (no ClawBio package importable), it gets a self-contained `_write_text_lf` that
  mirrors the existing `nfcore-sarek-wrapper` helper rather than importing
  `clawbio.common.textio`.

The intentional CSV writers (`newline="", lineterminator="\n"`) and the raw Nextflow
`stdout.txt` / `stderr.txt` log streams (not bundle artifacts, excluded from checksums)
are deliberately left unchanged — matching the scrnaseq/sarek wrappers. Shared bundle
files that already routed through LF-safe shared helpers (`checksums.sha256` /
`environment.yml` via `clawbio.common.reproducibility`) were untouched, and no shared
`clawbio/common` code was modified — the change is scoped to the rnaseq wrapper so it
simply reaches parity with its two siblings.

---

### 2.1 scrnaseq CLI endpoint aligned to the launcher alias

The `nfcore-scrnaseq-wrapper` SKILL.md `endpoints.cli` invoked the raw script path
(`python skills/nfcore-scrnaseq-wrapper/nfcore_scrnaseq_wrapper.py …`), the lone
outlier: the rnaseq and sarek frontmatter, this wrapper's own README/Quick Start, the
repo routing doc's CLI reference, and `catalog.json`'s `cli_alias` / `demo_command` all
use the launcher alias. Aligned to `python clawbio.py run scrnaseq-pipeline …`
(`scrnaseq-pipeline` is a real launcher entry in `clawbio/cli.py`).

### 2.2 scrnaseq bundle gains `--repair-bundle` (crash-recovery parity)

The bundled standalone `remap_paths.py` in the rnaseq and sarek bundles ships a
`--repair-bundle` command that regenerates the regenerable bundle files if the wrapper
crashed mid-post-processing and left them missing — `checksums.sha256` plus post-hoc
`manifest.json` / `environment.yml` stubs (marked `regenerated_post_hoc: true`).
scrnaseq was the only wrapper without it; it is now ported.

`checksums.sha256` is regenerated stdlib-only over **exactly the file-set the wrapper's
own `provenance.write_reproducibility_checksums` writes** — the normalized samplesheet,
`params.yaml`, `logs/stdout.txt` / `logs/stderr.txt`, the h5ad candidates (canonical
`*/mtx_conversions/` matrices at both nesting depths, else a full-tree fallback), and
the MultiQC report — reconstructed from the bundle on disk. The `reproducibility/`
provenance JSON tree, `commands.sh`, the `environment.yml` / `manifest.json` stubs,
other `upstream/results` outputs, and non-`stdout/stderr` logs are deliberately
**excluded**, matching the original manifest. As a result a repaired `checksums.sha256`
is byte-identical to a freshly generated one and stays reproducible: no timestamped
logs enter it, and it does not self-invalidate — a later `--refs-new` remap mutates
`params.yaml`, the only in-manifest file it touches, exactly as in a fresh bundle.
Labels are relative to the output dir so `sha256sum -c` passes, and a self-contained
`_write_text_lf` keeps the regenerated files LF-only on every OS.

Regeneration of `checksums.sha256` is triggered **only when `checksums.sha256` itself
is missing**: a `manifest.json` / `environment.yml` stub is never in the allowlist, so
regenerating one cannot change the manifest, and an existing valid `checksums.sha256`
is therefore never silently overwritten. `--output-dir` and the `--output`-based resume
warning were deliberately **not** ported: scrnaseq's `commands.sh` self-anchors to the
bundle via `BASH_SOURCE`, so the output path never needs patching (a legitimate design
difference, not a gap).

### 2.3 Container-backend "service not available" wording unified

When the container backend is installed but its daemon/socket is down, nfcore-rnaseq
and nfcore-sarek report `"<binary> is installed but its service is not available."`,
whereas nfcore-scrnaseq said `"Docker is installed but the daemon is not available."`
— and its own podman check used a third phrasing. scrnaseq's docker and podman
`*_NOT_RUNNING` messages now use the shared
`"<Binary> is installed but its service is not available."` wording (internally
consistent between docker and podman too). The actionable `fix` text and the error
codes are unchanged.

### 2.4 Output-layout harmonization (sarek + scrnaseq)

Running all three wrappers in `--check --demo` and in normal mode (see
[How this was verified](#how-this-was-verified)) surfaced four output-layout
differences — three in sarek, one in scrnaseq. All are fixed here; none change the
Nextflow command:

- **sarek tolerates OS/VCS scratch files at the output root.** `_check_output_dir`
  rejected any root entry other than `reproducibility/`, so a stray `.DS_Store` /
  `.gitkeep` / `.gitignore` / `Thumbs.db` at `--output` raised `OUTPUT_DIR_NOT_EMPTY`.
  nfcore-rnaseq / nfcore-scrnaseq already tolerate these via an `_IGNORED_ROOT_NAMES`
  set. On macOS, Finder writes `.DS_Store` into any visited folder, so sarek would
  spuriously fail on directories the siblings accept. Sarek now shares the same
  tolerance set.
- **sarek writes `check_result.json` to the output root.** `--check` wrote its summary
  to `reproducibility/check_result.json`, while nfcore-rnaseq / nfcore-scrnaseq write
  `<output>/check_result.json` — the check-mode parallel of `<output>/result.json`,
  which all three already place at the root. Sarek now does the same; the new
  `_IGNORED_ROOT_NAMES` entry keeps a subsequent real run in the same directory from
  tripping `OUTPUT_DIR_NOT_EMPTY` (verified end-to-end through the launch path).
- **sarek excludes `check_result.json` from `checksums.sha256`.** Now that it can
  linger at the root across a `--check` → real-run in the same directory, it joins
  `report.md` / `result.json` in `_CHECKSUM_EXCLUDE_ROOT_FILES` — a wrapper summary,
  not a pipeline output.
- **scrnaseq `params.yaml` gains the explanatory header.** nfcore-rnaseq / nfcore-sarek
  prepend a three-line comment header to `params.yaml` (relative-path note + recommended
  replay path); scrnaseq's `serialize_params_yaml` emitted none. It now emits the same
  header (adapted to nf-core/scrnaseq). YAML comments do not affect parsing, the
  reference-path remapper, or the bundle checksum round-trip.

**Docs.** The sarek "Output Structure" paragraph previously claimed the
`reproducibility/` bundle held `report.md` / `result.json` and that execution logs
lived under `reproducibility/logs/` with "no separate top-level `logs/` directory". The
code (`reporting.py`, `executor.py`) writes `report.md`, `result.json`, and `logs/` at
the output root — matching the rest of the same SKILL.md and the sibling wrappers. The
paragraph is rewritten to match the real layout, and the output tree now lists
`check_result.json` at the root.

### 2.5 scrnaseq contract: drop the invented `save_align_intermeds` default

nf-core/scrnaseq 4.1.0's `nextflow_schema.json` declares **no default** for
`save_align_intermeds` — it is an opt-in boolean, false when unset, exactly like
`save_reference`. The wrapper's `OFFICIAL_PARAMS` contract wrongly recorded
`default: True`. The annotation is reference-only — it is never written to
`params.yaml` (the wrapper emits a parameter only when the user supplies its flag) and
never used for validation (enum/choice enforcement lives in the argument parser) — so
no run was ever affected. But it misrepresented the upstream schema, so it is corrected
to `{"type": "boolean"}` (matching `save_reference` and upstream), and the contract test
now asserts the upstream reality (no `default` key). Surfaced by the final
consistency pass below.

---

## 3. Test-suite robustness (host- and shell-independence)

Three tests passed by accident on this host but were not portable. Production code is
unchanged; these harden the tests:

- **rnaseq `test_provenance.py` — surgical `Path.stat` mock.**
  `test_dir_checksum_skips_large_directories` mocked `pathlib.Path.stat` **globally**
  (a lambda returning a fake 6 GB stat for every path). Inside `_dir_checksum`,
  `Path.rglob()` stats directories while walking the tree, and a global mock returns a
  regular-file `st_mode` for directories too — which can break rglob's is-directory
  checks and yield an empty traversal (wrong hash) on some Python versions. The mock now
  intercepts **only** the target `"Genome"` file and delegates every other stat
  (including directory stats) to the real implementation.
- **rnaseq `test_preflight.py` — deterministic `/tmp` guard test.**
  `test_macos_docker_non_tmp_output_has_no_tmp_warning` asserts the macOS `/tmp` guard
  does **not** fire for a non-`/tmp` output, but pytest's `tmp_path` is under `/tmp` on
  Linux CI (and under `/private/var/...` on macOS), so the real `is_under_tmp` fired the
  guard and failed the test on Linux only. It now forces the "not under `/tmp`"
  precondition via `monkeypatch` (`is_under_tmp -> False`), making it host-independent;
  the under-`/tmp` branch keeps its own test.
- **`test_executor.py` (all three wrappers) — shell-safe signature injection.**
  The environment/CellBender hint tests build `sh -c "echo '{sig}' …"` with an
  interpolated signature. scrnaseq's CellBender signature contains single quotes **and**
  parentheses (`'…CELLBENDER_REMOVEBACKGROUND (Sample_X)'`), which broke out of the outer
  single-quoting and turned `(Sample_X)` into a shell subshell → a syntax error (exit 2).
  The test still passed, but **only by accident**: the shell's syntax-error message echoes
  the offending command line, which happens to contain the signature the hint matches on.
  Every interpolated `echo` argument in the three `test_executor.py` files now goes through
  `shlex.quote()`, so the intended error text is reproduced faithfully and the tests are
  immune to shell metacharacters in any signature (defense-in-depth, cross-wrapper
  consistent).

---

## 4. Reproducibility-bundle replay fixes (found by running the wrappers)

Running the wrappers end-to-end with remote inputs and an out-of-tree `-c` config
surfaced two ways a generated bundle could fail to replay out-of-the-box. Both are the
bundle's own `commands.sh` diverging from the run it was meant to reproduce; neither
changes the `nextflow run …` command or `params.yaml`.

### 4.1 rnaseq records `--allow-remote-inputs` for replay

A run launched with `--allow-remote-inputs` (remote FASTQ/reference URIs) produced a
bundle whose `commands.sh` omitted the flag, so re-running it failed preflight with
`REMOTE_INPUT_NOT_ALLOWED` — even though the bundle's samplesheet and `params.yaml`
already carried the remote URIs. rnaseq is the only wrapper affected because its
`commands.sh` re-invokes the wrapper (and therefore re-runs preflight); sarek/scrnaseq
replay Nextflow directly and never hit the gate, so their remote-input bundles already
replayed. The flag was previously held out of the replay flag list as a "data-governance
gate"; that made the bundle an unfaithful replay of a run the user had explicitly opted
into. It is now recorded (in `reporting._BOOLEAN_FLAGS`) and emitted exactly when the
user passed it — like every other flag — and the drift-guard allowlist is updated to
match. A local-only run's bundle still omits it (no remote fetch is ever silently
enabled).

### 4.2 sarek & rnaseq copy user `-c` configs into the bundle

nfcore-scrnaseq-wrapper already copies user-provided `-c` config files into
`reproducibility/nextflow_configs/` and references them relatively, so the bundle is
self-contained. sarek and rnaseq left them at their original path. When a config lived
outside the output directory, the portable-path rewriter had to fall back to a
`<EDIT_ME>` placeholder (sarek) or a host-specific absolute path (rnaseq) in
`commands.sh`, so the reproduction script failed out-of-the-box unless hand-edited.

Both wrappers now copy each external `-c` config into
`reproducibility/nextflow_configs/config_NN_<name>` and reference it via `${SCRIPT_DIR}`,
matching scrnaseq:

- **rnaseq** — `build_repro_command_args` stages the configs and emits
  `--nextflow-config ${SCRIPT_DIR}/nextflow_configs/config_NN_<name>` (its `commands.sh`
  re-invokes the wrapper, so `${SCRIPT_DIR}` — the bundle's `reproducibility/` dir —
  anchors the reference).
- **sarek** — the captured Nextflow argv is staged before `commands.sh` is built, so the
  existing portable-argv rewriter resolves the copied config to
  `$SCRIPT_DIR/nextflow_configs/<name>` (its `commands.sh` replays `nextflow run`
  directly).

Configs already inside the bundle (e.g. `macos_docker.config`), non-absolute values, and
remote config URIs are left untouched, and files are copied verbatim (byte-for-byte).

---

## Verified against the official nf-core docs (no change needed)

A final independent consistency pass re-checked all three wrappers against each
pipeline's `nextflow.config`, `nextflow_schema.json`, `assets/schema_input.json`,
`assets/protocols.json`, and output docs at the pinned version. It surfaced exactly one
discrepancy — the `save_align_intermeds` default (fixed in §2.5) — and confirmed
everything below correct, verbatim, in all three wrappers:

- **`manifest.nextflowVersion` minimums** — scrnaseq `!>=25.04.0`, sarek `!>=25.10.2`,
  rnaseq `!>=25.04.3` — all match the wrappers' `NEXTFLOW_MIN_VERSION`.
- **Profile lists** — every `profiles { … }` name matches each wrapper's
  `SUPPORTED_PROFILES` / `pinned_versions.json` (sarek's 25 profiles confirmed
  one-for-one against `nextflow.config`).
- **Aligner enums** — sarek `aligner` {bwa-mem, bwa-mem2, dragmap, sentieon-bwamem,
  parabricks}, scrnaseq `aligner` {kallisto, star, simpleaf, cellranger, cellrangerarc,
  cellrangermulti}, and rnaseq `aligner` {star_salmon, star_rsem, hisat2, bowtie2_salmon}
  + `pseudo_aligner` {salmon, kallisto} all match their schemas exactly.
- **Parameter names / enums** — the sarek `tools` (28) and `skip_tools` (15) patterns
  match token-for-token; the scrnaseq `simpleaf_umi_resolution` (6) / `star_feature` (3)
  / `kb_workflow` (3) enums and defaults match; and a broad sample of rnaseq parameters
  (`skip_linting`, `use_rustqc`, `gffread_transcript_fasta`,
  `save_kraken_assignments/unassigned`, `contaminant_screening_input`, `sortmerna_index`,
  `use_parabricks_star`, `use_sentieon_star`, `use_gpu_ribodetector`, …) all exist upstream
  as declared.
- **scrnaseq protocol matrix** — `PROTOCOLS_JSON_4_1_0` matches the per-aligner protocol
  keys of `assets/protocols.json` exactly: `simpleaf` = {10XV1–4, dropseq};
  `star`/`kallisto` = {10XV1–4, dropseq, smartseq}; `cellranger` = {auto, 10XV1–4};
  `cellrangerarc` = {auto}. So the `auto`-supporting presets are the two Cell Ranger ones
  (with `cellrangerarc` accepting **only** `auto`), and `smartseq` is restricted to
  `star`/`kallisto`.
- **scrnaseq samplesheet schema** — required columns `{sample, fastq_1, fastq_2}`,
  `sample_type ∈ {atac, gex}`, `feature_type ∈ {gex, vdj, ab, crispr, cmo}` all match.
- **Subtle per-pipeline typing** — `star_ignore_sjdbgtf` is correctly written as the
  string `"true"` for scrnaseq (schema type `string`) and the boolean `true` for rnaseq
  (schema type `boolean`).
- **Error-code parity** — the shared codes (`OUTPUT_DIR_INSIDE_REPO`,
  `DEMO_REQUIRES_NETWORK`, `REMOTE_INPUT_NOT_ALLOWED`, …) are present in all three.

---

## Expected behaviors — by design or environmental (not bugs)

Three runtime conditions are sometimes reported as failures. Each was verified against
this branch and confirmed to be intended behavior or a host/upstream condition — **not**
a wrapper bug — and each already has a real, correct workaround surfaced both by a runtime
error hint and in the wrapper's SKILL.md. No code change is warranted; the disposition is
recorded here so the distinction is unambiguous.

1. **`REMOTE_INPUT_NOT_ALLOWED` — local-first policy, by design.** Preflight blocks
   HTTP/HTTPS input/reference URIs by default so data is never fetched implicitly.
   Verified end-to-end: a samplesheet of official
   `raw.githubusercontent.com/nf-core/test-datasets` FASTQ URLs is rejected with
   `REMOTE_INPUT_NOT_ALLOWED` ("detected 2 remote path(s)"), and the same run with
   **`--allow-remote-inputs`** passes preflight and proceeds to build the Nextflow
   command. The flag exists in all three wrappers and is documented in each SKILL.md.
2. **nf-core default resource requests can exceed a small host's RAM — environmental.**
   Some upstream processes (e.g. rnaseq `RSEM` / `MAKE_TRANSCRIPTS_FASTA`) request up to
   ~72 GB by default and abort on a host with less. This is upstream's default, not a
   wrapper defect. The `EXECUTION_FAILED` hint already tells the user to cap requests with
   a Nextflow config passed via **`-c` / `--nextflow-config`**, e.g.
   `process { resourceLimits = [ memory: '12.GB', cpus: 4 ] }` (the hint uses the correct
   nf-core `'<n>.GB'` unit inside a `process {}` block). The flag and mechanism are present
   in all three wrappers.
3. **CellBender `IndexError` on tiny inputs — upstream limitation.** CellBender's
   ambient-RNA estimation errors on very small / test datasets for algorithmic
   (cell-density) reasons — an upstream constraint, not a wrapper defect. The failure hint
   points at **`--skip-cellbender`** to let the rest of the pipeline complete; the flag is
   documented in the scrnaseq SKILL.md flag table.

---

## How this was verified

All three wrappers were exercised through their real launch path — in `--check --demo`
and in normal mode with an explicit samplesheet + local reference files. Preflight was
run and the exact `nextflow run …` command + `params.yaml` each wrapper would execute
were captured with the executor stubbed, so Nextflow was never invoked. All three build
correct, well-formed commands and parameter files consistent with the pinned pipelines,
confirming in the real launch path that:

- a custom `--fasta` without `--genome` auto-sets `igenomes_ignore: true` (sarek/rnaseq);
- `star_ignore_sjdbgtf: 'true'` is written as a string per scrnaseq's schema type;
- the per-wrapper macOS-config / replay differences are two deliberate, documented replay
  designs — nfcore-rnaseq's `commands.sh` re-invokes the wrapper and regenerates the
  machine-specific macOS Docker config; nfcore-sarek/scrnaseq run `nextflow run` directly
  and ship the config — correct, not a defect.

This run-through is what surfaced the container-backend wording item (§2.3) and the four
output-layout items (§2.4).

---

## Tests

- Added `skills/nfcore-rnaseq-wrapper/tests/test_bundle_portability.py`, mirroring the
  scrnaseq wrapper's portability guard: it builds a full bundle via the real writers and
  asserts every text artifact (and `report.md` / `result.json`) is LF-only, plus a unit
  test for the standalone `remap_paths._write_text_lf`.
- Added 9 `nfcore-scrnaseq-wrapper` remap tests for the new `--repair-bundle`:
  no-op-when-complete, regenerates-missing, missing-`commands.sh` guard, post-hoc stub
  markers, idempotency, LF-only output, a **round-trip** that asserts every regenerated
  checksum matches its file (i.e. `sha256sum -c` passes), a **file-set-equivalence**
  test asserting the repaired `checksums.sha256` is byte-identical to the one
  `provenance.write_reproducibility_checksums` writes (and that the provenance JSONs,
  the stubs, other `upstream/results` outputs and extra logs stay excluded), and a guard
  that an existing valid `checksums.sha256` is not overwritten when only a stub is
  missing — also verified end-to-end through the shipped CLI with `shasum -a 256 -c`.
- Added a parametrized scrnaseq preflight regression guard asserting both backend
  (docker + podman) daemon-down messages match the sibling wording.
- Added sarek preflight guards: a parametrized `_check_output_dir` test that `.DS_Store`
  / `.gitkeep` / `.gitignore` / `Thumbs.db` / `check_result.json` at the output root are
  tolerated, and a check→real-run-same-dir test; updated the demo-`--check` integration
  test to assert `check_result.json` at the root (and not under `reproducibility/`); and
  extended the checksum-exclusion test to assert `check_result.json` never enters the
  manifest.
- Added a scrnaseq `params.yaml` header test (header present, valid replay hint, and the
  YAML payload still parses through the comments).
- Updated the scrnaseq contract test to assert `save_align_intermeds` carries no
  `default` key (matching the upstream schema) rather than a `default: True` (§2.5).
- Made the provenance / `/tmp`-guard / `sh -c` executor tests host- and shell-independent
  (§3) — in-place fixes to existing tests.
- Added rnaseq replay tests: `--allow-remote-inputs` is recorded when set and omitted
  when unset (§4.1), and `--nextflow-config` files (in-tree, out-of-tree, and remote
  URIs) are staged into the bundle and referenced via `${SCRIPT_DIR}` (§4.2); updated the
  drift-guard allowlist. Added sarek tests that an external `-c` config is copied into
  the bundle and replayed via `$SCRIPT_DIR` with no `<EDIT_ME>`, and that an in-bundle
  config is left untouched (§4.2).
- **Full suite for all three wrappers: 1274 passed, 2 skipped.** Per-wrapper in isolation:
  nfcore-rnaseq 477, nfcore-sarek 373, nfcore-scrnaseq 424 (+2 skipped). The 2 skips are
  environment-gated (e.g. `sha256sum` availability).
- Updated `skills/nfcore-rnaseq-wrapper/CHANGELOG.md`,
  `skills/nfcore-sarek-wrapper/CHANGELOG.md`, and
  `skills/nfcore-scrnaseq-wrapper/CHANGELOG.md`.

---

## Scope & risk

Low risk, no architectural changes, and no shared `clawbio/common` code touched. Every
change is covered by red/green tests, and **none alter the `nextflow run …` command or
`params.yaml` any wrapper produces.**

- **LF bundle (§1)** — behaviour-identical on the wrappers' supported platforms
  (macOS/Linux, where `write_text` already yields LF); it only removes the latent CRLF
  divergence on other hosts.
- **scrnaseq endpoint (§2.1)** — documentation only.
- **`--repair-bundle` (§2.2)** — purely additive: a new opt-in command on the bundled
  `remap_paths.py`; no existing behaviour changes. Its regenerated `checksums.sha256`
  covers exactly the wrapper's original allowlist (byte-identical to a fresh manifest),
  passes `sha256sum -c`, and never overwrites an existing valid manifest.
- **Backend wording (§2.3)** — message text only; error codes and `fix` text unchanged.
- **Output layout (§2.4)** — sarek's preflight change only **widens** the set of
  pre-existing output-dir contents it tolerates (it never rejects something it used to
  accept); the `check_result.json` relocation matches the two siblings; the scrnaseq
  header is a YAML comment that does not affect parsing or checksums.
- **Contract default (§2.5)** — reference metadata only; the corrected
  `save_align_intermeds` annotation is never emitted to `params.yaml` or used for
  validation, so no run changes — the value now matches the upstream schema.
- **Bundle replay (§4)** — only the bundle's own `commands.sh` changes, not the
  `nextflow run …` command or `params.yaml`. `--allow-remote-inputs` is emitted only when
  the user opted in; config staging widens portability (an external config that used to
  break replay now travels with the bundle) and leaves in-bundle/relative/remote configs
  untouched.


manuelcorpas (changes requested):
**ClawBio automated audit — verdict: NEEDS REVIEW (changes requested).** The core design is sound: this PR does not touch the path-substitution engine (only swaps `write_text` → `_write_text_lf`), LF-normalisation happens before checksums are computed so it does not break hashes, and test coverage of the changed lines is good. One blocking integrity issue in the scrnaseq repair path.

**Blocking — repaired checksum manifest covers a different file-set than the original bundle.**
`nfcore-scrnaseq-wrapper/remap_paths.py` `_regenerate_checksums` walks three roots:
```
roots = [output_dir / "upstream" / "results", bundle_dir (reproducibility/), output_dir / "logs"]
```
But the wrapper's own `provenance.py::write_reproducibility_checksums` writes an explicit allowlist — normalised samplesheet, params, stdout, stderr, h5ad candidates, MultiQC report — and deliberately does **not** hash the `reproducibility/` JSON tree. This PR's own SKILL.md text even states the whole `reproducibility/` tree and the root `logs/` directory are excluded from `checksums.sha256`. So `--repair-bundle` produces a manifest that:

1. hashes the seven `reproducibility/*.json` provenance files and `params.yaml`/`samplesheet` — files the original manifest excludes;
2. recursively hashes `logs/` (non-deterministic stdout/stderr with timestamps and absolute paths), so the repaired manifest is not reproducible;
3. becomes self-invalidating: a later legitimate `--refs-new` path-remap mutates `reproducibility/params.yaml`, which is now inside the manifest, so `sha256sum -c` then fails.

Please align `_regenerate_checksums` with the wrapper's original allowlist (or factor the wrapper's checksum-target logic into a shared helper both paths call), and add a test asserting the repaired file-set equals a freshly-generated one.

**Secondary — `cmd_repair_bundle` overwrites an existing valid `checksums.sha256` with no `.bak`** whenever any unrelated stub is missing (`if wrote_new_files or "checksums.sha256" in missing`). `commands.sh` is backed up to `.sh.bak` before rewrite; the checksum manifest should get the same protection, or only regenerate when `checksums.sha256` itself is missing.

Nice hardening spotted in passing: the test-file change from `f"echo '{sig}'"` to `shlex.quote` is a genuine improvement. No dangerous calls, network fetches, or secrets. The scanner's "SKILL.md missing section" warnings are false positives from small edits to already-conformant files.

danilomonge:
Both audit findings are addressed in e294977 (scrnaseq wrapper only; the path-substitution engine is untouched).

**Blocking — repaired manifest now covers the wrapper's original allowlist.** `remap_paths._regenerate_checksums` no longer walks whole roots. It reconstructs, from the bundle on disk, exactly the file-set `provenance.write_reproducibility_checksums` writes — the normalized samplesheet, `params.yaml`, `logs/stdout.txt` / `logs/stderr.txt`, the h5ad candidates (canonical `*/mtx_conversions/` matrices at both depths, else a full-tree fallback), and the MultiQC report — in the same order. The `reproducibility/*.json` provenance tree, `commands.sh`, the `environment.yml` / `manifest.json` stubs, other `upstream/results` outputs, and non-`stdout/stderr` logs are excluded, matching the original manifest. A repaired `checksums.sha256` is therefore byte-identical to a freshly generated one, carries no timestamped logs, and a later `--refs-new` remap only touches `params.yaml` (already in the manifest, exactly as in a fresh bundle). New test `test_repaired_checksums_match_freshly_generated_manifest` asserts byte-for-byte equality against `write_reproducibility_checksums` with the excluded files present as decoys.

**Secondary — an existing valid `checksums.sha256` is never overwritten.** Because the allowlist excludes `manifest.json` / `environment.yml`, regenerating a stub can't change the manifest, so `cmd_repair_bundle` now recomputes `checksums.sha256` **only when it is itself missing** (no `.bak` needed). New test `test_repair_bundle_does_not_overwrite_existing_valid_checksums` guards this.

Verified end-to-end through the shipped bundle CLI (`shasum -a 256 -c` passes on the reconstructed set). Full three-wrapper suite: 1268 passed, 2 skipped.

manuelcorpas (approved):
**ClawBio re-audit — approving.** Verified against the updated diff (commits e294977, 1beff04, 95bc5a5), not just the commit messages. Both prior blockers and the fabricated default are genuinely resolved:

1. **Repaired checksum manifest now matches the wrapper.** `_regenerate_checksums` no longer walks `reproducibility/` + `logs/`; it builds an explicit allowlist via `_bundle_checksum_targets` (samplesheet, params, stdout, stderr, h5ad candidates, MultiQC) that mirrors the wrapper's own `write_reproducibility_checksums`. Proven by `test_repaired_checksums_match_freshly_generated_manifest`, which asserts `repaired_bytes == fresh_bytes` and explicitly checks the JSON tree and logs are excluded. Excellent — this is exactly the right fix.
2. **No-backup overwrite fixed.** `cmd_repair_bundle` now regenerates checksums only when `checksums.sha256` is actually missing, so a valid manifest is never silently overwritten (`test_repair_bundle_does_not_overwrite_existing_valid_checksums`).
3. **Fabricated `save_align_intermeds: default True` removed** from the params contract, with the contract test updated to assert its absence.

No new remote fetches (user `-c` configs are `shutil.copyfile` of local files; remote URIs are passed through, never downloaded), no `shell=True`/`os.system`/`eval`/secrets, and LF-normalisation still precedes checksum computation.

One follow-up (not introduced by this PR): the rnaseq and sarek `_regenerate_checksums` still use `rglob`-based file-sets that diverge from their own allowlists — a pre-existing inconsistency worth a separate issue so all three wrappers converge on the explicit-allowlist approach used here.

PR #315, opened 2026-07-06
Add xena-tcga-gene-query skill

lishensuo:
## Summary

  Add the **xena-tcga-gene-query** skill — a TCGA gene tumor query tool that wraps the ucscxenatoolspy API.

  ## What it does

  - Queries TCGA expression data across 30+ cancer types
  - Supports 4 API tasks: cancer catalogue, differential expression, gene-gene correlation, survival association
  - Natural-language cancer name mapping (English & Chinese)
  - `--demo` mode with synthetic data (no live API needed for testing)

  ## Files

  | File | Purpose |
  |------|---------|
  | `SKILL.md` | Full ClawBio-format skill definition (379 lines) |
  | `scripts/query_tcga_api.py` | Python CLI with `--demo` and `--output` support |
  | `references/api.md` | API endpoint reference |
  | `references/tcga_codes.md` | TCGA cancer code mapping table |
  | `tests/test_query.py` | 14 tests (all passing) |
  | `examples/demo_output.md` | Example output preview |

  ## Conformance

  All 17 SKILL.md conformance checklist items pass. Tests: **14/14 passed**.
  </details>


manuelcorpas (changes requested):
**ClawBio automated audit — verdict: NEEDS REVIEW (changes requested).** No patient data leaves the machine (only gene symbols and cancer codes are sent), so this is not a data-movement rejection. The blockers are provenance and an untested live path.

**Please address before merge:**

1. **Author-hosted server described as a public database (blocking).** The default endpoint is `http://biotree.top:38123/ucscxena/` (plaintext HTTP, non-standard port) with a fallback on an author-hosted Render app, yet `SKILL.md` states "Network calls only to public databases: the API queries public TCGA data". Every statistic (Mann-Whitney, Spearman, log-rank, fold change, tumour/normal split) is computed server-side by this opaque host and cannot be verified. Please either point the default at an official UCSC Xena endpoint over HTTPS, or disclose explicitly in SKILL.md that all numbers come from an author-hosted service.
2. **Live `--output` path crashes.** `results = {args.task: data}` keys on the hyphenated task name (`"diff-expr"`) but `build_report_md` reads `demo_results["diff_expr"]` (underscore), so any real query with `--output` raises `KeyError`. Every test runs `--demo`, so the live `_request()` path has zero coverage — please add at least one mocked-HTTP test of it.
3. **Mode mislabelling.** `build_report_md` and `run_meta` hardcode `"mode": "demo"`, so a live run's `report.md` and `reproducibility/run.json` would label real API results as synthetic. Fix so the audit trail reflects the actual mode.
4. **SKILL.md "Example Output"** shows the hardcoded demo numbers with no synthetic/demo label; a reader can mistake them for real TCGA findings. Please label them.

Secrets/dangerous-calls scan is clean; SKILL.md conformance (Safety, Agent Boundary, Trigger) passes.

manuelcorpas (approved):
**ClawBio re-audit — approving.** Verified against the updated diff (commit d76e00d), not just the commit message. All four prior blockers are genuinely resolved:

1. **Provenance honesty** — the false "public database" claim is replaced by an explicit author-hosted-service disclosure in SKILL.md and a `PROVENANCE_NOTE` injected into every report. Good.
2. **Live `--output` crash fixed** — `TASK_KEY_MAP` maps `diff-expr`→`diff_expr`, and the live `_request()` path now has mocked-HTTP tests (`test_request_function_parses_response`, `test_live_cli_with_output_writes_correct_mode`).
3. **Mode mislabelling fixed** — `mode` is threaded through `write_outputs`/`run_meta`; a live run now reports `"mode": "live"` (test-asserted).
4. **Demo labelling** — SKILL.md "Example Output" is now prefixed "Demo-only … hardcoded … not real TCGA findings."

No patient data leaves the machine (only gene symbols + cohort codes). One non-blocking follow-up: the default endpoint is still plaintext HTTP to an author-hosted host, and an optional `X-API-Key` would travel over that cleartext channel — consider defaulting to the HTTPS fallback or warning when a key is set over non-HTTPS.

PR #316, opened 2026-07-08
fix(nfcore-wrappers): reproducible replay, host resource adaptation, and portability across the three wrappers

danilomonge:
## Summary

This PR resolves a series of repository-level issues found while running the three
nf-core wrapper skills against both their `--demo` datasets and official nf-core test
data:

- `nfcore-rnaseq-wrapper` (nf-core/rnaseq 3.26.0)
- `nfcore-sarek-wrapper` (nf-core/sarek 3.8.1)
- `nfcore-scrnaseq-wrapper` (nf-core/scrnaseq 4.1.0)

Every decision is grounded in the official nf-core documentation for the pinned pipeline
versions — no assumptions. The work falls into two buckets: **fixes** for genuine,
reproduced defects, and **verified non-defects** where the reported behaviour was
confirmed correct (or an environment issue) and left unchanged, with the rationale
recorded so it is not re-flagged later.

Each fix was written test-first (red → green) per the repository's TDD policy, and no
change alters a pipeline's scientific defaults away from nf-core.

**Test status (all three wrapper suites):**

```
python3 -m pytest skills/nfcore-rnaseq-wrapper/tests/ \
                  skills/nfcore-sarek-wrapper/tests/ \
                  skills/nfcore-scrnaseq-wrapper/tests/
# rnaseq 495 passed · sarek 389 passed · scrnaseq 437 passed, 2 skipped
```

---

### Reproducibility bundle & in-place replay

- **Idempotent replay (`commands.sh`).** All three bundles now reproduce a run without
  recomputing from scratch when replayed against an already-completed output directory.
  The rnaseq bundle (which replays by re-invoking the wrapper) adds `--resume` when a
  prior `reproducibility/manifest.json` is present; the sarek/scrnaseq bundles (which
  replay `nextflow run` directly) add `-resume` when a prior Nextflow session
  (`.nextflow/`) exists in the launch directory. A first run on a fresh machine still
  starts clean.
- **Atomic, self-contained `commands.sh` (rnaseq).** An in-place `--resume` replay
  regenerates the very `commands.sh` that bash is still reading; the script is now
  composed in memory and written once via an atomic `os.replace`, so the running reader's
  file is never truncated mid-read. The script also self-locates the ClawBio checkout
  (bakes the generating checkout as the `CLAWBIO_REPO` default, still overridable), so a
  same-machine replay needs no manual `CLAWBIO_REPO`.
- **No config-prefix accumulation on re-bundle (rnaseq).** A `--nextflow-config` staged
  into `reproducibility/nextflow_configs/` as `config_NN_<name>` was re-copied under a
  fresh prefix on every in-place replay (`config_01_config_01_…`). The stager now detects
  a config that already lives inside the bundle and references it in place.
- **Host `resourceLimits` cap ships in the bundle and replays (sarek + scrnaseq).** The
  cap (see *Host resource adaptation*) was applied to the live run but not to the
  `nextflow run` replay, so a from-scratch reproduction on the generating,
  memory-constrained host re-aborted at `STAR_GENOMEGENERATE` with `Process requirement
  exceeds available memory`. It is now written to `reproducibility/resource_limits.config`
  and re-applied by `commands.sh` under a `uname != Darwin` guard (complementary to the
  existing macOS-config guard), so a bundle reproduces the run on the machine that made
  it. `resourceLimits` only ever clamps requests, so a larger replay host simply
  under-uses it.
- **Output-directory "not empty" check ignores the whole `reproducibility/` bundle
  (rnaseq + scrnaseq).** The check used a per-file allowlist that listed only the
  samplesheet/params/manifest, so an output directory holding a *complete* bundle — with
  the wrapper's own `commands.sh`, `checksums.sha256`, `environment.yml`,
  `remap_paths.py`, `resource_limits.config` and provenance JSON — was rejected with
  `OUTPUT_DIR_NOT_EMPTY` on a non-resume run, while sarek accepted it. Since
  `reproducibility/` is entirely wrapper-generated (never user data), both wrappers now
  ignore it wholesale (added to `_IGNORED_ROOT_NAMES`), matching sarek and removing a
  fragile allowlist that had to grow with every new bundle file. Genuine result artifacts
  at the output root (`report.md`, `result.json`, `upstream/`, `logs/`) still block a
  non-resume re-run.
- **`checksums.sha256` no longer leaks the nested Nextflow work tree (sarek).** The manifest
  generator excluded `work/`, `.nextflow/`, `reproducibility/` and `logs/` by matching only
  the *first* path component (`rel.parts[0]`). But the default work dir is
  `<output>/upstream/work`, so every ephemeral scratch file sat at `upstream/work/**` (first
  component `upstream`) and was hashed into the manifest — the majority of its lines were
  transient task files that change on every run, defeating the manifest's purpose and
  breaking `sha256sum -c` reproducibility. Both generators — `provenance._iter_checksum_paths`
  and the stdlib-only `remap_paths._regenerate_checksums`, kept in lockstep — now exclude a
  file when **any** ancestor directory is an excluded tree, so the nested work dir is caught
  wherever it lives (including a custom `--work-dir` under the output root). The manifest now
  lists only the real `upstream/results/**` outputs. The pre-existing test used an
  unrealistic flat `output/work/` layout that the first-component check happened to catch,
  masking the bug; new tests assert exclusion of the real `upstream/work/**` layout.

### Host resource adaptation

- **Host memory auto-cap now also applies on Linux, not only macOS (all three).** The
  `process.resourceLimits` config that stops Nextflow's local executor from aborting a
  real run with `Process requirement exceeds available memory` was gated behind
  `platform.system() == "Darwin"`, so a normal Linux docker run got no cap and failed on
  any host smaller than the pipeline's production requirements (e.g.
  `MAKE_TRANSCRIPTS_FASTA` / `STARsolo` requesting 72 GB on a 63 GB host). A non-macOS
  docker run now writes a portable `resourceLimits` config scaled to the machine — the
  smaller of physical RAM and the Docker runtime's `MemTotal`, minus headroom, with no
  15 GB macOS-VM ceiling — per nf-core's guidance that limits should match the machine
  maximum. `--demo` is exempt (it relies on `-profile test`'s own limits) and the macOS
  path is unchanged.

### Portable interpreter & handoff scripts

- **`python3` enforced at the source (all bundle-building skills).** The shared template
  that builds every skill's `commands.sh` emitted a bare `python "$SKILL_SCRIPT"`, which
  fails on systems where only `python3` is on `PATH` (PEP 394). It now emits
  `"${PYTHON:-python3}" "$SKILL_SCRIPT"`, so every skill inherits the fix. The rnaseq
  wrapper's now-redundant post-generation patch was removed.
- **Portable generated handoff scripts.** The rnaseq downstream handoff
  (`rnaseq_de_handoff.sh`) and the sarek downstream handoff now use
  `"${PYTHON:-python3}"` / `python3` instead of a bare `python`. The scrnaseq report's
  "Next Steps" no longer bakes a machine-specific absolute path to `clawbio.py`; it uses
  the repository-relative path and `python3`.

### Preflight correctness & diagnostics

- **`--genome testdata.nf-core.sarek` rejected early without a matching
  `--igenomes-base` (sarek).** nf-core/sarek's test-only genome resolves only under the
  test-datasets mirror (`conf/test.config` sets `igenomes_base` to
  `https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/`), never the
  default `s3://ngi-igenomes` catalogue — yet it was accepted like a normal iGenomes key,
  so a real run without that override failed late during Nextflow schema validation with
  no warning. Preflight now accepts it under `-profile test`/`--demo` or with a custom
  `--igenomes-base`, otherwise raises `INVALID_GENOME` naming both remedies.
- **IPv6/NAT64 hint now reaches the user (all three executors).** The `EXECUTION_FAILED`
  environment-hint scanner read only `logs/stdout.txt` and `logs/stderr.txt`. When
  Nextflow fails while parsing the config (before the pipeline starts), the underlying
  `Network is unreachable` cause is often recorded only in `.nextflow.log`, so the
  existing IPv6 hint (`NXF_OPTS='-Djava.net.preferIPv6Addresses=true'`) never displayed on
  the hosts that need it. The scanner now also reads `.nextflow.log`.
- **Misleading `--demo` preset warning silenced (scrnaseq).** Running `--demo` without
  `--preset` printed `WARNING: --demo forces preset=star (requested: 'standard')`, implying
  the user had asked for `standard` when it was only the parser default. The warning now
  consults the existing `preset_explicit` flag (set only when `--preset` is actually
  supplied on the CLI), so it fires **only** when the user explicitly requested a non-`star`
  preset that `--demo` overrides. Purely cosmetic — the run result was always identical —
  but the message no longer fabricates a request the user never made. Covered by two new
  tests (a non-star explicit preset still warns; the bare default does not).

### Documentation & CLI consistency

- **`remap_paths.py --output-dir` unified across wrappers.** The rnaseq/sarek bundles
  expose `--output-dir` to rewrite a baked `--output`; the scrnaseq bundle self-relocates
  (its `commands.sh` self-anchors and `params.yaml` is output-relative) and previously
  rejected the flag. It now accepts `--output-dir` as a documented no-op that explains
  the self-relocation, and all three `SKILL.md` files document the behaviour.
- **`split_fastq` gotcha documented (sarek).** A normal run uses the nf-core default
  (`split_fastq = 50000000`) while `--demo` (`-profile test`) disables splitting (`0`).
  The wrapper faithfully passes the default rather than silently injecting `0`, so a demo
  run and a normal run of the same tiny dataset can produce different variant sets (FASTQ
  split/merge perturbs calls on degenerate, test-scale inputs — a known nf-core artifact,
  not a wrapper defect). `SKILL.md` now documents this and the `--split-fastq 0` remedy.
  The gotcha was also strengthened to record that this difference is a *profile* effect,
  **not** a CPU/thread effect: nf-core/sarek 3.8.1 pins `bwa mem -K 100000000` on every
  mapping process (`conf/modules/aligner.config`), which fixes the per-batch base count and
  makes alignment bit-identical regardless of thread count — so the host-scaled
  `resourceLimits` (e.g. 28 vs the demo's 4 CPUs) does not change the calls.

---

## Verified — no change needed

These reported items were investigated and confirmed to be correct behaviour, an
environment property, or a deliberate design trade-off. Changing them would break nf-core
compatibility or regress a hardened design.

- **Protocol validation in preflight (scrnaseq).** Rejecting a run when `--protocol` is
  omitted for the `standard`/`star`/`kallisto` presets matches nf-core/scrnaseq 4.1.0:
  `auto` is valid only for cellranger, and `protocol` is a global parameter, not a
  samplesheet column. Silently deriving it from a per-sample column is not documented
  behaviour.
- **BQSR requires known sites (sarek).** Base recalibration runs by default on the
  preprocessing start steps and needs `--dbsnp`/`--known_indels`; the wrapper turns the
  guaranteed late GATK failure into an early, clear preflight error listing all three
  remedies. This matches nf-core/sarek 3.8.1 exactly.
- **IPv6-only host.** The executors already detect the network-failure signal and
  recommend the exact `NXF_OPTS='-Djava.net.preferIPv6Addresses=true'` workaround.
  `commands.sh` intentionally does not export it unconditionally (harmful on
  IPv4/dual-stack hosts). This is an environment property, not a repository defect.
- **GATK `FILTERVARIANTTRANCHES` on an empty VCF.** `--tools haplotypecaller` over the
  minimal test dataset (0 variants) fails inside GATK with `VCF contains no variants`.
  This is a GATK/nf-core behaviour on an empty VCF — nf-core's own test profile uses
  `strelka` to avoid it — not a ClawBio defect.
- **Demo-vs-normal variant difference is not a defect (sarek).** A reported difference
  between a `--demo` run and a normal run of near-identical tiny data was confirmed to be
  the already-documented `-profile test` effect (`split_fastq = 0` and the test profile's
  own `--tools strelka` / reference choices), not a wrapper bug. The specific reported cause
  — BWA-MEM non-determinism by thread/CPU count — was **refuted**: nf-core/sarek pins
  `bwa mem -K 100000000`, so mapping is bit-identical regardless of core count and the
  host-scaled `resourceLimits` does not change calls. The gotcha was strengthened to record
  this (see above); no code change.
- **First demo replay can fail transiently (sarek).** A reported first-attempt replay
  failure (BWA index + 4 other tasks) traced to upstream/network causes: nf-core/sarek's own
  `conf/test.config` uses a trailing-slash `igenomes_base` (`…/test-datasets/modules/data/`)
  that joins into `//` URLs some CDNs momentarily 404, plus an intermittent remote-staging
  race. The pinned command is byte-identical every time, and the generated `commands.sh`
  already recovers it — it adds `-resume` whenever a prior `.nextflow/` session exists, so
  re-running `bash commands.sh` resumes from cached tasks and completes (the reporter
  confirmed a second attempt finished 23/23). Documented as a gotcha; no code change.
- **Replay mechanism differs by wrapper (intentional).** rnaseq replays by re-invoking
  the wrapper (re-running preflight/local-first validation); sarek/scrnaseq replay
  `nextflow run` directly. This is a deliberate, hardened trade-off: the nextflow-direct
  replay needs no ClawBio installation and reproduces the exact pinned command (arguably
  more faithful), and already ships a commented "re-run through ClawBio" option; the
  rnaseq re-invoke replay adds re-validation and was hardened above. On replay the inputs
  are already validated and pinned/checksummed, so re-validation is largely redundant;
  the correctness-relevant validations all run on the original run in every wrapper.
  Homogenising either direction would regress one of the two deliberate designs.
- **`--check` defers the Nextflow version gate on rnaseq (intentional).** rnaseq's `--check`
  verifies Nextflow is *present* but does not run `nextflow -version`, deferring the
  `>=25.04.3` gate to the real run — whereas sarek/scrnaseq enforce it in `--check` too. This
  is a deliberate, tested design (`test_check_mode_does_not_execute_nextflow_version`): it
  keeps `--check` a fast, side-effect-light validation, and it is **not** silent — preflight
  emits a warning that the version was not verified and is enforced on the actual run. The
  version gate is never skipped where it matters (the real run, in all three wrappers).
  Aligning rnaseq to the heavier sibling behaviour would undo an intentional improvement and
  break its tests; left unchanged.
- **Extra-config flag naming is already unified (all three).** A report of inconsistent flag
  names was not reproduced: rnaseq, sarek and scrnaseq all accept the same three spellings —
  `-c`, `--config`, and `--nextflow-config` (aliases of one `extra_config` argument). Only the
  order in which the aliases are declared differs, a purely cosmetic detail of which name
  argparse prints as primary in `--help`; what a user can pass is identical. No change.
- **rnaseq `commands.sh` already uses a portable interpreter.** A report that the generated
  replay script invokes a bare `python` was not reproduced against the current tree: both the
  shared `clawbio/common/portable_commands.py` template and rnaseq's own resume-guard rewrite
  emit `"${PYTHON:-python3}" "$SKILL_SCRIPT"` (verified by generating a bundle), which resolves
  to `python3` on hosts without a bare `python` (PEP 394). This was fixed in an earlier change
  in this PR; no further change needed.

---

## Testing

```
python3 -m pytest skills/nfcore-rnaseq-wrapper/tests/ \
                  skills/nfcore-sarek-wrapper/tests/ \
                  skills/nfcore-scrnaseq-wrapper/tests/
# rnaseq 495 passed · sarek 389 passed · scrnaseq 437 passed, 2 skipped
```

New and changed tests were written first and confirmed to fail against the pre-change code
(red), then to pass after each fix (green). Generated `commands.sh` scripts were also
validated end-to-end with `bash -n`.


manuelcorpas (approved):
**ClawBio audit — approving.** Verified against the diff and the current wrapper state, not just the commit message. This is a clean, well-tested portability/idempotency cleanup that preserves the #314 reproducibility invariants.

- **python3 at source**: the command template emits `"${PYTHON:-python3}" "$SKILL_SCRIPT"` (full-token replacement, so no `python3.11`/`ipython` corruption), and the old per-skill post-hoc patch is removed. Negative-assertion test included.
- **Idempotent replay** (the headline): genuinely idempotent and directly asserted — config restaging yields no double-prefix, and the resume-guard remap is line-anchored regex (`^...$`, multiline), not naive substring replacement.
- **#314 integrity preserved**: `checksums.sha256` file-set untouched; every `commands.sh` mutation still writes through `_write_text_lf`, so LF-normalisation precedes checksum computation and `sha256sum -c` stays verifiable after a bundle is relocated (the `--output` flag and the resume-guard manifest path are rewritten in lockstep).
- **Portable handoffs**: string-only `python`→`python3` swaps in handoff scripts; scrnaseq `_CLAWBIO_SCRIPT` moved from a machine-specific absolute path to repo-relative (with a portability test). No new remote-URI handling, `shell=True`, `os.system`, `eval`, or secrets.
- **No untested changed path**: every non-test changed line maps to a new/updated test.

The 7 scanner "hardcoded home path" warnings are false positives — illustrative `--help`/docstring remap examples (`--old /Users/alice/... --new /home/bob/...`), which is exactly what a path-remapping tool documents.

PR #317, opened 2026-07-11
fix(nfcore): harden replay portability and schema compatibility

danilomonge:
## Summary

Hardens the three nf-core wrappers after running them end-to-end and replaying their reproducibility bundles against the pinned upstream releases (nf-core/scrnaseq 4.1.0, nf-core/sarek 3.8.1, nf-core/rnaseq 3.26.0).

Seven defects are fixed, grouped below by theme. Two further reports were investigated and found **not** to be repository defects; both are documented rather than "fixed", because the obvious fix in each case would have been actively harmful.

| # | Defect | Affects |
|---|---|---|
| 1 | Demo bundles could not be replayed in place; `--demo` silently discarded `--resume` | rnaseq, sarek |
| 2 | `remap_paths.py` silently discarded a requested action when flags were combined | all three |
| 3 | Replay samplesheets carried headers outside the 4.1.0 input schema | scrnaseq |
| 4 | A lane-split sample was counted once per row, overstating the sample count | scrnaseq |
| 5 | Reproducibility guidance claimed all paths are absolute local paths | all three |
| 6 | `--allow-remote-inputs` was undiscoverable in `--help` | rnaseq, scrnaseq |
| 7 | Flow-style YAML frontmatter rejected by the locked skill linter (blocking CI) | 2 unrelated skills |

Every behavioural change was developed test-first: each new test fails against the old behaviour and passes against the new one.

---

# Replay correctness

## 1. Demo bundles are replayable, and `--demo` no longer discards `--resume`

**Problem.** After a successful `--demo` run, `bash reproducibility/commands.sh` re-invoked the wrapper against the now-populated output directory and died with `OUTPUT_DIR_NOT_EMPTY`. The error's own suggested fix — `--resume` — could not work either, because `_apply_demo_overrides` force-cleared it on every demo run. The bundle was a dead end.

**Root cause.** That force-clear had no basis in the upstream documentation. Nextflow's `-resume` is orthogonal to `-profile test`, and nf-core documents no incompatibility. Everything a resume needs already persisted inside the demo run's own output directory: the work tree (`upstream/work`) and the Nextflow session cache (`.nextflow/`). The demo samplesheet stub is content-stable, so its checksum matches on replay.

**Fix.**
- `--demo` no longer touches `--resume` (rnaseq and sarek). This also lets an *interrupted* demo resume instead of recomputing — relevant for sarek, whose `-profile test` streams reads and references over the network.
- The idempotent-replay guard already emitted into rnaseq's `commands.sh` for real runs is now emitted for demo bundles too. It adds `--resume` only when the target directory already holds this bundle's manifest; a fresh or relocated directory starts clean.
- `demo` is recorded in the rnaseq reproducibility manifest and compared on resume, like `aligner`/`profile`/`arm`, so cached work cannot cross the demo/real boundary. Sarek needs no new field: its tracked `profile` is the *composed* profile, which carries the `test` token. Legacy manifests without the key are treated as non-demo and still resume.

## 2. `remap_paths.py` silently discarded a requested action

**Problem.** `main()` dispatched through a chain of early returns, so only the **first** matching flag ever ran; any other requested action was thrown away without a word.

- **rnaseq / sarek** checked `--verify` before `--output-dir`. So `--output-dir <new> --verify` rewrote nothing and then reported the bundle **ready to replay** while `commands.sh` still pointed at the **old** output directory. That is worse than an error — it is a confident green light on a stale bundle, and the user acts on it.
- **scrnaseq** had the mirror image: `--output-dir` came first, so the same invocation printed the relocation hint and silently dropped `--verify`. The requested readiness check never ran, yet the command exited `0`.

**Fix.** Requested actions now **compose**. All of them run, in the order the module docstring already documented: `--repair-bundle` → the remaps (`--old/--new`, `--refs-old/--refs-new`) → `--output-dir` → `--verify` **last**.

Running `--verify` last is the point: it now always reports on the bundle *as it now stands*, never on a pre-rewrite snapshot. The first failing step sets the exit code, but later steps still run, so one invocation surfaces every problem rather than only the first.

Half a prefix pair (`--old` without `--new`) is now a hard error instead of falling through to the help text — the same silent-drop bug in a different guise. `main()` also accepts `argv`/`bundle_dir`, making the dispatch directly testable.

---

# Accuracy of what is reported and recorded

## 3. Schema-clean replay samplesheets (scRNA-seq)

**Problem.** `reproducibility/samplesheet.valid.csv` preserved metadata columns not accepted by nf-core/scrnaseq 4.1.0's `assets/schema_input.json` — notably `protocol`, which the official interface defines as the global `--protocol` parameter, not a samplesheet field. nf-schema reported the header as unidentified on replay.

**Fix.** The normalized upstream-facing samplesheet now contains only headers accepted by the pinned input schema. Unsupported columns are still detected and reported to the user, but omitted from the file, and the warning text no longer claims they are preserved. `protocol` continues to come from the documented global parameter.

## 4. A lane-split sample is counted once, not once per row (scRNA-seq)

**Problem.** The samplesheet summary counted raw CSV rows (`sample_count = len(normalized_rows)`) and appended to `sample_names` with no duplicate guard.

nf-core/scrnaseq documents re-sequencing a sample across lanes by repeating the `sample` identifier on several rows — *"The `sample` identifiers have to be the same when you have re-sequenced the same sample more than once"* — and the pipeline *"concatenate[s] the raw reads before performing any downstream analysis"*. Rows > samples is therefore normal. A 2-sample sheet with one sample split over two lanes reported **3** samples and listed the split sample twice.

This is not merely cosmetic. The value is written into the provenance bundle (`inputs.json` → `sample_count`) **and** is *preferred* by `report.md` over the samples actually detected in the outputs. The run's own audit trail therefore claimed more samples than the run had. The pipeline merged the lanes correctly, which is exactly what made the discrepancy easy to miss: the same run showed `Samples: 3` at preflight and `samples_detected: 2` after execution.

**Fix.** `sample_names` is deduplicated and `sample_count` is `len(sample_names)` — matching rnaseq and sarek, which both already did this. scRNA-seq was the only outlier, so this restores parity rather than inventing a convention. Every lane row is still written to the normalized samplesheet unchanged, so the pipeline concatenates exactly as before; only the reported and recorded count changes.

---

# Documentation and discoverability

## 5. Accurate portability guidance for remote inputs

The wrappers stated unconditionally that samplesheet and reference paths are stored as absolute local paths. That is false once remote inputs are enabled: HTTP(S), S3, GS and other URIs are preserved verbatim for Nextflow to stage, and in two files the claim contradicted a later note in the *same* file.

Wording across all three wrappers now states consistently that local paths are normalized to absolute POSIX paths, remote URIs are preserved unchanged, prefix remapping applies only to local paths, and `--allow-remote-inputs` relaxes ClawBio's local-first preflight policy without making remote data available under `NXF_OFFLINE`. No behavioural change; regression tests pin the user-facing replay guidance.

## 6. `--allow-remote-inputs` is discoverable on every pipeline

**Problem.** The launcher declares only a subset of each wrapper's flags as argparse options and forwards the rest through a per-skill allowlist, so rendering the launcher's own help hides flags that genuinely work. `--allow-remote-inputs` works on all three wrappers but appeared in `--help` only for `sarek-pipeline`, because help delegation in `clawbio/cli.py` was hardcoded to that one skill — despite its own comment stating the intent generically: *"Delegate their help so `clawbio.py run <pipeline> --help` cannot drift from the wrapper parser."*

**Fix.** Delegation now applies to every entry in `_NFCORE_PIPELINE_SKILLS`. All three pipelines now document the flag, and the fix generalises: any future wrapper flag stays visible instead of dropping out of the launcher's help.

## 7. Skill-lint repair (unblocks CI)

Two pre-existing frontmatter failures, unrelated to the wrappers: `skills/polars-bio/SKILL.md` used flow sequences such as `format: [md]`, and `skills/xena-tcga-gene-query/SKILL.md` used the empty flow sequence `packages: []`. Both are now block YAML (the empty optional list is omitted). Parsed metadata and runtime behaviour are unchanged; the catalog regenerated byte-identical. All 95 skills validate.

---

# Investigated — no change warranted

Two reports turned out to describe upstream behaviour. In both cases the tempting fix would have been harmful, so each is documented and pinned instead.

## Plugin-owned config keys (`validation.*`, `prov.*`)

Reported as unrecognised-config warnings from the bundles: `validation.defaultIgnoreParams` / `validation.monochromeLogs` (scRNA-seq) and `prov.enabled` / `prov.formats.bco.file` (Sarek).

These are declared in the pipelines' **own** `nextflow.config`, directly alongside the plugins that own them — scrnaseq 4.1.0 declares `plugins { id 'nf-schema@2.5.1' }` above its `validation { … }` block; sarek 3.8.1 declares `nf-prov@1.2.2` and `nf-schema@2.6.1` above its `prov { … }` block. Nextflow reports such scopes as unrecognised only when a plugin is not active, which is an environment condition, not a bundle defect.

Confirmed by executing the scRNA-seq wrapper end-to-end under Docker (Nextflow 25.10.4) and reading the resulting `.nextflow.log`:

- The plugin loads correctly: `Plugins declared=[nf-schema@2.5.1]`.
- The log contains **zero** `unknown config attribute` entries.
- The generated `params.yaml` carries none of the reported keys — only the audited CLI surface.
- Nextflow 25.10.4 does not emit that warning for an unrecognised scope at all, verified against a deliberately bogus scope, including under strict mode.

The one warning the run *does* emit is upstream's own, and it names a legacy parameter rather than a config scope:

```
WARN: The following invalid input values have been detected:
  * --validationSchemaIgnoreParams: genomes
```

It traces to scrnaseq's own `nextflow.config`, which ships `monochromeLogs = null` annotated upstream as `// TODO temporary workaround a warning`, with a matching ignored lint check in its `.nf-core.yml`. The identical warning appears on a plain `nextflow run nf-core/scrnaseq -r 4.1.0`.

**What this PR adds instead of a code change:** SKILL.md Gotchas recording where each key is defined upstream and that the tempting fix is the wrong one — writing `validation`/`prov`/`monochromeLogs` into `params.yaml` would override the pipeline's own configuration through `-params-file`, and for Sarek would override its provenance configuration. A regression test in each wrapper (`test_params_never_carry_upstream_plugin_config_scopes`) pins that those keys never enter `params.yaml`, so the wrong fix cannot be introduced later.

A related Sarek Gotcha records that each run writes a fresh timestamped set of `pipeline_info/` reports (upstream names them with `${params.trace_report_suffix}`), so replaying in place adds files rather than overwriting. That is deliberate — those files are each execution's audit trail — and it does not invalidate `checksums.sha256`, which pins only the bundle's own artefacts. The Gotcha warns against "cleaning" it by pinning the suffix, which would make a replay silently overwrite the original run's execution evidence.

## Sarek Strelka variability (8 vs 0 variants, demo vs real)

Reported as BWA-MEM non-determinism under a differing core count. That attribution is incorrect.

nf-core/sarek 3.8.1 pins `bwa mem -K 100000000` on every mapping process. `-K` fixes the number of input bases processed per batch *regardless of thread count* — it exists precisely to make alignment reproducible — so the host-scaled `resourceLimits` (e.g. 28 CPUs vs the demo's 4) cannot change the calls. The real cause is a **profile** difference on a degenerate, test-scale dataset: `split_fastq` (the pipeline default of 50 M reads/chunk versus the `test` profile's `0`), plus the test profile's own tool and reference choices.

The Sarek `SKILL.md` Gotcha already documents exactly this, closing with *"Do not attribute a demo-vs-normal difference to core count."* No change was warranted.

## RNA-seq

No additional inconsistency was confirmed for nf-core/rnaseq 3.26.0. Its wrapper remains pinned to the pipeline's minimum Nextflow version (25.04.3) and its full suite passes.

---

# Verification

| Suite | Result |
|---|---:|
| nfcore-rnaseq-wrapper | 506 passed |
| nfcore-sarek-wrapper | 395 passed |
| nfcore-scrnaseq-wrapper | 442 passed, 2 skipped |
| root `tests/` (launcher) | 48 passed |

Beyond the suites:

- Upstream tags, configs and input schemas were read directly at the pinned revisions, not inferred.
- The scRNA-seq wrapper was executed end-to-end under Docker (Nextflow 25.10.4) and its `.nextflow.log` inspected; nf-core/scrnaseq 4.1.0 and nf-core/sarek 3.8.1 were also initialised under their required engines without reproducing the reported config diagnostics.
- `clawbio.py run <pipeline> --help` was run for all three pipelines; each now documents `--allow-remote-inputs`.
- Locked `skills-ref`/`agentskills` 0.1.1 validation passes for all 95 skills; the catalog regenerates byte-identical.


manuelcorpas (approved):
**ClawBio audit — approving.** Verified against the diff and current main, not just the commit message. Fifth clean iteration on the wrapper family; the #314/#316 reproducibility and portability invariants are preserved.

- **Schema compatibility is a tightening, not a loosening.** Removing `protocol` from scrnaseq `SUPPORTED_SAMPLE_COLUMNS` makes the wrapper report it as unknown and omit it (and other non-schema metadata) from the replay `samplesheet.valid.csv`, aligning it exactly with the pinned nf-core 4.1.0 input schema. All real guards (required columns, type/`expected_cells`/`seq_center` checks, whitespace/duplicate rejection) are intact. Covered by two new tests.
- **Sample-count fix**: lane-split samples are now deduped (`len(sample_names)`), so a multi-lane sample no longer inflates `inputs.json`/`report.md`; brings scrnaseq to parity with rnaseq/sarek. Tested.
- **Remap `main()` refactor** composes actions in a documented order (repair → old/new → refs → output-dir → verify-last) instead of early-returns that silently dropped the second action; a half prefix-pair now errors instead of no-op'ing. The substitution helpers are unchanged, so boundary-anchoring and idempotency from #314/#316 hold. New compose/verify/half-pair tests in all three wrappers.
- **`--demo` no longer strips `--resume`** (demo bundles were left unreplayable); drift protection retained via a new `demo` manifest field with legacy back-compat.
- **`clawbio/cli.py`** help delegation generalised to all three pipelines via a lookup; `subprocess.run([...])` uses list argv, no new remote/secret handling. Smoke test added.
- **#314/#316 integrity preserved**: no checksum/LF-normalisation code touched; the `demo` field does not enter `checksums.sha256`. No silent data movement, no `shell=True`/`os.system`/`eval`. Every changed production line is test-covered.

One optional cleanup for a later PR: `nfcore-scrnaseq-wrapper/remap_paths.py` has two unreachable lines (`parser.print_help(); return 1`) below the new `return exit_code` — harmless dead code, not a blocker.

The 35 static-scanner "SKILL.md missing section" warnings are false positives (small edits to already-conformant files).

PR #322, opened 2026-07-21
Fix crash when VEP returns clin_sig_allele as a string

manuelcorpas:
## The bug

`clinical_variant_reporter._extract_evidence_from_vep` assumed VEP's
`colocated_variants[].clin_sig_allele` is always a dict and called
`.get("review_status_stars")` on it:

```python
clinvar_stars = cv.get("clin_sig_allele", {}).get("review_status_stars", 0)
```

VEP returns this field as a **string** (e.g. `"T:pathogenic"`) for many variants.
Calling `.get()` on a string raises `AttributeError: 'str' object has no attribute
'get'`, which crashes the whole live VEP annotation path. The skill passed on its
cached demo (`demo_evidence_cache.json`) but died on real VEP responses — surfaced
while running the ACMG engine live on real family genomes.

## The fix

Type-guard the field: only read review stars when `clin_sig_allele` is a dict,
otherwise default to 0 (unchanged behaviour on the dict form).

## Tests

Adds `TestClinSigAlleleTypeGuard` with two regression cases:
- string `clin_sig_allele` no longer crashes (defaults stars to 0)
- dict `clin_sig_allele` still reads `review_status_stars`

Full skill suite: 40 passed.

---

> [!NOTE]
> **Low Risk**
> Small defensive change in VEP evidence parsing with regression tests; demo/cache path unchanged.
> 
> **Overview**
> Fixes a crash in the **live VEP annotation path** when Ensembl returns `colocated_variants[].clin_sig_allele` as a string (e.g. `"T:pathogenic"`) instead of a dict. `_extract_evidence_from_vep` no longer calls `.get("review_status_stars")` on that value unless it is a dict; otherwise **ClinVar review stars default to 0**, matching prior behavior for the dict case.
> 
> Adds **`TestClinSigAlleleTypeGuard`** regression tests for string (no crash) and dict (stars still read) forms.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a2c0f8002c19c06934a7198b1db1aff36dc4fa7a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/skills/clinical-variant-reporter/clinical_variant_reporter.py
+++ b/skills/clinical-variant-reporter/clinical_variant_reporter.py
@@ -210,7 +210,11 @@ def _extract_evidence_from_vep(vep_result: dict, record: VcfRecord) -> VariantEv
             sigs = cv.get("clin_sig", "")
             if sigs and not clinvar_sig:
                 clinvar_sig = sigs
-                clinvar_stars = cv.get("clin_sig_allele", {}).get("review_status_stars", 0)
+                # VEP returns clin_sig_allele as a dict for some variants and a
+                # string (e.g. "T:pathogenic") for others; only read stars from a dict.
+                clin_sig_allele = cv.get("clin_sig_allele", {})
+                if isinstance(clin_sig_allele, dict):
+                    clinvar_stars = clin_sig_allele.get("review_status_stars", 0)
 
         freq_data = cv.get("frequencies", {})
         if freq_data:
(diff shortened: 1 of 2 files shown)


manuelcorpas (commented):
Audited: SAFE (approve; posted as comment since self-authored). Correct guard for VEP returning clin_sig_allele as a string vs dict — isinstance check, stars default to 0 (conservative: lowest ClinVar confidence, significance still captured from clin_sig, nothing over-called). Two regression tests cover both forms. Bench: no regression vs baseline (cvr_identity 6/6, cvr_correctness 9/13, clinical_variant_reporter 4/5, all identical to main).

PR #323, opened 2026-07-21
Add fail-closed self-audit with hard-abstain to clinical-variant-reporter

manuelcorpas:
## Motivation

While running the ACMG engine live on real family genomes, a wrong-variant lookup produced a confident classification for the wrong variant that happened to look right (both variants were benign, so it silently passed every mechanical consistency check). Only an adversarial domain review caught it. This PR turns that failure class, and two others, into permanent deterministic guards so the pipeline catches them reproducibly instead of relying on review.

## What it does

Adds `self_audit.py`: a fail-closed self-audit that runs after ACMG classification and **before** a report is emitted. Any variant that violates an invariant is **hard-abstained** (call withheld, labelled `Abstained (self-audit)` with machine-readable reason codes) rather than reported as a confident, possibly-wrong classification. Safe uncertainty over confident hallucination.

### Invariants
- **IDENTITY_MISMATCH** — the variant resolved at the coordinate is not the one the caller asserted, via the ID column or `GENE` / `EXPECTED_HGVSP` / `EXPECTED_HGVSC` INFO keys. Catches the wrong-variant / wrong-coordinate lookup class.
- **CONTRADICTORY_EVIDENCE** — PP3 and BP4 both triggered. Per ClinGen SVI these are mutually exclusive; they co-fire when SIFT and PolyPhen disagree.
- **MISSING_PROVENANCE** — a triggered criterion carries no evidence source.

These are deterministic and reproducible; they cannot hallucinate. The design intent is that every new failure class an adversarial review surfaces gets promoted to an invariant here.

## Behaviour
- `run_classification` runs the audit per variant and replaces the call with `Abstained (self-audit)` on any violation.
- `result.json` variants gain `abstained: bool` and `audit_violations: [{code, detail}]`.
- Variants with no asserted identity and coherent evidence are unaffected (demo output unchanged).

## Tests
Adds `test_self_audit.py` (11 tests) including the CTH wrong-variant regression (a coordinate resolving to p.Ser403Ile while the caller asserted p.Thr67Ile must abstain). Full clinical-variant-reporter suite: 49 passed.

---

> [!NOTE]
> **Medium Risk**
> Changes clinical output semantics for variants with asserted INFO identity or inconsistent in-silico/provenance evidence; abstentions may surprise callers who relied on always getting P/LP/VUS/LB/B.
> 
> **Overview**
> Introduces a **fail-closed self-audit** that runs after ACMG classification and before reports are emitted. Variants that fail deterministic checks are **hard-abstained** (`Abstained (self-audit)`) instead of receiving a confident five-tier call.
> 
> New module `self_audit.py` enforces three invariants: **IDENTITY_MISMATCH** (VCF `GENE` / `EXPECTED_HGVSP` / `EXPECTED_HGVSC` vs VEP-resolved identity), **CONTRADICTORY_EVIDENCE** (PP3 and BP4 both triggered), and **MISSING_PROVENANCE** (triggered criteria with empty or “no data” sources). `run_classification` wires `audit_classified` per variant; `result.json` gains `abstained` and `audit_violations`. `SKILL.md` documents the capability; `test_self_audit.py` covers the invariants including a CTH wrong-variant regression and an end-to-end demo identity-mismatch case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 311ce4af2c9616f7aaedbdea77b3086a4dfcb1bf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/skills/clinical-variant-reporter/SKILL.md
+++ b/skills/clinical-variant-reporter/SKILL.md
@@ -65,7 +65,12 @@ You are **Clinical Variant Reporter**, a specialised ClawBio agent for guideline
 4. **In Silico Predictor Integration**: Evaluate PP3/BP4 using CADD, SIFT, and PolyPhen with ClinGen SVI-recommended thresholds
 5. **Secondary Findings Screening**: Flag variants in ACMG SF v3.2 genes (81 genes; Miller et al., 2023) and classify them independently
 6. **Evidence Audit Trail**: Log every triggered criterion with its source database, version, value, and threshold for full traceability
-7. **Clinical Report Generation**: Structured Markdown report following ACMG laboratory reporting standards (Rehm et al., 2013) — methodology, classified variants, secondary findings, limitations, and disclaimer
+7. **Fail-Closed Self-Audit**: Before emitting any call, run deterministic invariants and **hard-abstain** any variant that violates one, rather than report a confident, possibly-wrong classification (safe uncertainty over confident hallucination):
+   - `IDENTITY_MISMATCH`: the variant resolved at the coordinate is not the one asserted (gene / HGVS via the ID column or `GENE` / `EXPECTED_HGVSP` / `EXPECTED_HGVSC` INFO keys) — catches wrong-variant / wrong-coordinate lookups
+   - `CONTRADICTORY_EVIDENCE`: mutually exclusive computational criteria (PP3 and BP4) both fired (ClinGen SVI: exclusive)
+   - `MISSING_PROVENANCE`: a triggered criterion carries no evidence source
+   Abstained variants are labelled `Abstained (self-audit)` in `result.json` with `abstained: true` and machine-readable `audit_violations`.
+8. **Clinical Report Generation**: Structured Markdown report following ACMG laboratory reporting standards (Rehm et al., 2013) — methodology, classified variants, secondary findings, limitations, and disclaimer
 
 ## Input Formats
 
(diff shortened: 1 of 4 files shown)


manuelcorpas (commented):
Audited: SAFE (approve; posted as comment since self-authored). Fail-closed self-audit with hard-abstain — encodes safe-uncertainty-over-confident-hallucination. Deep domain read verified the two points that would silently defeat the guarantee: (1) IDENTITY_MISMATCH join is exact because VariantEvidence copies chrom/pos/ref/alt verbatim from the source VcfRecord, not from VEP's allele string; (2) MISSING_PROVENANCE won't over-abstain because every acmg_engine criterion populates a non-empty source, and the only 'no data' strings are exactly the pathology it targets. Bench: cvr_identity 6/6; cvr_correctness 9/13 and clinical_variant_reporter 4/5 are IDENTICAL to the pre-PR baseline (zero new regressions). Two low/advisory notes: _norm_protein does not convert 1-letter vs 3-letter amino-acid codes, so a mismatched-convention EXPECTED_HGVSP would spuriously abstain a correct variant (fails safe = over-abstain; suggest normalising AA codes or documenting the 3-letter requirement); and the skill's SKILL.md is missing 5 template sections (pre-existing, not this PR).

PR #324, opened 2026-07-21
Add just-prs MCP skill for VCF and WGS scoring

antonkulaga:
## Summary

* Adds a `just-prs` skill backed by the pinned `just-prs-mcp` **0.3.1** local stdio server. It gives ClawBio users a direct route from a local VCF or WGS callset to more than 5,000 polygenic scores in the PGS Catalog.
* Supports trait search, exact PGS IDs, curated trait panels, ancestry-specific percentiles (including `AUTO`), model quality and coverage, model comparison, and absolute-risk estimates when the upstream evidence is reliable enough.
* Keeps the existing `gwas-prs` path for 23andMe and AncestryDNA files. The orchestrator sends VCF and WGS requests to `just-prs` without changing the current DTC workflow.

## Review follow-ups (manuelcorpas)

* **Ancestry disclosure (blocking):** omitting `--superpopulation` still defaults to EUR, but now emits a visible stderr warning and a report.md WARNING callout. Requested ancestry and reference-panel ancestry are shown in the report header, summary, and score table.
* Softened the oversold routing keyword from `absolute genetic risk from VCF` to `evidence-aware PRS from VCF`.
* Documented that absolute-risk prevalence baselines may be ancestry-inconsistent with ancestry-stratified percentiles.
* Allowed `--superpopulation AUTO` (upstream 0.3.1) so callers can opt into inferred ancestry instead of EUR.
* Non-blocking advisory on uvx hash pinning left as version-pin only (uvx `@0.3.1`); noted as an upstream packaging trade-off.

## Why this helps ClawBio

ClawBio currently has a useful PRS workflow for consumer genotype files, but VCF and WGS users need a broader scoring path. `just-prs` can search and compute more than 5,000 published PGS Catalog models. A user can ask for a trait or provide a PGS ID instead of finding scoring files and assembling the workflow by hand. Scoring files are cached locally, so repeat runs avoid downloading the same data again.

Personal genomic data stays local. ClawBio passes a resolved file path to a local stdio process and does not serialize VCF contents into MCP messages or result files.

## Skill affected

New skill: `just-prs-mcp`, CLI alias `just-prs`.

## Checklist

## Test output

- `uv run --extra just-prs --with pytest python -m pytest skills/just-prs-mcp/tests/test_just_prs_mcp.py skills/bio-orchestrator/tests/test_orchestrator.py -q`: **81 passed**
- `uv run --extra just-prs agentskills validate skills/just-prs-mcp`: valid skill
- Offline demo writes ancestry fields in `report.md` / `result.json`

Live integration tests still available behind `CLAWBIO_RUN_LIVE_JUST_PRS=1` and now expect `just-prs-mcp==0.3.1`.

Made with [Cursor](https://cursor.com)


manuelcorpas (changes requested):
Audited: NEEDS REVIEW — strong, privacy-conscious work; one equity fix requested before merge.

What is clean (verified, not assumed):
- Data egress: no genotype leaves the machine. Local stdio child (uvx just-prs-mcp@0.2.0, version-pinned) via FastMCP StdioTransport; the VCF is passed as a local path string only, downstream MCP tools receive only derived scalars. Tests enforce this (no vcf_content in serialised calls; no #CHROM/GT in live output). Satisfies Safety Rule #1.
- Risk math: primary output is percentile/z-score/spread, not naive PRS-to-risk. Absolute risk is calibrated upstream (population prevalence, or_per_sd) and gated on percentile reliability, recorded 'unavailable' with a reason otherwise. Not invented in the bridge.
- No shell/injection (static pinned command list, user input flows only as JSON MCP args); tested env allowlist (_CHILD_ENV_ALLOWLIST keeps unrelated keys out of the child); full per-score provenance (PGS ID/URL/method/effect size/AUROC + input SHA-256); disclaimer emitted via the shared report footer (test-asserted); honest offline demo + env-gated live test; edge cases handled conservatively.
- Bench: bio-orchestrator harness 53/54 at both baseline and this PR — no regression. Static scan SAFE. SKILL.md conformance strong (all required sections, <500 lines, demo file, 7 gotchas).

Requested change (blocking) — ancestry disclosure:
- just_prs_mcp_bridge.py:779-781: --superpopulation silently defaults to EUR when omitted. And the human-readable report.md (header ~459-466 and score table ~491) never states which reference-panel ancestry produced the percentile / absolute-risk figures. A non-EUR user who forgets the flag therefore gets EUR-referenced numbers presented without the assumption visible in the narrative (the structured result.json/scores.csv do carry requested_superpopulation + reference_panel_ancestry, which is why this is a warning rather than a hard block). This is the ancestry-over-confidence / Eurocentric-default class we care about.
- Fix: (1) emit a visible warning when EUR is defaulted rather than explicitly chosen, and (2) surface reference_panel_ancestry in the report header and as a score-table column.

Non-blocking advisories:
- The routing keyword 'absolute genetic risk from VCF' oversells; the report body is honest/gated — please soften the keyword to match.
- Absolute-risk prevalence baseline may be single-population upstream, so the % can be ancestry-inconsistent with an ancestry-stratified percentile; compounds with the ancestry point.
- uvx pins version but not hash; introduces a runtime dependency on dna-seq/just-prs-mcp.

Very close to mergeable — everything structural is right; the ancestry disclosure is the one thing to land first. Happy to re-review promptly.

antonkulaga:
Rebased onto latest `main` and addressed the review:

1. **Ancestry disclosure** — omitting `--superpopulation` now warns on stderr and in `report.md`; requested + reference-panel ancestry appear in the header, summary, and score table.
2. Softened keyword `absolute genetic risk from VCF` → `evidence-aware PRS from VCF`.
3. Noted prevalence/ancestry inconsistency in interpretation boundaries.
4. Bumped pinned engine `just-prs-mcp` **0.2.0 → 0.3.1**, with `--superpopulation AUTO` support.

Unit/orchestrator suite: 81 passed. Ready for re-review.

manuelcorpas (changes requested):
## Re-review after `f62e266` ("disclose EUR ancestry default")

Thanks for the ancestry-disclosure commit — the **omitted-flag** path is now handled well: `normalize_superpopulation` returns the disclosure flag, the report carries a banner, and there is a stderr warning. That is exactly right.

One equity-critical path was missed, and it is the one most likely to bite non-European users.

### CRITICAL — `--superpopulation AUTO` still falls back to EUR in silence
`just_prs_mcp_bridge.py`:
```python
panel = superpopulation
if panel == "AUTO":
    panel = str(row.get("reference_panel_ancestry") or _DEFAULT_SUPERPOPULATION)
```
When the engine does not report a `reference_panel_ancestry`, EUR is substituted and **`superpopulation_defaulted` stays `False`**. The report then prints "Requested ancestry: AUTO" with no indication that EUR was applied — no banner, no stderr warning, nothing in `result.json`. This is the undisclosed-EUR failure your own Gotchas section forbids, and no test covers it: `test_ten_run_mapping_stress` parametrises only AFR/AMR/EAS/EUR/SAS, never AUTO.

Suggested shape: record `superpopulation_applied` and `superpopulation_auto_defaulted` per model, raise the existing `superpopulation_defaulted` flag when any model fell back, and name the affected models in the banner.

### CRITICAL — the docs contradict each other on what AUTO does
- `clawbio/cli.py`: `"(AUTO infers ancestry; omit to default EUR with a warning)"`
- `SKILL.md`: *"It does not call variants, **infer ancestry**, diagnose disease..."*

The code does the latter — it reads the model's reference panel and never inspects the sample. The CLI help should say so; as written it advertises a capability that would be a serious equity hazard if real.

### WARNING — `build_mismatch` is reported but not acted on
SKILL.md says *"Do not interpret a build-mismatched score until the coordinate build is resolved."* The code prints `- Build mismatch: **True**` and then renders percentiles, absolute risk and a model-agreement verdict regardless. On mismatch, variants were matched across assemblies, so absolute risk should be withheld and the report should open with a warning. There is also no `--build` flag, while `PRS_MCP_DEFAULT_GENOME_BUILD` is env-allowlisted — so build can be set invisibly by environment with no user-facing control.

### WARNING — the local-first claim names the wrong trust boundary
SKILL.md: *"The VCF remains on the machine. The bridge passes only its resolved path to a local stdio child process and has no upload path."*

The bridge itself is genuinely clean — I checked: stdio only, path-only requests, no `http(s)`/`s3`/`gs`/`ftp` anywhere, and `test_server_launch_is_pinned_and_local` asserts no URL in the command. But `uvx just-prs-mcp@0.3.1` fetches and runs a third-party package at runtime (version-pinned, not hash-pinned), and *that* process is what actually reads the genome, with `HTTP_PROXY`/`HTTPS_PROXY` deliberately forwarded. Suggest wording like "ClawBio performs no upload; egress is delegated to a version-pinned dependency" — accurate, and still a good story.

### Smaller points
- **`catalog.json` drops the `clinical-variant-prioritizer` entry.** That skill directory does not exist on `main` either, so this reads as regeneration clearing a stale orphan rather than a real deletion — but please confirm, since catalog edits change what the orchestrator can route to. A `lit-synthesizer` stub assertion is also removed from `test_orchestrator.py`.
- **Orchestrator routing is greedy.** The PRS block sits above the scRNA/Illumina/sarek branches, so "annotate my VCF and give me a clinical risk score" hijacks `variant-annotation` and `clinical-variant-reporter`. Bare substring `"prs"` will also match inside unrelated tokens.
- **`percentile_mean` across LD-correlated models** is persisted to `result.json`. The report wisely shows only min/max/spread, but `profile-report` is a declared chaining partner and will read that field as a consensus percentile. It is not one — consider dropping or renaming it.
- **Demo fixture numbers cannot come from the demo VCF.** `demo_patient.vcf` holds three telomeric chr1 variants with zero weight-mass coverage (your own live test asserts this), yet `demo_trait_report.json` claims match_rate 0.92, percentile 78 and `OR=1.55` against real IDs PGS000014/PGS000013 — and `run_demo` binds that to a real SHA-256 of the VCF with live PGS Catalog links. It *is* labelled synthetic, so this is not covert, but it is one screenshot from being read as a computed result. Suggest moving the label into the report H1 and using reserved PGS IDs throughout.
- **No strand/ambiguous-SNP or missing-genotype policy is surfaced.** Nothing addresses effect-allele vs REF/ALT flips or A/T and G/C ambiguity, the classic silent sign-flip. Unmatched variants deflate the raw score toward zero, and that deflated score is placed against a full-coverage reference distribution to produce a percentile. Delegating the maths upstream is fine; the convention should be stated in the report.
- **`accepts_genotypes: False`** — I flagged this initially and then withdrew it. Per the comment at `cli.py:321` the flag means "accepts a genotype dict rather than a file", and nothing reads it as a privacy gate, so `False` is correct here. No change needed.

### Credit where due
The engineering discipline is above the hackathon average: pinned server, env allowlist, path-only requests, reliability-gated absolute risk, honest filter provenance, and `test_unreliable_percentile_blocks_absolute_risk` asserting the exact call sequence is the right shape of test. `assess_quality` gating and the `weight_mass_coverage` honesty signal are good calls.

### Offer
I have the AUTO disclosure, the build-mismatch gate and the two doc corrections implemented and tested locally (skill suite 35 -> 37, 113 passing with bio-orchestrator). Maintainer edits are enabled on this PR — say the word and I will push them, or you may prefer to take the descriptions above and implement them yourself.

manuelcorpas:
@antonkulaga I have pushed two commits to this branch (maintainer edits are enabled) rather than leaving you a list to work through. Please review them — revert or amend freely if you disagree with any of it.

**`b356f3e` — Disclose the AUTO ancestry fallback and enforce the build-mismatch stop**
- `--superpopulation AUTO` now records `superpopulation_applied` and `superpopulation_auto_defaulted` per model. When the engine reports no reference panel, the EUR fallback raises the existing `superpopulation_defaulted` flag and names the affected models in a report banner, instead of printing "Requested ancestry: AUTO" with EUR silently applied.
- `build_mismatch` now withholds absolute risk with a stated reason and opens the report with a warning, matching what SKILL.md already promised.
- CLI help in `clawbio/cli.py` and the bridge no longer says "AUTO infers ancestry" — it reads the model's reference panel and never inspects the sample, and now says so.
- SKILL.md local-first wording now names the real trust boundary: ClawBio performs no upload, but `uvx` runs the pinned third-party package that actually reads the genome.

**`bc14583` — Stop PRS routing pre-empting annotation and pipeline skills**

Measured on your branch before the change:

```
annotate my VCF and give me a clinical risk score  -> just-prs-mcp
run sarek on my WGS then give me a risk score      -> just-prs-mcp
```

The PRS detector sat above the scRNA/Illumina/sarek branches, so a risk score mentioned alongside a stronger primary intent claimed the query. PRS routing now yields when a competing intent is present. Routing for those queries is identical to `main` again; the only behavioural difference from `main` is that a pure PRS-from-VCF query reaches `just-prs-mcp`, which is the point of this PR. `"compute a polygenic risk score from my VCF"`, the DTC path to `gwas-prs`, and the explicit `just-prs` escape hatch are unchanged.

Tests: 37 in the skill suite, 55 in bio-orchestrator, 118 together. Five routing cases added, three of which failed beforehand.

**Left for you, deliberately** — these are your design calls, not defects I should overwrite:
- `percentile_mean` across LD-correlated models is still persisted to `result.json`; `profile-report` is a declared chaining partner and may read it as a consensus percentile.
- The demo fixture still carries real PGS IDs and a real input SHA-256 with live catalog links. Labelled synthetic, so not covert, but worth moving the label into the report H1 and using reserved IDs.
- No strand/ambiguous-SNP or missing-genotype policy is surfaced in the report.
- `reproducibility/commands.sh` embeds an absolute local path, leaking the username.
- Please confirm the `clinical-variant-prioritizer` removal from `catalog.json` was just regeneration clearing a stale orphan — that directory does not exist on `main` either, so I believe it is benign.

Thanks for the pinned server, env allowlist, path-only requests and reliability-gated absolute risk — that groundwork is why the remaining fixes were small.

PR #325, opened 2026-07-22
Fix pharmgx GRCh38 reference positions (3 were GRCh37), unblocking GRCh38 input

manuelcorpas:
## The bug

`pharmgx_reporter._GRCH38_POSITIONS` is used by `detect_reference_genome` to check the input's reference build. If enough positions mismatch, it returns `GRCh37_mismatch` and every diplotype is forced to `Indeterminate (reference genome mismatch)`.

Three of the five entries held **GRCh37** coordinates, mislabelled as GRCh38:
- `rs4244285` (CYP2C19\*2): 96541616 → should be **94781859**
- `rs1799853` (CYP2C9\*2): 96702047 → should be **94942290**
- `rs9923231` (VKORC1): 31107689 → should be **31096368**

So genuinely GRCh38 input (e.g. genotypes extracted from a GRCh38 whole-genome VCF) matched only 2 of 5 reference positions and was wrongly classified as GRCh37, blocking all star-allele calling. This silently disables the skill on modern WGS data.

## Fix

Correct the three positions to their true GRCh38 coordinates. After the fix, GRCh38 input is detected as GRCh38 and diplotypes are called normally (verified on a real GRCh38 family genome: CYP2C9 \*1/\*2 → Intermediate, CYP2D6 \*1/\*2 → Normal, CYP2C9 \*2/\*2 called correctly, CYP2C19 correctly abstains on phase ambiguity).

## Tests

Adds a regression test asserting all five reference positions are GRCh38 and that GRCh38 input is detected as GRCh38. Full pharmgx suite: 61 passed.

---

> [!NOTE]
> **High Risk**
> Changes pharmacogenomic reporting logic: wrong build detection previously blocked all star-allele calls on modern GRCh38 data and could mis-handle GRCh37; fixes are critical-path for drug guidance accuracy.
> 
> **Overview**
> Fixes **reference genome handling** so GRCh38 WGS/DTC inputs are not misclassified and GRCh37 samples still get full diplotype calls.
> 
> **Detection:** Replaces the GRCh38-only position table (three entries were wrong GRCh37 coords) with `_PGX_BUILD_POSITIONS` holding **GRCh37 and GRCh38** per panel SNP plus chromosome. `detect_reference_genome` now votes between builds (with chr normalization), returns `GRCh37`, `GRCh38`, `unknown_build`, or `None`, and drops `GRCh37_mismatch`.
> 
> **Calling policy:** Diplotypes stay **rsID-based** on GRCh37 and GRCh38. The old blanket “all genes Indeterminate” on GRCh37 mismatch is removed; only **`unknown_build`** forces every gene to Indeterminate. CLI messages and **SKILL.md** document the four outcomes.
> 
> **Tests:** New regression suite for coordinates (Ensembl truth), single-SNP detection, wrong-chrom rejection, demo GRCh37 end-to-end, lifted GRCh38 parity, and corrupt-file withholding.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5f294195e8887152f3acb4e7c76c21403fcda4d9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/skills/pharmgx-reporter/SKILL.md
+++ b/skills/pharmgx-reporter/SKILL.md
@@ -54,6 +54,23 @@ You are **PharmGx Reporter**, a specialised ClawBio agent for pharmacogenomic an
 | 23andMe raw data | `.txt`, `.txt.gz` | rsid, chromosome, position, genotype | `demo_patient.txt` |
 | AncestryDNA raw data | `.txt` | rsid, chromosome, position, allele1, allele2 | — |
 
+### Reference genome build
+
+Both **GRCh37 and GRCh38 inputs are supported**. Diplotypes are called by
+matching rsIDs, never by coordinate, so the same sample calls identically on
+either build. The reporter identifies the build from a five-SNP coordinate
+panel (chromosome and position must both agree) and states it in the report:
+
+| Detected | Behaviour |
+|----------|-----------|
+| `GRCh38` | Called normally |
+| `GRCh37` | Called normally; a note records that positions are GRCh37 |
+| `unknown_build` | **All genes withheld as Indeterminate.** Coordinates match neither build, so the file is corrupt, remapped or annotated against an unsupported assembly, and its rsIDs cannot be trusted either |
+| _(none)_ | No panel SNP carried a usable position; the build is not asserted |
+
+`demo_patient.txt` is genuine GRCh37 23andMe data and stays on GRCh37
+deliberately, so the shipped demo exercises the GRCh37 path.
+
 ## Workflow
 
 1. **Parse**: Read raw genetic data, auto-detect format (23andMe vs AncestryDNA)
--- a/skills/pharmgx-reporter/pharmgx_reporter.py
+++ b/skills/pharmgx-reporter/pharmgx_reporter.py
@@ -957,47 +957,91 @@ def detect_format(lines: list[str]) -> str:
         os.unlink(tmp)
 
 
-# Known GRCh38 positions for key PGx SNPs (used for reference genome mismatch detection)
-_GRCH38_POSITIONS = {
-    "rs4244285": 94781859,   # CYP2C19*2 (was 96541616, the GRCh37 position)
-    "rs3892097": 42128945,   # CYP2D6*4
-    "rs1799853": 94942290,   # CYP2C9*2 (was 96702047, the GRCh37 position)
-    "rs9923231": 31096368,   # VKORC1 (was 31107689, the GRCh37 position)
-    "rs1801133": 11796321,   # MTHFR C677T
+# Reference coordinates for key PGx SNPs, used only to identify which genome
+# build the input file is annotated against. Allele calling itself is rsID-based
+# and does not use these positions.
+#
+# Source: Ensembl release 112 / dbSNP b156.
+#   GRCh38: https://rest.ensembl.org/variation/human/<rsid>
+#   GRCh37: https://grch37.rest.ensembl.org/variation/human/<rsid>
+# Verified 2026-07-30. Both builds are stored so that a GRCh37 file is
+# positively identified rather than inferred from "did not match GRCh38",
+# which cannot distinguish an older build from a corrupt one.
+_PGX_BUILD_POSITIONS = {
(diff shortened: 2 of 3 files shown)


manuelcorpas:
## Audit: pharmgx GRCh38 coordinate fix

**Verdict: approve, with an additional commit pushed to this branch.**

### The coordinates are right
Verified independently against Ensembl REST (GRCh38 via `rest.ensembl.org`, GRCh37 via `grch37.rest.ensembl.org`), not against the PR's own table:

| rsID | Gene | This PR | True GRCh38 | True GRCh37 |
|---|---|---|---|---|
| rs4244285 | CYP2C19\*2 | 94781859 | 94781859 | 96541616 |
| rs1799853 | CYP2C9\*2 | 94942290 | 94942290 | 96702047 |
| rs9923231 | VKORC1 | 31096368 | 31096368 | 31107689 |
| rs3892097 | CYP2D6\*4 | 42128945 (unchanged) | 42128945 | 42524947 |
| rs1801133 | MTHFR | 11796321 (unchanged) | 11796321 | 11856378 |

All three changed values were verbatim GRCh37 positions. The premise is sound.

### But the fix was incomplete, and silently disabled the reporter
`detect_reference_genome()` inferred GRCh37 as "not GRCh38". Before this PR the contaminated table produced 3 spurious matches vs 2 mismatches, so the `mismatches >= matches_38` guard returned `"GRCh38"` and GRCh37 files were called normally. Correcting the constants flipped every real GRCh37 file to `GRCh37_mismatch`, and that path overwrites **every gene** with Indeterminate.

`demo_patient.txt` is genuine GRCh37 23andMe data, so the shipped demo regressed to a wall of Indeterminate. I confirmed by executing both variants.

`clawbio-bench --smoke --harness pharmgx`: **1/44 (2.3%)** on this branch as originally pushed, identical to `main`. The entire harness was floored.

### The added commit
Diplotypes are matched by rsID, never by coordinate, so a GRCh37 sample calls identically to the same sample on GRCh38. Withholding those calls was never scientifically necessary.

- `_PGX_BUILD_POSITIONS` stores **both** builds per SNP with chromosome and a dbSNP/Ensembl citation, so GRCh37 is identified positively rather than by elimination. An unrecognised assembly is no longer indistinguishable from an older one.
- `detect_reference_genome()` returns `GRCh38` / `GRCh37` / `unknown_build` / `None`. One unambiguous SNP now settles the build; the old rule needed two mismatches, so a sparse targeted GRCh37 panel was reported as GRCh38 and called anyway. Chromosome is checked where present.
- Only `unknown_build` withholds calls, which is the case that warrants it.
- `demo_patient.txt` stays GRCh37. It is real public data; lifting its coordinates to make the code pass would have falsified its provenance. This also means the GRCh37 demo files in `gwas-prs`, `profile-report` and the docs are supported input rather than a liability.

**Result: 1/44 (2.3%) -> 42/44 (95.5%).**

### On the test
The original regression test asserted the module's constants against a copy of themselves and would have passed both before and after a 1.7 Mb coordinate error — confirmed: 60 tests passed either way. Replaced with coordinates sourced from Ensembl, a GRCh37 -> `GRCh37` negative case (the assertion that would actually have failed pre-fix), chromosome and sparse-panel cases, and end-to-end assertions. Mutation-tested: re-contaminating a GRCh38 constant and disabling the chromosome check each fail a test. 60 -> 71 tests.

### Two remaining bench failures, neither caused by this PR
- `grch37_reference_mismatch` — a **bug in the bench scorer**. The fixture's own text reads *"rsids are stable across assemblies, so an rsid-based tool should produce the same calls regardless"* with `FINDING_CATEGORY: correct_determinate`, but its `GROUND_TRUTH_PHENOTYPE` string demands Indeterminate. The genotype is `rs3892097 TT` = `*4/*4` = Poor Metaboliser, which is what the skill now outputs and what the fixture's design section calls correct. Worth raising upstream at `biostochastics/clawbio_bench`; the skill has not been bent to match it.
- `tpmt_compound_het` — pre-existing. The skill abstains on `*3B`/`*3C` phase ambiguity; the bench expects Poor Metaboliser. Genuinely arguable, untouched here.

PR #327, opened 2026-07-30
fix(cvr): report live Ensembl release in Data Sources table

### LINKED ISSUE #303: fix(cvr): Data Sources table hardcodes stale ClinVar and gnomAD versions
## Summary

The clinical-variant-reporter Data Sources table hardcodes `ClinVar 2025-03-01 release` and `gnomAD v4.1` regardless of which Ensembl release is actually queried at runtime.

## Location

`skills/clinical-variant-reporter/clinical_variant_reporter.py`, lines 498-499.

## Suggested fix

In live mode, fetch the Ensembl release number from `https://rest.ensembl.org/info/software` and report ClinVar/gnomAD as bundled with that release. In demo mode, label as demo cache.

AmirF194:
## Root cause

`_write_markdown_report`'s Data Sources section hardcoded `ClinVar 2025-03-01 release`
and `gnomAD v4.1` unconditionally, in both demo and live mode, and the GRCh37 code path
queried the GRCh38 Ensembl host for both variant annotation and version lookup. A report
generated against a newer live release, or against GRCh37 data, still claimed the
2025-03-01 GRCh38 ClinVar snapshot.

## Fix

In live mode, fetch ClinVar/dbSNP/OMIM versions from Ensembl's
`GET /info/variation/homo_sapiens` (not `info/software`, which doesn't carry them) and
report them as returned for the run, using the correct host for the requested assembly
(`grch37.rest.ensembl.org` for GRCh37, matching the VEP annotation call itself). gnomAD
stays hardcoded and labeled as such, since Ensembl's variation endpoint doesn't list it.
`annotation_backend` in `database_versions.json` now names the assembly actually used
instead of hardcoding GRCh38. A failed lookup (network error, timeout, non-JSON body)
degrades to an explicit "unavailable" label distinct from "annotation succeeded but the
lookup failed" vs. "nothing was annotated," rather than silently falling back to a stale
hardcode. Demo mode says "demo cache" explicitly instead of implying a live query.
`result.json`/`database_versions.json` carry these as structured fields (string or null),
not just prose in `report.md`.

## Verification

- New/updated tests fail on `main` (hardcoded strings, wrong host for GRCh37, wrong
  backend label) and pass with this fix: GRCh37 routes to the GRCh37 host for both the
  VEP call and the version lookup, `annotation_backend` matches the assembly, and the
  three annotation outcomes (succeeded, ran but lookup failed, nothing annotated) each
  get a distinct label.
- Full `skills/clinical-variant-reporter/tests/` suite: 72 passed.
- Live-verified this session against the real Ensembl REST API (not just fixtures):
  `GET /info/variation/homo_sapiens` on the GRCh38 and GRCh37 hosts returns ClinVar
  `09/2025` and `06/2023` respectively, confirming both the endpoint shape and the
  GRCh37 host-routing fix are load-bearing, not just internally consistent.
- Not checked: gnomAD's hardcoded version against a live source, since Ensembl exposes
  none; no end-to-end run against a real patient VCF.

Fixes #303

### DIFF
--- a/skills/clinical-variant-reporter/SKILL.md
+++ b/skills/clinical-variant-reporter/SKILL.md
@@ -206,7 +206,7 @@ output_directory/
 
 ## Safety
 
-- **Local-first**: All classification logic runs locally. Only variant coordinates and alleles are sent to public Ensembl VEP REST — no patient identifiers or phenotype data ever leave the machine
+- **Local-first**: All classification logic runs locally. Only variant coordinates and alleles are sent to public Ensembl VEP REST. In live mode, the Data Sources report section also queries Ensembl's `info/variation/homo_sapiens` endpoint (no variant data in that request) to report the actual ClinVar/dbSNP/OMIM versions bundled with the release — no patient identifiers or phenotype data ever leave the machine
 - **Disclaimer**: Every report includes the ClawBio medical disclaimer
 - **No hallucinated science**: Every classification traces to specific evidence codes, database entries, and published thresholds
 - **Audit trail**: Full evidence provenance logged to `reproducibility/database_versions.json`
--- a/skills/clinical-variant-reporter/clinical_variant_reporter.py
+++ b/skills/clinical-variant-reporter/clinical_variant_reporter.py
@@ -42,7 +42,9 @@ DISCLAIMER = (
 VEP_REST_URL = "https://rest.ensembl.org/vep/homo_sapiens/region"
 VEP_BATCH_SIZE = 200
 VEP_RATE_LIMIT_SECONDS = 0.07  # ~15 requests/second
-ENSEMBL_INFO_SOFTWARE_URL = "https://rest.ensembl.org/info/software"
+ENSEMBL_INFO_VARIATION_URL = "https://rest.ensembl.org/info/variation/homo_sapiens"
+GNOMAD_VERSION_LABEL = "v4.1"  # info/variation does not list gnomAD; no live source exists
+DEMO_CLINVAR_LABEL = "2025-03-01 release"
 
 
 # ---------------------------------------------------------------------------
@@ -269,26 +271,43 @@ def _extract_evidence_from_vep(vep_result: dict, record: VcfRecord) -> VariantEv
     )
 
 
-def _fetch_ensembl_release() -> int | None:
-    """Query the Ensembl release actually queried, for the Data Sources report section."""
+def _fetch_ensembl_data_versions() -> dict[str, str] | None:
+    """Query the live source versions bundled with Ensembl's variation data
+    (ClinVar, dbSNP, OMIM), for the Data Sources report section. gnomAD is not
+    listed by this endpoint and stays a hardcoded constant (GNOMAD_VERSION_LABEL)."""
     try:
         import requests
         resp = requests.get(
-            ENSEMBL_INFO_SOFTWARE_URL,
+            ENSEMBL_INFO_VARIATION_URL,
             headers={"Content-Type": "application/json"},
             timeout=10,
         )
         resp.raise_for_status()
-        return resp.json().get("release")
-    except Exception:
+        sources = resp.json()
+    except Exception as exc:
+        print(f"WARNING: could not fetch Ensembl data source versions ({exc}).", file=sys.stderr)
         return None
+
+    versions: dict[str, str] = {}
+    name_to_key = {"ClinVar": "clinvar", "dbSNP": "dbsnp", "OMIM": "omim"}
+    for src in sources if isinstance(sources, list) else []:
+        key = name_to_key.get(src.get("name", ""))
+        if key and src.get("version"):
+            versions[key] = src["version"]
+    return versions or None
 
 
 def annotate_variants_vep(
     records: list[VcfRecord],
     assembly: str = "GRCh38",
-) -> list[VariantEvidence]:
-    """Annotate variants via Ensembl VEP REST API and return evidence objects."""
+) -> tuple[list[VariantEvidence], dict[str, str] | None]:
+    """Annotate variants via Ensembl VEP REST API.
+
+    Returns (evidence_list, source_versions). source_versions is fetched once,
+    only if at least one VEP batch succeeded, so a run where every batch fails
+    never stamps a report with data-source versions for annotation that never
+    happened.
+    """
     try:
         import requests
     except ImportError:
@@ -296,6 +315,7 @@ def annotate_variants_vep(
         sys.exit(1)
 
     evidence_list: list[VariantEvidence] = []
+    any_batch_succeeded = False
 
     for batch_start in range(0, len(records), VEP_BATCH_SIZE):
         batch = records[batch_start:batch_start + VEP_BATCH_SIZE]
@@ -320,6 +340,7 @@ def annotate_variants_vep(
                 ))
             continue
 
+        any_batch_succeeded = True
         result_map: dict[str, dict] = {}
         for vr in vep_results:
             loc = vr.get("input", "")
@@ -337,7 +358,8 @@ def annotate_variants_vep(
 
         time.sleep(VEP_RATE_LIMIT_SECONDS)
 
-    return evidence_list
+    source_versions = _fetch_ensembl_data_versions() if any_batch_succeeded else None
+    return evidence_list, source_versions
 
 
 # ---------------------------------------------------------------------------
@@ -348,13 +370,19 @@ def run_classification(
     demo: bool = False,
     gene_filter: set[str] | None = None,
     assembly: str = "GRCh38",
-) -> list[ClassifiedVariant]:
-    """Run the full ACMG classification pipeline on VCF records."""
+) -> tuple[list[ClassifiedVariant], dict[str, str] | None]:
+    """Run the full ACMG classification pipeline on VCF records.
+
+    Returns (classified, source_versions); source_versions is always None in
+    demo mode and is the Ensembl data-source versions captured at annotation
+    time in live mode (None if annotation never succeeded).
+    """
     if demo:
         cache = load_demo_evidence_cache()
         evidence_list = [build_evidence_from_cache(r, cache) for r in records]
+        source_versions = None
     else:
-        evidence_list = annotate_variants_vep(records, assembly=assembly)
+        evidence_list, source_versions = annotate_variants_vep(records, assembly=assembly)
 
     if gene_filter:
         evidence_list = [e for e in evidence_list if e.gene in gene_filter]
@@ -372,7 +400,7 @@ def run_classification(
         if not audit.passed:
             cv.classification = ABSTAIN_LABEL
         classified.append(cv)
-    return classified
+    return classified, source_versions
 
 
 # ---------------------------------------------------------------------------
@@ -382,14 +410,52 @@ CLASS_ORDER = ["Pathogenic", "Likely Pathogenic", "Uncertain Significance", "Lik
 CLASS_SHORT = {"Pathogenic": "P", "Likely Pathogenic": "LP", "Uncertain Significance": "VUS", "Likely Benign": "LB", "Benign": "B"}
 
 
+def _data_source_versions_for_report(
+    demo: bool, source_versions: dict[str, str] | None,
+) -> dict[str, str]:
+    """Single source of truth for the Data Sources table, result.json and
+    database_versions.json, so the three can't drift independently.
+
+    Never performs a network call: source_versions must already be captured
+    (annotate_variants_vep, at annotation time) or None.
+    """
+    if demo:
+        return {
+            "ClinVar": f"{DEMO_CLINVAR_LABEL} (demo cache)",
+            "gnomAD": f"{GNOMAD_VERSION_LABEL} (demo cache)",
+        }
+
+    if not source_versions:
+        note = "unavailable (no variant was successfully annotated this run)"
+        return {"ClinVar": note, "gnomAD": note}
+
+    clinvar_v = source_versions.get("clinvar")
+    clinvar_label = (
+        f"{clinvar_v} (via Ensembl VEP colocated variants)" if clinvar_v
+        else "unavailable (Ensembl did not report a ClinVar version this run)"
+    )
+    return {
+        "ClinVar": clinvar_label,
+        "gnomAD": f"{GNOMAD_VERSION_LABEL} (via Ensembl VEP colocated variants)",
+    }
+
+
 def generate_report(
     classified: list[ClassifiedVariant],
     output_dir: Path,
     demo: bool = False,
     assembly: str = "GRCh38",
     input_path: str = "demo",
+    source_versions: dict[str, str] | None = None,
 ) -> None:
-    """Generate all output files: report.md, result.json, tables/, figures/, reproducibility/."""
+    """Generate all output files: report.md, result.json, tables/, figures/, reproducibility/.
+
+    source_versions is the Ensembl data-source versions captured at annotation
+    time (see annotate_variants_vep / run_classification); this function makes
+    no network call of its own, so regenerating a report from stored
+    ClassifiedVariant objects without passing source_versions again correctly
+    shows "unavailable" rather than re-fetching today's release for an old run.
+    """
     output_dir.mkdir(parents=True, exist_ok=True)
     (output_dir / "tables").mkdir(exist_ok=True)
     (output_dir / "figures").mkdir(exist_ok=True)
@@ -402,13 +468,14 @@ def generate_report(
         counts[cv.classification] = counts.get(cv.classification, 0) + 1
 
     sf_variants = [cv for cv in classified if cv.is_secondary_finding]
+    data_sources = _data_source_versions_for_report(demo, source_versions)
 
-    _write_markdown_report(classified, counts, sf_variants, output_dir, timestamp, demo, assembly, input_path)
+    _write_markdown_report(classified, counts, sf_variants, output_dir, timestamp, demo, assembly, input_path, data_sources)
     _write_classification_table(classified, output_dir)
     _write_secondary_findings_table(sf_variants, output_dir)
-    _write_result_json(classified, counts, sf_variants, output_dir, timestamp, demo, assembly)
+    _write_result_json(classified, counts, sf_variants, output_dir, timestamp, demo, assembly, data_sources)
     _write_classification_figure(counts, output_dir)
-    _write_reproducibility(output_dir, demo, assembly, input_path, timestamp)
+    _write_reproducibility(output_dir, demo, assembly, input_path, timestamp, data_sources)
 
 
 def _write_markdown_report(
@@ -420,6 +487,7 @@ def _write_markdown_report(
     demo: bool,
     assembly: str,
     input_path: str,
+    data_sources: dict[str, str],
 ) -> None:
     lines: list[str] = []
     lines.append("# Clinical Variant Report — ACMG/AMP Classification")
@@ -529,14 +597,8 @@ def _write_markdown_report(
     lines.append("")
     lines.append("| Source | Version / Release |")
     lines.append("|--------|-------------------|")
-    if demo:
-        lines.append("| ClinVar | 2025-03-01 release (demo cache) |")
-        lines.append("| gnomAD | v4.1 (demo cache) |")
-    else:
-        release = _fetch_ensembl_release()
-        release_label = f"Ensembl release {release}" if release is not None else "Ensembl release unavailable"
-        lines.append(f"| ClinVar | bundled with {release_label} (via Ensembl VEP REST) |")
-        lines.append(f"| gnomAD | bundled with {release_label} (via Ensembl VEP colocated variants) |")
+    lines.append(f"| ClinVar | {data_sources['ClinVar']} |")
+    lines.append(f"| gnomAD | {data_sources['gnomAD']} |")
     lines.append("| Ensembl VEP | REST API, assembly %s |" % assembly)
     lines.append("| ACMG SF list | v3.2 (Miller et al., 2023; 81 genes) |")
     lines.append("")
@@ -603,6 +665,7 @@ def _write_result_json(
     timestamp: str,
     demo: bool,
     assembly: str,
+    data_sources: dict[str, str],
 ) -> None:
     result = {
         "tool": "ClawBio Clinical Variant Reporter",
@@ -612,6 +675,7 @@ def _write_result_json(
         "assembly": assembly,
         "timestamp": timestamp,
         "mode": "demo" if demo else "live",
+        "data_source_versions": data_sources,
         "total_variants": len(classified),
         "classification_counts": counts,
         "secondary_findings_count": len(sf_variants),
@@ -677,6 +741,7 @@ def _write_reproducibility(
     assembly: str,
     input_path: str,
     timestamp: str,
+    data_sources: dict[str, str],
 ) -> None:
     cmd = f"python {Path(__file__).name}"
     if demo:
@@ -694,6 +759,7 @@ def _write_reproducibility(
         "sf_list": "ACMG SF v3.2 (Miller et al. 2023)",
         "sf_gene_count": len(ACMG_SF_V32_GENES),
         "annotation_backend": "demo_cache" if demo else "Ensembl VEP REST (GRCh38)",
+        "data_source_versions": data_sources,
         "generated": timestamp,
     }
     (output_dir / "reproducibility" / "database_versions.json").write_text(
@@ -746,12 +812,15 @@ def main() -> None:
         print(f"[CVR] Filtering to genes: {', '.join(sorted(gene_filter))}")
 
     print(f"[CVR] Running ACMG classification ({'demo mode' if args.demo else 'live mode'})...")
-    classified = run_classification(
+    classified, source_versions = run_classification(
         records, demo=args.demo, gene_filter=gene_filter, assembly=args.assembly,
     )
 
     print(f"[CVR] Generating report in: {output_dir}")
-    generate_report(classified, output_dir, demo=args.demo, assembly=args.assembly, input_path=input_label)
+    generate_report(
+        classified, output_dir, demo=args.demo, assembly=args.assembly,
+        input_path=input_label, source_versions=source_versions,
+    )
 
     counts = {}
     for cv in classified:
--- a/skills/clinical-variant-reporter/clinical_variant_reporter.py
+++ b/skills/clinical-variant-reporter/clinical_variant_reporter.py
@@ -46,7 +46,6 @@ ENSEMBL_INFO_VARIATION_PATH = "/info/variation/homo_sapiens"
 VEP_REST_URL = VEP_REST_HOST + VEP_REST_PATH
 VEP_BATCH_SIZE = 200
 VEP_RATE_LIMIT_SECONDS = 0.07  # ~15 requests/second
-ENSEMBL_INFO_VARIATION_URL = VEP_REST_HOST + ENSEMBL_INFO_VARIATION_PATH
 GNOMAD_VERSION_LABEL = "v4.1"  # info/variation does not list gnomAD; no live source exists
 DEMO_CLINVAR_LABEL = "2025-03-01 release"
 
@@ -485,7 +484,7 @@ def _data_source_versions_for_report(
 
     clinvar_v = source_versions.get("clinvar")
     clinvar_label = (
-        f"{clinvar_v} (via Ensembl VEP colocated variants)" if clinvar_v
+        f"{clinvar_v} (via Ensembl /info/variation/homo_sapiens)" if clinvar_v
         else "unavailable (Ensembl did not report a ClinVar version this run)"
     )
     return {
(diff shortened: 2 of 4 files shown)


manuelcorpas (changes requested):
## Audit: live Ensembl release in Data Sources

Thanks for this — the bug in #303 is real. `ClinVar 2025-03-01` was stale by about six months, and the PR follows the approach the issue prescribed. The problem is that the prescription itself points at the wrong endpoint, so the fix trades one wrong field for two vague ones.

**Verdict: changes requested.**

### 1. The accurate ClinVar version is one call away, and this discards it
Queried live just now:

- `GET /info/software` -> `{"release": 116}`  (what this PR uses)
- `GET /info/variation/homo_sapiens` -> **`ClinVar | 09/2025`**, `dbSNP | 156`, `OMIM | 09/2025`

So the true answer is "ClinVar 09/2025", and Ensembl serves it directly. "ClinVar bundled with Ensembl release 116" is a pointer, not a provenance record: a clinician cannot resolve a release number to a ClinVar snapshot without consulting external release notes. This skill's SKILL.md sells traceability to "source database, version, and threshold", so this is a net loss of traceability.

`/info/software` is also the wrong field semantically — it reports the API/software version, not the data release (`/info/data`), though both read 116 today.

**gnomAD is a straight regression.** The removed `v4.1` appears correct and current, and `/info/variation/homo_sapiens` does not list gnomAD at all, so there is no live source for it. Better to leave the hardcoded `v4.1` alone than dissolve it into an Ensembl release number that says nothing about gnomAD.

### 2. The report can assert provenance for annotations that never happened
`_fetch_ensembl_release()` fires inside `_write_markdown_report`, i.e. at report-writing time. Nothing carries a release from the actual VEP responses — and I confirmed VEP REST replies expose no release header (only `x-ratelimit-*`, `x-runtime`), so a separate call is defensible. The placement is not:

- `generate_report` is public and takes stored `ClassifiedVariant` objects, so regenerating a report later stamps today's release onto old annotations.
- Worst case is silent: `annotate_variants_vep` swallows a failed batch and emits empty evidence with only a stderr warning. If **every** VEP batch fails (proxy, 429, outage) but the later `/info/software` GET succeeds, the report still prints "ClinVar bundled with Ensembl release 116 (via Ensembl VEP REST)" for a run in which VEP returned nothing. The old hardcoded rows at least made no claim about *this* run.

Capture the versions once at annotation time, thread them through, and suppress the claim when no batch annotated successfully.

### 3. One new test cements the regression
```python
assert "v4.1" not in text
```
This makes it a **test failure** to ever print the correct gnomAD version in a live report, blocking the correct fix.

### 4. The new tests exercise only the mock
All three call `generate_report([], ...)` with an **empty variant list**, so the annotation path never runs and no test could detect point 2. The live-mode test patches `requests.get` to a `MagicMock` with a no-op `raise_for_status`, so it asserts only that a hardcoded `999` survives string formatting. Untested: 4xx/5xx, a payload missing the `release` key (-> the "unavailable" branch), non-JSON bodies. And `test_demo_mode_labels_as_demo_cache` does not patch `requests.get`, so it never pins the "demo mode stays offline" invariant that is this PR's own purpose — patch it to raise and assert the demo report still renders.

Credit where due: no test makes a real network call, so CI stays hermetic.

### 5. Smaller points
- **New network egress, undocumented, no opt-out.** No patient data leaks (bare GET, no body), and SKILL.md already documents coordinate-level VEP egress, so the safety posture holds. But SKILL.md:209 now describes only one endpoint, there is no `--no-network` flag, and an offline run blocks for `timeout=10` before degrading. No memoisation, so batch-cohort loops pay that per report and risk 429s.
- **`except Exception: return None`** renders DNS failure, timeout, 500, **429**, TLS error and `ImportError` identically as "unavailable", with nothing on stderr — unlike the sibling VEP handler, which warns. A 429 matters, because it implies the VEP calls in the same run may also be degraded.
- **The fetched version never reaches the audit trail.** SKILL.md promises `database_versions.json` contains "ClinVar date, gnomAD version, VEP release, SF list version". It contains `acmg_framework`, `sf_list`, `sf_gene_count`, `annotation_backend`, `generated` — none of the first three, before or after this PR. Having fetched a release, it belongs there and in `result.json` far more than in a prose table.

### Suggested changes
1. Use `/info/variation/homo_sapiens` and print the real `ClinVar 09/2025`; keep `gnomAD v4.1`; use `/info/data` if a release number is still wanted.
2. Capture versions at annotation time and thread them to `generate_report`; suppress the claim when nothing annotated.
3. Write them into `database_versions.json` and `result.json`.
4. Add `--no-network`, memoise the fetch, update SKILL.md Safety to name the second endpoint.
5. `except Exception as exc:` plus a stderr warning, matching house style.
6. Drop `assert "v4.1" not in text`; add a demo-makes-zero-network-calls test and a missing-`release`-key test.

Bench: `clinical_variant_reporter` 4/5, `cvr_correctness` 9/13, `cvr_identity` 6/6 — identical to `main`, 0 case-level changes. Skill suite: 54 passed.

AmirF194:
Thanks, the review holds up on most points. I re-verified both endpoints live rather than trusting the numbers quoted here: `info/software` returns `{"release": 116}`, and `info/variation/homo_sapiens` returns a list that includes `{"name": "ClinVar", "version": "09/2025"}`, `{"name": "dbSNP", "version": "156"}`, `{"name": "OMIM", "version": "09/2025"}`, with no gnomAD entry anywhere in it. Matches what's posted above.

1. Switched to `info/variation/homo_sapiens` and now print the real ClinVar version (`09/2025` as of this run). Reverted gnomAD to its own hardcoded `v4.1`; there's no live endpoint for it, so tying it to an Ensembl release number was wrong, agreed.

2. Moved the fetch out of `_write_markdown_report` and into `annotate_variants_vep`, so it runs once per annotation pass and only when at least one VEP batch actually succeeded. `generate_report` now takes the captured `source_versions` and makes no network call of its own, so regenerating a report later from stored `ClassifiedVariant` objects no longer stamps a fresh release onto an old run; when nothing annotated, both Data Sources rows say so explicitly instead of printing an unearned claim.

3. Dropped `assert "v4.1" not in text`. The new tests assert that `v4.1` and the correct ClinVar version both appear in a live report, and that neither appears when nothing annotated.

4. Added tests for annotation success vs failure driving `source_versions` (including that the version fetch never even runs when annotation failed), a demo-mode zero-network-calls test, and HTTP error / non-JSON body / missing-ClinVar-entry responses on the fetch itself.

5. `except Exception as exc:` plus a stderr warning now, matching the VEP handler's style. Versions are written into `result.json` and `database_versions.json` under `data_source_versions`. SKILL.md's Safety section now names the second endpoint.

What I did not do: `--no-network` and an explicit memoisation cache. Moving the fetch to once per annotation pass already removes the repeated-refetch-on-every-regeneration problem, so most of what memoisation would buy is gone; a persistent cache only matters for a long-lived process, which this CLI isn't. `--no-network` would need to decide what live mode even means with VEP itself disabled, not just skip the version stamp, which is a bigger design question than fits in this PR, so I left it out rather than rush it.

Verified in a clean `python:3.12-slim` Docker container against this commit: full `skills/clinical-variant-reporter/tests/` suite, 61 passed, 0 failed. Also installed `clawbio-bench` from the commit pinned in `ci.yml` and ran the three harnesses named in this review against HEAD: `clinical_variant_reporter` 4/5, `cvr_correctness` 9/13, `cvr_identity` 6/6, 0 harness errors, identical to the numbers quoted above. Expected: this fix only touches Data Sources report metadata, not classification.

AmirF194:
Checking in a week after opening this, no rush. I addressed the review feedback above (verified both endpoint numbers live) and I'm happy to adjust further if anything still needs work.

manuelcorpas (changes requested):
## Audit summary: changes requested

**First, the red tick on this PR is our fault, not yours.** Your own suite passed in CI (`61 passed in 2.72s`). The job failed on `skills/just-prs-mcp`, which raised `ModuleNotFoundError: No module named 'typer'`, a skill you never touched. It got pulled in because our changed-skill step diffed against `pull_request.base.sha` and so attributed everything merged into main since 30 July to your PR. Both bugs are fixed in a separate CI PR. Please ignore that failure.

The core of this change is right, and we verified it against the live API rather than taking it on trust:

```
GET https://rest.ensembl.org/info/variation/homo_sapiens
-> {"name":"ClinVar","version":"09/2025"}, {"name":"dbSNP","version":"156"}, {"name":"OMIM","version":"09/2025"}
```

The parser at `:292` matches that payload exactly, and `ClinVar` appears exactly once among the sources, so there is no ambiguity. You also avoided the trap that `/info/rest` returns the REST API version (`15.12`) rather than the data release. Moving the fetch to annotation time at `:361` is the right call, since it stops a regenerated report stamping today's version onto an old run. Credit also for the named exception and stderr warning, and for removing the `assert "v4.1" not in text` regression.

### Blocking

**1. The report can state something that is factually untrue.** At `:429`, when annotation succeeds but only the version lookup fails, the report prints "unavailable (no variant was successfully annotated this run)". Everything was annotated. Three distinct states collapse into one wrong sentence, because `return versions or None` at `:297` makes "HTTP 200 with no recognised source names" indistinguishable from total failure. This is a clinical document, so please split the states and give each its own label:

- no variant was successfully annotated this run
- annotation succeeded; Ensembl version lookup failed
- Ensembl reported no version for this source

**2. `--assembly GRCh37` now certifies a version it never fetched.** This is a pre-existing bug that your change amplifies rather than one you introduced, but it should not ship in this form. The main host ignores the assembly parameter, and the two hosts disagree on precisely the field this PR reports:

```
rest.ensembl.org        -> ClinVar 09/2025, OMIM 09/2025
grch37.rest.ensembl.org -> ClinVar 06/2023, OMIM 06/2023
```

So a `--assembly GRCh37` run produces GRCh38 annotations under a GRCh37 header, now carrying a ClinVar version that is 28 months out. Either derive the host from `assembly` for both the VEP POST and the version GET and add a test that the host switches, or exit on `--assembly GRCh37` and open a follow-up issue. What it must not do is certify a version it did not obtain from the matching host.

**3. The new tests cannot catch the bug this PR fixes.** All 191 added lines patch `requests.get` with an argument-ignoring lambda, so no test asserts which URL is called. The suite passes identically if the code reverts to `/info/rest`, which is the exact defect being fixed here. Please capture `args`/`kwargs` and assert `ENSEMBL_INFO_VARIATION_URL` and `timeout=10`, mirroring your existing `test_transcript_version_param_sent`. Two other gaps worth closing: there is no test for "annotation succeeded, version fetch failed" (which is how item 1 shipped), and `raise_for_status` is never exercised as a real failure, since the mock raises from `get` itself and `MagicMock.raise_for_status()` is a silent no-op.

### Smaller items

- **`:439` labels a hardcoded constant as a live lookup.** `f"{GNOMAD_VERSION_LABEL} (via Ensembl VEP colocated variants)"` asserts a route that was not travelled, and will go stale silently. Suggest `(hardcoded; Ensembl serves no gnomAD version endpoint)`.
- **`VEP release` is still missing from `database_versions.json`.** `SKILL.md:193` promises ClinVar date, gnomAD version, VEP release and SF list version. The first two now arrive, which closes most of the earlier finding, but VEP release does not, and your first revision already had that number in hand via `/info/software`. `/info/data` returns `{"releases":[116]}` and is one call away. `dbsnp` and `omim` are captured then discarded too, despite `SKILL.md:209` advertising them.
- **Keep the audit JSON machine-readable.** `data_source_versions` currently receives English prose, including the "unavailable (...)" strings, while the structured `{"clinvar": "09/2025", "dbsnp": "156", "omim": "09/2025"}` is dropped. Emit the structured form there and keep prose for the Markdown table.
- **`:282` sends `Content-Type` on a GET**, which is a request-body header. It works today, but `Accept` is the correct content negotiation.
- **`annotation_backend` at `:761` hardcodes `(GRCh38)`** regardless of `--assembly` (pre-existing).

### Not your problem

The scanner reports `SKILL.md` missing `## Trigger`, `## Scope`, `## Example Output`, `## Gotchas` and `## Agent Boundary`. We checked against main: all five were already absent, and your PR removes nothing. That is our pre-existing debt and is not being asked of you here.

Also confirmed clean during review: no ACMG logic is touched (`acmg_engine.py` and `self_audit.py` are byte-identical to the merge base), the new call sends no variant data or identifiers (bare GET, fixed URL, no query string), a timeout is set, and demo mode remains provably offline.

AmirF194:
Addressed all three blocking items and most of the smaller ones. Commit: 2e0f184.

**1. False claim on partial failure (blocking).** Fixed. `annotate_variants_vep` (`:389-395`) now returns `{}` (not `None`) when at least one VEP batch succeeded but `_fetch_ensembl_data_versions` failed, keeping `None` reserved for "nothing was annotated". `_data_source_versions_for_report` (`:448`) now branches on three states instead of one falsy check: `None` -> "unavailable (no variant was successfully annotated this run)", `{}` -> "unavailable (annotation succeeded, but the Ensembl version lookup failed)", non-empty dict missing `clinvar` -> the existing "Ensembl did not report a ClinVar version this run" (unchanged wording, already correct). New test: `test_source_versions_empty_dict_when_annotation_succeeds_but_lookup_fails` and `test_live_mode_flags_version_lookup_failure_distinctly_from_no_annotation`.

**2. GRCh37 host mismatch (blocking).** Fixed by deriving the host, not by exiting. Added `_ensembl_rest_host(assembly)`, used by both the VEP POST (`annotate_variants_vep:346`) and the version GET (`_fetch_ensembl_data_versions:295`), routing GRCh37 to `grch37.rest.ensembl.org` and everything else to the main host. I ran this live against both hosts this session rather than trusting the numbers in the review:

```
GRCh38 host: {'dbsnp': '156', 'omim': '09/2025', 'clinvar': '09/2025'}
GRCh37 host: {'dbsnp': '156', 'omim': '06/2023', 'clinvar': '06/2023'}
```

Confirms the 09/2025 vs 06/2023 split you measured. New tests `test_vep_post_switches_host_for_grch37` and `test_version_get_switches_host_for_grch37` assert the actual URL called differs by host, not just that the assembly param was passed through.

**3. Tests don't test the fix (blocking).** Rewrote the mocks in `test_source_versions_captured_when_batch_succeeds` and `test_payload_missing_clinvar_entry` to capture `args`/`kwargs` and assert the URL equals `VEP_REST_HOST + ENSEMBL_INFO_VARIATION_PATH` and `timeout=10`, mirroring `test_transcript_version_param_sent`. Added `test_raise_for_status_real_failure_returns_none_and_warns`, which returns a plain object (not a `MagicMock`) whose `raise_for_status()` genuinely raises `HTTPError`, so `_fetch_ensembl_data_versions`'s own `resp.raise_for_status()` line is actually executed rather than short-circuited at `requests.get()`. Added `test_source_versions_empty_dict_when_annotation_succeeds_but_lookup_fails` for the annotation-succeeded/lookup-failed case.

**Smaller items:**
- `:439` gnomAD label reworded to `"v4.1 (hardcoded; Ensembl serves no gnomAD version endpoint)"`.
- `:282` now sends `Accept: application/json` instead of `Content-Type` on the GET.
- Added `_structured_data_source_versions`: `result.json` and `database_versions.json`'s `data_source_versions` field is now `{"clinvar": ..., "dbsnp": ..., "omim": ..., "gnomad": "v4.1"}` with `null` for missing/failed sources, instead of the prose strings. The Markdown table still gets the prose form from `_data_source_versions_for_report`, unchanged in spirit.
- **Not done: VEP release via `/info/data`.** Threading a third fetched value through `annotate_variants_vep` would change its 2-tuple return signature, which several existing tests and call sites unpack positionally; doing that cleanly (without silently entangling the release-fetch's success/failure with the ClinVar/dbSNP/OMIM empty-dict sentinel from item 1) looked like more surface area than "one call away" suggested. Left out rather than rushed. SKILL.md:193 still overpromises on this specific field; that gap is unchanged by this PR.
- `annotation_backend`'s GRCh38 hardcode and the SKILL.md scanner-header complaints: left alone, per "not your problem" / pre-existing in the review.

**Verified this session**, clean `python:3.12-slim` Docker container, commit 2e0f184:

```
$ docker run --rm -v <worktree>:/repo python:3.12-slim bash -c \
  'pip install -q pytest requests matplotlib numpy pandas scipy scikit-learn biopython && \
   cd /repo && python -m pytest skills/clinical-variant-reporter/tests/ -v'
...
============================== 71 passed in 3.07s ==============================
```

71 = 60 in `test_clinical_variant_reporter.py` (50 before this commit, +10 new) plus 11 in `test_self_audit.py`, untouched. I did not run `clawbio-bench`; installing `uv` and syncing the lockfile plus the bench package felt like more setup than the pytest suite warranted for this round, and the task said that's optional. Happy to run it if you want the harness numbers before merge.

The `test (3.12)` red tick predates this push and, per your note above, is the unrelated `just-prs-mcp`/`typer` CI bug, not this PR.

manuelcorpas:
## Audit: NEEDS REVIEW, one substantive fix then approve

Verified at head `2e0f184`. 0 CRITICAL, no patient-data leak, no ACMG perturbation, 71/71 tests pass.

### I checked the thing you said you could not

Your PR body says: *"Not checked: the live Ensembl endpoint's actual response shape."* I hit it live.

- `GET /info/variation/homo_sapiens` returns HTTP 200 and a list of 38 dicts with `name`/`version`, exactly the schema your fixtures use. Real values: ClinVar `09/2025`, dbSNP `156`, OMIM `09/2025`. Your `name_to_key` parse matches the real payload.
- **gnomAD is genuinely absent** from all 38 sources. Your comment "no live source exists" is true, and the honest `(hardcoded; ...)` label is the right call rather than asserting false confidence.
- **Your GRCh37 fix is real and load-bearing.** `grch37.rest.ensembl.org` returns ClinVar `06/2023` against GRCh38's `09/2025`, the 28-month gap your test docstring claims, confirmed live. Without that routing a GRCh37 run would certify a ClinVar version never served for its data. That catch was not asked for in the original issue and it is a genuine clinical-provenance improvement.
- Failure mode is honest. I empirically hit an Ensembl rate-limit non-JSON response on my first call: it degrades to `"unavailable (annotation succeeded, but the Ensembl version lookup failed)"`, with no crash and no silent fallback to a stale hardcode. That was the key requirement and it is met. Timeouts are bounded at 10s and 60s, so no hang path.
- The new request is a parameterless GET, so no patient data leaves the machine, and SKILL.md matches the code.

### One fix needed

**`annotation_backend` hardcodes GRCh38, which your own new fields now contradict.** `clinical_variant_reporter.py:840`

```python
"annotation_backend": "demo_cache" if demo else "Ensembl VEP REST (GRCh38)",
```

`assembly` is already a parameter of this function. I reproduced the artefact for a GRCh37 run:

```json
"annotation_backend": "Ensembl VEP REST (GRCh38)",
"data_source_versions": { "clinvar": "06/2023", ... }
```

while `report.md` alongside it says `| Ensembl VEP | REST API, assembly GRCh37 |`. That is a self-contradictory clinical audit record: the GRCh37 ClinVar version stamped GRCh38. The line predates your PR, but your PR is what makes GRCh37 a genuinely distinct, differently-versioned path, so it converts a dormant string into a live provenance falsehood. One line: `f"Ensembl VEP REST ({assembly})"`.

### Three smaller ones

**ClinVar version mislabels its own source.** `clinical_variant_reporter.py:488` renders `f"{clinvar_v} (via Ensembl VEP colocated variants)"`, but the version came from `/info/variation/homo_sapiens`, not from colocated variants. Same release so the value is right, but naming the wrong endpoint inside a provenance-correctness PR is the exact error class being fixed here.

**"VEP release" is documented but still never emitted.** `SKILL.md:193` promises `VEP release` in `database_versions.json`. ClinVar and gnomAD now arrive, but no Ensembl release number is recorded anywhere, and the PR title is "report live Ensembl release". It is one call away: `/info/data` returns `{"releases":[116]}`, verified live. It is also the single most reproducibility-critical provenance field for a VEP run.

**Dead constant re-arms the bug you just fixed.** `clinical_variant_reporter.py:49` `ENSEMBL_INFO_VARIATION_URL` is never referenced and is pinned to the GRCh38 host. Any future use reintroduces the GRCh37 falsehood. Worth deleting.

Minor: `DEMO_CLINVAR_LABEL = "2025-03-01 release"` (`:51`) is unsourced, since `example_data/demo_evidence_cache.json` records no version provenance, and it puts prose into `result.json` (`"clinvar": "2025-03-01 release"`) while live values are bare (`"09/2025"`), breaking the "real values or null, never prose" contract that function's own docstring asserts.

Also, your PR description still says the release is fetched from `info/software`, but the code uses `/info/variation/homo_sapiens`. Worth updating so a maintainer approving on the description approves what actually ships.

### Resolved since the last review

- **Wrong-endpoint provenance claim: resolved in code.** Endpoint, schema, field and both assembly hosts verified live.
- **Failure-state conflation: resolved.** `None`, `{}` and a populated dict now produce three distinct report labels, each tested.
- **Prose in machine-readable JSON: resolved** for live mode via `_structured_data_source_versions`.
- **Documented-output divergence: improved.** On main, `database_versions.json` contained no ClinVar date and no gnomAD version while SKILL.md claimed both. This PR makes that documentation true for the first time.

### Our problems, not yours

- **CI had never run.** Run `31313190828` sat at `action_required` with zero jobs, which displays as "no checks reported" and must not be read as a pass. I have approved it.
- **The 2026-08-08 red tick was our bug, not yours.** Your suite passed (61 passed); the failure was `just-prs-mcp` raising `ModuleNotFoundError: typer`, a skill you never touched, pulled in because the changed-skill step diffed against `pull_request.base.sha`. Fixed in #332, merged 2026-08-09.
- **`main` was red on `skill-lint`** for an unrelated skill's frontmatter, three days running. Fixed just now, so you should get a clean signal on re-run.
- **Our scanner's five "missing SKILL.md section" warnings are misattributed.** `## Trigger`, `## Scope`, `## Example Output`, `## Gotchas` and `## Agent Boundary` are absent on `main` too, and this PR changed exactly one line of SKILL.md. Ignore them.

### Clinical integrity

Confirmed clean. The diff touches only constants, the version fetch, return-type threading and report writers. No evidence-code logic, no thresholds, no combining rules, no `classify_variant`/`self_audit` changes. Signature changes are fully propagated. `DISCLAIMER` present in both `report.md` and `result.json`.

### Out of scope, worth a separate issue

`gnomad_af_popmax` is initialised and never reassigned in the live path, so it is always `None` in live mode despite popmax being the ACMG-relevant frequency for BA1/BS1. Not touched by this diff.

### What I could not verify

That `gnomAD v4.1` is correct for Ensembl release 116, since no API exposes it. That is precisely why your `(hardcoded; Ensembl serves no gnomAD version endpoint)` label is the right response. Also that `2025-03-01` matches the demo cache's true snapshot, since the cache records no provenance. No end-to-end run against a real patient VCF; live endpoints were exercised with synthetic coordinates only.

Fix the `annotation_backend` line and the label, and this is good to merge. Thanks for responding thoroughly to the earlier review.

AmirF194:
Fixed both required items in `ae73722`.

`annotation_backend` now stamps `f"Ensembl VEP REST ({assembly})"` instead of a hardcoded GRCh38, and the ClinVar label now names the endpoint that actually returns it (`/info/variation/homo_sapiens`, not "colocated variants"). Also dropped `ENSEMBL_INFO_VARIATION_URL`, the dead GRCh38-pinned constant you flagged, since it was exactly the kind of loose end that could reintroduce the GRCh37 bug later.

Added a regression test asserting `annotation_backend == "Ensembl VEP REST (GRCh37)"` on a GRCh37 run, and updated the existing endpoint-label assertion. Full suite: 72 passed. Rewrote the PR body to match what actually shipped, including the `/info/variation/homo_sapiens` correction you caught.

Left the VEP-release field and the demo-cache provenance string as separate concerns, since both change return signatures or touch data this PR doesn't otherwise have provenance for. Happy to open a follow-up issue for either if useful.

Thanks for the live verification and for catching the stale PR description.

manuelcorpas (commented):
Thanks @AmirF194. The core of this is right and I verified it live rather than by reading.

## The provenance fix is real

The wrong-endpoint problem from the earlier round **is fixed at head**. `clinical_variant_reporter.py:45,299` calls `/info/variation/homo_sapiens` and parses `src["version"]` keyed off `src["name"]`. I hit both hosts:

- `rest.ensembl.org` -> ClinVar `09/2025`, dbSNP `156`, OMIM `09/2025`
- `grch37.rest.ensembl.org` -> ClinVar `06/2023`

That 28-month split is exactly what the `_ensembl_rest_host` routing exists to prevent, and it works. These are the versions bundled with the Ensembl variation data the VEP call itself annotates against, so the provenance describes the data actually used. gnomAD is genuinely absent from all 38 entries in the payload, so labelling it `(hardcoded; ...)` is honest rather than a shortcut.

Also checked and clean: no variant data in the new request (parameterless GET), no network added to any offline path, demo mode asserted network-free, ACMG logic untouched by the diff, GET bounded at 10s, and no stale-hardcode fallback presenting an old date as live.

## Two things to fix before I approve

**1. The endpoint tests are constant-relative, so the bug you just fixed can be reintroduced with a green suite.**

Every URL assertion pins against the module constant:

```python
assert called_url == VEP_REST_HOST + ENSEMBL_INFO_VARIATION_PATH
```

Nothing pins the literal `"/info/variation/homo_sapiens"`. I mutation-tested it: changing line 45 to `ENSEMBL_INFO_VARIATION_PATH = "/info/software"` - the exact wrong endpoint blocked last round - leaves **72 passed, unchanged**.

It compounds, because the report label at `:487` is a decoupled hardcoded string:

```python
f"{clinvar_v} (via Ensembl /info/variation/homo_sapiens)" if clinvar_v
```

so under that mutation the report would name an endpoint it never called and no test would fire. Either assert the literal path once, or derive the label from the constant so the two cannot drift.

**2. `report.md` and `result.json` contradict each other when nothing annotates.**

`_structured_data_source_versions:511-516` does `sv = source_versions or {}` then unconditionally emits `"gnomad": GNOMAD_VERSION_LABEL`. Reproduced with `source_versions=None`:

```
report.md   | gnomAD | unavailable (no variant was successfully annotated this run) |
result.json {"clinvar": null, "dbsnp": null, "omim": null, "gnomad": "v4.1"}
```

Two artefacts of one run disagreeing on whether a gnomAD claim may be made. The prose path has it right; the JSON should be `null`.

## Not blocking

- `SKILL.md:193` documents `VEP release` in `database_versions.json`, but no Ensembl release number is emitted anywhere. `/info/data` returns `{"releases":[116]}`. Pre-existing, but the PR title makes it conspicuous - happy to take it as a follow-up issue.
- `:315` `return versions or None` reports "lookup failed" when a 200 response simply names none of the three sources. Unreachable against the real payload; cosmetic.

## On the tooling, not on you

The audit initially reported five WARNINGs against your SKILL.md for missing `## Trigger`, `## Scope`, `## Example Output`, `## Gotchas` and `## Agent Boundary`. Those were **my tool blaming you for repo state** - the sections are absent on `main` identically, and this PR changes one line of that file. Fixed on my side by attributing missing sections against the merge-base, so your warning count was entirely spurious. Sorry about that. Your actual SKILL.md change (the Safety bullet disclosing the Ensembl query) is accurate against the code.

Your CI had also never run - the fork approval gate had it parked in `action_required`, which surfaces as "no checks reported". I have approved it; results incoming.

manuelcorpas:
CI is green now that the run is approved: `test` on 3.11/3.12/3.13, `benchmark`, `lockfile-check`, `skill-lint` and `scientific-audit` all pass. The two items in my review above are the only things between this and an approval.

AmirF194:
Thanks for the thorough audit, both points fixed at fbc1265:

1. Added a literal-path assertion (`called_url == "https://rest.ensembl.org/info/variation/homo_sapiens"`) alongside the constant-relative one, so the mutation you found (reverting `ENSEMBL_INFO_VARIATION_PATH` to `/info/software`) now fails the suite instead of passing it unchanged. Verified locally: with the mutation applied, `test_source_versions_captured_when_batch_succeeds` fails on exactly that assertion.

2. `_structured_data_source_versions` now returns `gnomad: None` (not the hardcoded constant) when `source_versions is None`, matching report.md's "unavailable (no variant was successfully annotated this run)" for that same case. Added a regression test for it.

Full suite: 73 passed (72 prior + the new test). Left the two non-blocking notes as-is per your comment.

manuelcorpas (approved):
Both blocking items are fixed, and I verified each by mutation rather than by reading.

**Item 1, the endpoint pin.** `tests/test_clinical_variant_reporter.py:738` now asserts the literal alongside the constant:

```python
assert called_url == "https://rest.ensembl.org/info/variation/homo_sapiens"
assert called_url == VEP_REST_HOST + ENSEMBL_INFO_VARIATION_PATH
```

Re-running the exact mutation that previously went undetected, changing line 45 to `/info/software`, now gives `1 failed, 72 passed` on `test_source_versions_captured_when_batch_succeeds`. Before your fix it left 72 passing unchanged. The guard is live.

**Item 2, the gnomAD contradiction.** `clinical_variant_reporter.py:511-515` now returns all four keys as `None` when `source_versions is None`, and reverting it to `sv = source_versions or {}` fails `test_live_mode_nulls_gnomad_too_when_nothing_annotated`. `report.md` and `result.json` no longer disagree about whether a gnomAD claim may be made.

Baseline is 73 passed, and CI is green across `test` 3.11/3.12/3.13, `benchmark`, `lockfile-check`, `skill-lint` and `scientific-audit`.

Nice touch that both new tests carry the reasoning and the mutation result in their comments, so the next person to touch them knows what they are for rather than guessing. That is exactly the discipline that would have prevented the original constant-relative gap.

Two things I said earlier stand as non-blocking follow-ups rather than changes here: `SKILL.md:193` still documents a `VEP release` field that is never emitted, and `:315`s `return versions or None` conflates "lookup failed" with "payload named none of the three sources". Neither is reachable against the real payload; happy to take them as separate issues.

Thanks for the quick turnaround, and apologies again for the five spurious SKILL.md warnings in the first pass, which were my tooling blaming you for repo state. That is fixed on my side.

AmirF194:
Thanks for the mutation-testing rigor across both rounds, that's what actually caught the endpoint pin gap rather than just plausible-looking code. Appreciate the thorough audit.

PR #331, opened 2026-08-04
feat(deepspot-m): virtual spatial transcriptomics from H&E tiles

KalinNonchev:
## Summary

- Adds `deepspot-m`, which scores a 224x224 H&E tile with DeepSpot-M and writes per-gene log1p-CPM virtual spatial transcriptomics.
- `deepspotm` and `torch` import lazily inside the prediction function, so the model stack stays out of `pyproject.toml` and `uv.lock`. Dependencies live in a skill-level `requirements.txt`, following `cell-detection`.
- `--demo` reads a bundled offline fixture, so it runs with no weights, no GPU and no network. Its report is tagged "(demo)".
- Registry wiring: `clawbio/cli.py` SKILLS, `pytest.ini` testpaths, `MVP_FOLDERS` plus `CHAINING`, and a regenerated `skills/catalog.json`.

Requirements are stated positively in SKILL.md: tiles are exactly 224x224 at roughly 20x (about 0.5 microns per pixel), upstream code is PolyForm Noncommercial 1.0.0, and weights are CC-BY-NC-SA-4.0 and gated behind an access request plus `huggingface-cli login`. Nothing from upstream is vendored.

Paper https://doi.org/10.64898/2026.06.19.26356060, code https://github.com/ratschlab/DeepSpotM.

## Skill affected

new skill: deepspot-m

## Checklist

## Test output

```bash
python clawbio.py run deepspot-m --demo
uv run pytest skills/deepspot-m/tests -v
```


manuelcorpas (changes requested):
## Audit summary: changes requested

Thanks for this, and thanks for bringing DeepSpot-M over yourself. The engineering is solid: the SKILL.md is one of the better structured ones in the repo, demo mode is labelled honestly at four separate points, the gene ordering is bound to the model's own `genes` list rather than assumed positionally, and CI is green across 3.11, 3.12 and 3.13. There is no fabricated-output path anywhere, which is the thing we check hardest for.

The blocking items are about what the report claims, not about the code quality.

### Blocking

**1. The headline output is a cross-gene ranking, which the model card does not support.** `deepspot_m.py:151` sorts genes by predicted value and `:244` reports the top one as the headline. The card states the model "predicts relative expression rather than absolute counts", and notes under-sequenced genes are predicted less reliably. Ranking EPCAM against COL1A1 within one tile largely recovers each gene's training-set mean rather than tile-specific biology. Could you either report values unranked with a per-gene reference distribution (z-score or percentile against the training distribution), or, if the ranking stays, state plainly in the report that it approximates average abundance?

**2. The three upstream limitations are absent from SKILL.md.** The model card gives all three: trained on oncology cohorts and so not representative of healthy or non-oncology tissue; performance may degrade on unseen tissue types, stains, scanners or resolutions; relative rather than absolute expression. None appear in `## Gotchas` or `## Safety`. Meanwhile `SKILL.md:106` frames the skill against "sending tissue for a spatial assay, which costs weeks and thousands of pounds per sample", and no accuracy figure appears anywhere in the file. Please import the three limitations verbatim and soften that framing.

**3. Microns-per-pixel is asserted, never validated, and attributed to a source that does not state it.** `TARGET_MPP = 0.5` at `:53` is a display constant; `load_tile` at `:93` checks pixel dimensions only. Every report therefore prints 0.5 um/px as fact (`:211`, and into `result.json` at `:382`) for a tile that could be 40x or a downsampled thumbnail. `SKILL.md:180` cites "(source: DeepSpot-M model card)", and the card carries no MPP or magnification figure. Please either add a validated `--mpp` (or read the TIFF resolution tags) or remove the figure and the citation.

**4. `## Example Output` presents demo fixtures as a real run.** The values at `SKILL.md:195` are byte-identical to `examples/demo_expression.json`, but they appear under a `## Predicted Expression` heading with `**Tile**: tile.png` and no `(demo)` tag. That is the exact thing your own `## Agent Boundary` rule at `:275` prohibits. Regenerating the block from an actual `--demo` run, including the `(demo)` tag and the `## Output Files` table, would fix it.

### Worth addressing before merge

- **No out-of-distribution guard.** A blank, non-H&E, or non-oncology tile is scored and returned with full confidence. Upstream's `examples/predict.py:84` ships `--skip-background/--white-mean`; porting it, or at minimum warning on near-white or implausibly non-H&E tiles, would close the gap.
- **Unpinned weights.** `from_pretrained` at `:125` has no `revision=`, and `requirements.txt` has `deepspotm>=1.0` with no upper bound, so both float. A pinned commit SHA plus `<2` would make runs reproducible. A `--allow-download` opt-in before the first network fetch would also match the local-first posture of the other skills.
- **`C9orf72` is destroyed by uppercasing.** `:75` upper-cases every symbol, but HGNC uses lowercase `orf` in roughly 200 genes, so a correct input becomes unresolvable. Case-fold for lookup only, and emit the canonical symbol.
- **`KeyError` is not caught.** Upstream `genes_to_indices` raises `KeyError` for an off-panel symbol; the handler at `:372` catches `ValueError`, `FileNotFoundError`, `OSError` and `RuntimeError` but not that, so an off-panel gene produces a raw traceback rather than your documented message.
- **The live path has no test coverage.** `predict_expression` is exercised only by forcing an `ImportError` (`tests/test_deepspot_m.py:179`), so the actual inference call, the gene-to-score mapping and the tensor handling are untested and CI passes without weights present. A stubbed `deepspotm` injected into `sys.modules` would cover it.
- **Input paths propagate into shareable artefacts.** `--input /data/PATIENT_123/slide.png` is written verbatim into `report.md`, `result.json` and `reproducibility/commands.sh`.

### Two questions for the maintainer, not defects

- **Licence.** This would be the repo's first non-commercial skill (PolyForm-NC plus CC-BY-NC-SA-4.0) in an MIT repo where 91 of 96 skills are MIT, and `catalog.json` carries no `license` field, so the restriction is invisible to catalog consumers. That needs a maintainer decision rather than a code change.
- **Attribution.** `author:` is set to "ClawBio Contributors". Given you are the upstream first author, naming yourself would be more accurate and is worth declaring rather than flattening.

### Notes

`model.eval()` was flagged as an `eval()` call by our scanner. That was a false positive in our tooling, now fixed; no action needed from you. We also confirmed the HuggingFace repo ships `model.safetensors` rather than a pickled checkpoint, so there is no `torch.load` deserialisation concern.

Verified during this review: the medRxiv DOI resolves, the GitHub and HuggingFace repos exist, the PyPI package exists, and the Midnight backbone is built from the bundled config rather than fetched separately. Note that the HuggingFace repo is `gated: manual`, so "the download is automatic" at `SKILL.md:370` understates an approval step that can be refused.

KalinNonchev:
All four blocking items and every "worth addressing" item are fixed in 935eaca.

**Blocking**

1. **Ranking.** Genes now report in requested order, no top-gene headline. A `## How to Read These Values` section states that cross-gene ordering within a tile recovers average abundance, not tile biology. `rank` stays as a secondary CSV column with the caveat attached. No reference distribution: the release ships no per-gene training statistics, and I'd rather ship none than derive one.
2. **Limitations.** All six sentences of "Limitations and biases" quoted verbatim in `## Safety` and reproduced in every report. Cost framing removed. No accuracy figure added — the model card has no evaluation section and the abstract reports no correlations, so SKILL.md says that and points at the preprint.
3. **MPP.** `TARGET_MPP` deleted. The skill reads TIFF/PNG resolution tags, accepts `--mpp`, else records `null` / "not declared". Declared wins over tags, and the report says when they disagreed. The 20x citation now points at the README (`### Command line`); the MPP citation is gone.
4. **Example Output.** Replaced with the byte-for-byte `report.md` of a real `--demo` run. A test asserts it still matches, so it can't drift.

**Worth addressing**

- Ported `--skip-background` / `--white-mean` (220 default) plus a saturation check. Warnings land in `report.md` and `result.json`, not just stderr.
- Pinned `revision="86113ee431248c892d25cf55e1f8017cccec2926"`, `deepspotm>=1.0,<2`, and added `--allow-download` — local cache only without it.
- `C9orf72` confirmed: 236 lower-case-`orf` symbols in `tokens.csv`, zero case-fold collisions. No more upper-casing; lookup case-folds against `model.gene_names` and reports the panel's spelling.
- `KeyError` caught, with symbols resolved against the panel first so the documented message fires.
- Live path covered by a stub `deepspotm`/`torch` in `sys.modules`.
- `report.md` and `result.json` record the file name only. **`commands.sh` still keeps the full path** — deliberate, since replay needs it, matching `cell-detection`. `## Safety` names it as the file to scrub. Happy to redact it if you'd rather.

**Your two questions**

- **Attribution.** `author:` is now `Kalin Nonchev`. Stating it plainly: I'm first author of DeepSpot-M and the upstream package, so I'm not neutral on whether this belongs in ClawBio.
- **Licence.** `catalog.json` now carries `license` from frontmatter — 91 MIT, 1 PolyForm-NC, 1 GPL-3.0, 1 Apache-2.0, 1 PROPRIETARY, 1 undeclared.

One flag before you rule: `WEIGHTS_LICENSE.md` applies NonCommercial to "the weights **or their outputs**", so the values this skill writes are themselves non-commercial and need attribution. Real runs now stamp that on `report.md` and `result.json`; demo runs don't. `THIRD_PARTY_LICENSES.md` confirms the backbone and all five embedding sources are MIT or Apache-2.0, so the restriction is DeepSpot-M's own.

Also fixed the `gated: manual` wording — approval is a human decision that can be refused.

60 skill tests, 154 repo tests pass.

manuelcorpas:
## Audit: SAFE, approve after two small reproducibility fixes

Verified at head `935eaca`. No critical findings. This is the best-sourced skill PR submitted to this repo so far, and I want to be specific about that rather than vague, because every external claim I could independently check, checked out:

| Claim | Result |
|---|---|
| DOI `10.64898/2026.06.19.26356060` | Resolves via Crossref: medRxiv, 2026-06-22, title and all five authors exactly as cited |
| "28,664 slides across 32 cancer types" | Verbatim in upstream `CITATION.cff` |
| Pinned revision `86113ee4...c2926` | Equals the live Hugging Face API `sha` exactly; `gated: "manual"` confirms the access-request claim |
| Six "Limitations and biases" quotes | Verbatim against the live model card |
| "Model card publishes no accuracy figure" | Confirmed, and the skill correctly quotes none |
| Licensing (PolyForm NC 1.0.0 / CC-BY-NC-SA-4.0 / outputs / ShareAlike) | All verbatim against upstream `LICENSE` and `WEIGHTS_LICENSE.md` |
| Example Output "verbatim" | Diffed against a live `--demo` run: **byte-identical** apart from the timestamp |

No fabricated data anywhere: no `np.random`, the fixture JSON self-labels "not model output", and a test asserts demo output claims no weights licence. No patient-data movement: the only network path is the HF weight fetch behind `--allow-download`, with `HF_HUB_OFFLINE=1` forced otherwise and tested. The demo tile is visibly synthetic, not histology. 60/60 tests pass locally.

### Two things worth fixing before merge

**1. The reproducibility bundle does not replay the run it documents.** `deepspot_m.py:633-634`

```python
    if meta["demo"]:
        cmd_args.append("--demo")
```

The `else` branch records `--genes`; the demo branch does not. Confirmed by execution: `--demo --genes MKI67,EPCAM,CD68` writes a 3-gene `report.md` beside a `commands.sh` reading `--demo --output "$OUTPUT_DIR"`, which on replay produces the default 10-gene panel. The bundle's own comment says "Replays this deepspot-m run". Moving the `--genes` append above the branch fixes it. Currently untested; the only test touching `commands.sh` is the existence check at `tests/test_deepspot_m.py:427`.

**2. Tile-check thresholds are never recorded.** `deepspot_m.py:635-644` records `--input/--genes/--source/--mpp/--skip-background/--allow-download` but not `--white-mean` or `--min-saturation`. A run that raised `--white-mean` to accept a pale tile replays at the 220 default and, with `--skip-background`, exits 1.

### One maintainer decision, not a defect

`SKILL.md:4` declares `license: PolyForm-Noncommercial-1.0.0` in an MIT repo. This would be our first NonCommercial skill. The NC term genuinely rides into outputs: upstream `WEIGHTS_LICENSE.md` says no commercial use of the weights "or their outputs", so `result.json` from real runs is encumbered. The PR already mitigates this by surfacing `license` in `catalog.json`, and precedent exists for non-MIT skills (`fastreer` GPL-3.0, `wes-clinical-report-en` PROPRIETARY), but not for NonCommercial. That is ours to decide, not yours to fix. Separately, `deepspot_m.py` carries no per-file licence header, so which licence governs your own wrapper code is currently undetermined; a header would settle it.

### Our bugs, not yours

Our scanner reported three CRITICALs on this PR. **All three are false**, and I am recording them here so they are not held against the PR:

- "`eval()` call" is `def eval(self):` at `tests/test_deepspot_m.py:85`, a method **definition** stubbing PyTorch's mode switch.
- "`__import__()` call" is `__import__("os").environ.get("HF_HUB_OFFLINE")`, a stdlib lookup in a test recorder.
- "tests assert output paths absent from SKILL.md: SKILL.md" is the gate treating the tests' own `SKILL.md` read as an output artifact. Your real output contract is exact: the six paths at `tests/test_deepspot_m.py:422-429` match `SKILL.md:284-294` and match what a live `--demo` writes, byte for byte.

Both scanner bugs are being fixed. Also, CI had never actually run on this PR (the run was parked at `action_required`, which displays as "no checks reported"); I have approved it. `main` was separately red on `skill-lint` for an unrelated skill, now fixed, so you should get a clean signal on re-run.

### Notes, no action needed

- No per-gene uncertainty: upstream says under-sequenced genes are predicted less reliably, but every gene renders as an equal-weight 2-dp point estimate (`deepspot_m.py:573`). The prose compensates well, so this is not overclaiming, but a per-gene reliability flag is the natural next version.
- `deepspotm>=1.0,<2` is a range rather than a pin, though the model revision itself is pinned exactly.
- `--demo --input patient.png` silently ignores the input and reports fixture values. Clearly tagged `(demo)`, so not a safety issue.

### What I could not verify

The weights are gated behind manual approval, so the live path is exercised only through an in-test stub. `DeepSpotM.from_pretrained(...)`, `model.gene_names` and `predict_genes` return shape are checked against the upstream README but never executed. If upstream's API differs, nothing in this PR catches it. Worth a note in Gotchas.

Thanks for declaring the upstream authorship in `SKILL.md:7` and enforcing it with a test. That is the right way to handle it.

KalinNonchev:
@manuelcorpas merge?

manuelcorpas (approved):
Approving. This is the strongest skill submission the repo has had, and several failure classes that usually turn up in skill PRs are pre-empted explicitly in the code rather than by accident.

Verified at head `935eaca` by running it, not by reading it:

- Test suite `60 passed`; `--demo` output `report.md` is byte-identical to the SKILL.md Example Output apart from the timestamp, so that block is a real rendered sample.
- Regenerated `skills/catalog.json` with `scripts/generate_catalog.py` and diffed against the committed file: identical. The catalog was not hand-edited.
- `MODEL_REVISION` pins commit `86113ee`, which the HF API confirms is the current `sha` for `ratschlab/DeepSpotM`, and I checked in the published `deepspotm==1.0.0` wheel that `revision` is a real kwarg forwarded to `hf_hub_download`. A commit-SHA pin on a content-addressed repo is an adequate supply-chain pin.
- Default-deny on network: `HF_HUB_OFFLINE=1` unless `--allow-download`, asserted in both directions. No `requests`, `urllib` or `socket` anywhere in the skill.
- No `np.random`. The demo fixture self-labels as fixture and `write_result_json` deliberately withholds the weights-licence stamp on demo runs because those values never touched the weights - with a test asserting the absence.
- Every provenance claim checked against upstream: the medRxiv DOI resolves, the 224x224/~20x tiling, the `--white-mean` 220 default, the ~19,000 gene panel and the log1p-CPM unit all match the model card or README. SKILL.md correctly refuses to quote a per-gene accuracy figure because upstream publishes none.

Two things I want to single out.

`tile_label()` strips the directory from the tile path so `report.md` and `result.json` cannot leak a patient or case directory name, with a test using a `PATIENT_12345` path. Nobody asked for that.

And the skill guards the interpretation error that actually matters for a relative-expression model: ranking genes within a single tile largely recovers each gene average abundance in the training data, not biology. Leading the report with requested order and demoting `rank` to a CSV column - enforced by `test_the_report_no_longer_headlines_the_highest_gene` - is the correct scientific read.

The out-of-folder changes are all legitimate: the `cli.py` registry entry uses a pre-existing key already used by four other skills, the `pytest.ini` line is additive, and the `catalog.json` diff is 95 lines of a new `license` field appearing on existing skills plus the new entry. Surfacing `license` was the right call rather than scope creep - it was already in every SKILL.md frontmatter and simply was not exposed.

**One question for me to settle, not a change request.** `SKILL.md:4` declares `license: PolyForm-Noncommercial-1.0.0` in a repo whose root `LICENSE` is MIT. The wrapper is your code so the choice is yours, and given the weights are `cc-by-nc-sa-4.0` it is arguably the only coherent choice. But it means ClawBio ships a file commercial downstream users cannot use. There is precedent already in the catalog (2 PROPRIETARY, 1 GPL-3.0) and this PR surfaces licences rather than hiding them, which is the good-faith move. I will add a note to the root `LICENSE` that the repo is MIT except where a skill declares otherwise, so nobody infers blanket MIT from the top-level file.

Two trivia, neither blocking: `deepspot_m.py:376` calls `model.eval()`, which the wheel `from_pretrained` already does at `model.py:329`, so it is redundant; and the demo silently overrides a non-default `--source` (correctly, since the fixture was recorded under scgpt) without SKILL.md mentioning it.

manuelcorpas:
Following up on the licensing question from my review, I have now written the policy down rather than leaving you to guess: #339.

The short version is that ClawBio was collapsing two different things into one `license` field, and your PR is what made that visible. Splitting them:

- `license:` covers **your wrapper code**
- `model_license:` covers **the weights the skill downloads**

For deepspot-m that would be:

```yaml
license: MIT
model_license: cc-by-nc-sa-4.0
```

**The ask: would you relicense the wrapper MIT and declare the weights separately?**

The reasoning, and please push back if you disagree. The non-commercial protection you presumably want is already carried by the weights themselves, which are `cc-by-nc-sa-4.0` and which no licence in this repository can override. PolyForm on the wrapper therefore adds a second restriction on top of one that already binds, while extending it to code that did not need it: your tiling, TIFF resolution decoding, offline guard and path scrubbing are genuinely reusable independently of DeepSpot-M, and under PolyForm nobody can reuse them.

Meanwhile the catalog would keep reporting the real constraint, because `model_license` carries it. A commercial user gets the correct answer either way; the difference is only whether your wrapper code is reusable.

The wider reason it matters is that a single non-permissive file triggers legal review of the whole repository rather than of one skill, and the institutions we most want depending on ClawBio are exactly the ones with that review process.

That said, it is your code and your call. If you would rather keep PolyForm, say so and it stays as a declared exception; #339 keeps that route open and the catalog will surface it honestly. I would rather have deepspot-m in ClawBio under PolyForm than not have it.

One mechanical thing: #339 adds `"license": str(skill_meta.get("license", ""))` to the same dict your PR touches in `scripts/generate_catalog.py`, so once it lands please rebase and drop that line from your diff, since the field arrives with mine. Sorry for the churn; your PR is the reason the gap was found.

Nothing else is blocking. The approval stands.

KalinNonchev:
The DeepSpotM code is 

Code, [PolyForm Noncommercial 1.0.0](https://github.com/ratschlab/DeepSpotM/blob/main/LICENSE). Non-commercial use only.
Model weights, [CC-BY-NC-SA-4.0](https://github.com/ratschlab/DeepSpotM/blob/main/WEIGHTS_LICENSE.md). Non-commercial, ShareAlike, with attribution.

KalinNonchev:
ClawBio wrapper could be MIT

manuelcorpas (changes requested):
## Correcting my approval: two blockers at the current head

I approved this on 13 August at `935eaca`. The head has since moved to `3eb540a1`, the merge of `main` on 15 August, and I had not re-audited it. Two blockers live at that new head, one of which is ours rather than yours. Apologies for the reversal.

### 1. The `--allow-download` gate does not work

`deepspot_m.py:319-334, 348-364`

`_hub_offline()` sets `HF_HUB_OFFLINE` at line 326, but `huggingface_hub` reads that variable once, at import, into `constants.HF_HUB_OFFLINE`. Setting it afterwards has no effect. Reproduced on hub 1.9.0:

```
after import, then os.environ['HF_HUB_OFFLINE'] = '1':
   constants.HF_HUB_OFFLINE = False
   is_offline_mode()        = False

set before import:
   constants.HF_HUB_OFFLINE = True
```

The ordering makes it unavoidable: line 350 runs `from deepspotm import DeepSpotM` before the context manager opens at line 361, and `deepspotm/model.py:6` imports `peft`, which imports `huggingface_hub`. Upstream's loader calls `hf_hub_download(repo_id_or_path, fn, revision=revision)` with no `local_files_only`, so that environment variable was the only thing between a default run and the network.

That makes `SKILL.md:311` ("nothing reaches the network without it") and `SKILL.md:333` ("the model loads from the local Hugging Face cache unless `--allow-download` is passed") inaccurate as implemented. No tile is transmitted, so this is silent fetching rather than data egress, but it is the guarantee the Safety section leads with.

Upstream accepts a local directory (`model.py:280`, `if os.path.isdir(repo_id_or_path)`), so pre-resolving the three files with `hf_hub_download(..., local_files_only=not allow_download)` and passing the resulting directory to `from_pretrained` should close it.

The test at `tests/test_deepspot_m.py:239` asserts that the environment variable was set at call time. That is true, and it does not tell you whether the library honoured it.

### 2. The catalogue is inconsistent at this head, and this one is our fault

`skills/catalog.json` has `skill_count: 96` against 97 entries, so `tests/test_generate_catalog.py::test_checked_in_catalog_is_current` fails. `main` is 96/96 and consistent; the merge brought in the 97th skill without the count. `deepspot-m` is also the only entry of the 97 missing `model_license` and `data_license`.

`gh api .../check-runs` returns 0 for this head, so CI has never run against it. I should have caught that before approving.

### 3. Licensing

Following your note on 15 August: `CONTRIBUTING.md:137-140` asks contributors not to put a non-commercial licence on their own wrapper code. So `license: MIT` in the frontmatter plus `model_license: cc-by-nc-sa-4.0` is the right declaration, which matches what you suggested.

### Non-blocking, worth a look

- `SKILL.md:211` quotes the 28,664-slide, 32-cancer-type atlas. Upstream's README notes those predictions used cancer-specific finetuned models, whereas this skill pins the base checkpoint in zero-shot mode. Sitting under "Key parameters", it reads as a claim about what the user is about to run.
- `write_reproducibility` emits `--skip-background` but not `--white-mean` or `--min-saturation`, so replaying a bundle can refuse to score where the original scored, while `report.md` calls it "Exact command that produced this run".
- The Hugging Face gate terms are stricter than `SKILL.md:310` describes. They exclude anyone holding a concurrent commercial affiliation and prohibit internal evaluation, benchmarking and proof-of-concept work in a commercial setting. That bears on whether we can put this through ClawBench, so I wanted it on the record now rather than discovered later.

### Credit where it is due

The provenance in this skill is the best the repo has had. The DOI resolves to the exact title and all five authors, `tokens.csv` holds 19,338 gene tokens against the stated "roughly 19,000", `WHITE_MEAN_DEFAULT = 220.0` matches upstream exactly, all six upstream limitations travel into every report and into `result.json`, and no accuracy or performance figure is quoted anywhere, which was the right call. No fabricated data, no patient-data movement, and the tile label is stripped of its directory and tested against a `PATIENT_12345` path.

Close the download gate and I will re-run and merge.

KalinNonchev:
Both blockers closed at `3b4ab02`.

### 1. The download gate

Confirmed your diagnosis against `huggingface_hub` 1.28.0: `constants.py:192` computes `HF_HUB_OFFLINE` at import and `is_offline_mode()` just returns that constant, so setting the variable at call time never reached the library.

The skill now resolves the three files itself and passes the flag explicitly, then hands upstream the directory rather than the repo id:

```python
resolved = [
    hf_hub_download(
        MODEL_REPO, filename,
        revision=MODEL_REVISION,
        local_files_only=not allow_download,
    )
    for filename in MODEL_FILES
]
```

Once those are cached the load itself reaches no network: `from_pretrained` takes a local directory (`model.py:280`) and sets `backbone_pretrained=False` (`model.py:298`), and that branch of `MidnightEncoder._build` constructs from the bundled `deepspotm/assets/midnight_config.json` instead of fetching `kaiko-ai/midnight`. Every other download in `image_encoder.py` already pins `local_files_only=True`.

The revision pin moved onto the download, which is the call that can leave the machine; a directory has no revision to pass on.

The test now asserts what was passed to `huggingface_hub` rather than what was in the environment. All five gate tests fail against the previous implementation.

### 2. Catalogue

Regenerated: 97/97, and `deepspot-m` carries `model_license` / `data_license` like the other 96. Also dropped a duplicated `"license"` key this PR had added twice to the same entry literal in `generate_catalog.py`.

### 3. Licensing

Frontmatter is now `license: MIT` with `model_license: cc-by-nc-sa-4.0`.

### Non-blocking

- **Atlas figure**: qualified where it appears — the 28,664 slides came from cancer-specific finetuned models, whereas this skill pins the base checkpoint zero-shot.
- **Reproducibility**: `--white-mean` and `--min-saturation` are now emitted when non-default, so a replay cannot refuse to score a tile the original scored.
- **Gate terms**: written into the licensing section, quoting the "exclusively academic or public non-profit" eligibility condition and naming internal evaluation, benchmarking and proof-of-concept work. Worth settling before ClawBench rather than after.

### Not touched

`index.html` still reads 96 skills and "+ 91 more"; the regenerated count makes that 97 and 92. Left for the site-stats agent unless you would rather it rode along here.

CI now has a head to run against.

KalinNonchev:
@manuelcorpas merge?

manuelcorpas (changes requested):
This is the most carefully built skill I have reviewed in this repo. Pinned 40-character commit SHA with the length asserted in a test, safetensors rather than pickle, patient-path scrubbing tested against a `PATIENT_12345` directory, licence propagation that correctly refuses to stamp fixture output, upstream limitations travelling into every report, and full SKILL.md conformance on all 17 checks. The offline gate is genuinely enforced rather than nominal: resolving the three files yourself and handing upstream a local directory instead of a repo id is exactly what stops upstream's ungated loader fetching, so `SKILL.md:314`'s claim that "nothing reaches the network without it" is substantiated. I checked the Hugging Face repo listing and the gate over those three filenames is complete.

The tests are not tautological either. Answering the question I most wanted answered: a model returning zeros **would** turn them red, because `tests:239` asserts exact panel-index-derived values.

Requesting changes on two cheap defects with expensive failure modes.

**1. The gene CSV carries no provenance and no demo marker.** `deepspot_m.py:498-507`:

```python
writer = csv.DictWriter(handle, fieldnames=["gene", "expression", "unit", "rank"])
```

`report.md` and `result.json` are exemplary here: the heading switches between "Fixture Expression" and "Predicted Expression" at `:590`, the title carries `(demo)` at `:549`, and `result.json` stamps `demo`, `model` and `model_revision` at `:516-523`. The CSV carries none of it, and the column is called `expression`.

This matters because the CSV is designed to be detached; `SKILL.md:354` names the consumer: "`diff-visualizer`: downstream. The gene CSV feeds heatmaps and dot plots." So an agent runs `--demo` to inspect the format, passes `tables/gene_expression.csv` on, and ten handwritten fixture numbers become a heatmap with nothing in the file, the plot or the caption recording that no model was ever loaded. `tests:570` pins the header exactly, which locks it in.

Suggest renaming to `predicted_expression_log1p_cpm`, adding `model` and `model_revision` columns, and in demo mode either adding `provenance=demo_fixture` or writing to `gene_expression.DEMO.csv`.

**2. `dict(zip(canonical, scores))` has no length or order check.** `deepspot_m.py:405-412`:

```python
canonical = resolve_genes(genes, model.gene_names)
...
values = model.predict_genes(batch, canonical)
scores = [float(v) for v in values.squeeze(0).tolist()]
return dict(zip(canonical, scores))
```

Two unguarded assumptions about third-party code: that `predict_genes` returns values in request order rather than panel-index order, and that it returns exactly `len(canonical)` of them. `zip` truncates silently, so a length mismatch yields a short result rather than an error, and there is no length assertion anywhere in the skill or tests.

If upstream ever resolves through `genes_to_indices` and gathers in sorted index order, which is the natural vectorised implementation and which your own comment at `:115-117` confirms is the upstream API, then `--genes MKI67,EPCAM,CD68` returns `[EPCAM, CD68, MKI67]` zipped onto `[MKI67, EPCAM, CD68]`. Every gene receives another gene's value. Units are right, magnitudes plausible, ranks clean, reproducibility bundle valid, nothing fails. `requirements.txt` permits `deepspotm>=1.0,<2`, so a minor release can introduce this silently.

The test that appears to cover it is your own fake at `tests:96`, written to return in request order, so it encodes the assumption it purports to verify. Suggest asserting `len(scores) == len(canonical)` and pinning the order contract with a two-gene reversed round-trip.

Worth raising but not blocking:

- **`--mpp` is collected and never checked.** The 224x224 pixel check at `:179-186` is strict with no silent resize, which is better than most. But pixel count is not field of view: a 224x224 tile at 40x covers a quarter of the area of one at 20x and passes identically. `PLAUSIBLE_MPP_RANGE = (0.05, 5.0)` admits both, and `--mpp 0.25` renders into the report with no warning. A one-line warning outside roughly 0.4 to 0.6 would close the single most likely real-world misuse, and you already have the datum in hand.
- **No uncertainty of any kind.** A grep for confidence, interval, std, variance, OOD or calibration across the skill returns nothing; every gene gets a bare point estimate. `assess_tile` is honest that it is an image-level check ("Neither decides whether the tissue is in distribution"), and `SKILL.md:323` states the limitation plainly, which I credit. But a well-stained brain or normal-kidney tile passes both checks and returns confident log1p-CPM values from an oncology-trained model. Even a per-gene `null` confidence key would stop downstream consumers reading absent uncertainty as high confidence.
- `requirements.txt` omits `huggingface_hub` and `torch`, both imported directly at `:339` and `:375`. It fails closed via the ImportError catch at `:388`, so not a hole, but the operator sees a missing library reported as a cold cache.
- No stain normalisation, and SKILL.md does not mention it, despite upstream naming unseen stains and scanners as a degradation mode. Delegated silently to upstream's `image_processor`.

Two things I checked and cleared, worth recording because a previous audit of this PR got them wrong: `model.eval()` at `:403` is PyTorch's mode switch and `def eval(self)` at `tests:86` is the fake implementing it, neither is the builtin; and there is no `__import__` in this skill at all. Apologies for that earlier noise. I also checked the `pytest.ini` edit for blast radius and it is a single additive line inside `testpaths`, with no ignore, marker or `addopts` change; `generate_catalog.py` likewise only adds to `MVP_FOLDERS`.

Output contract matches across SKILL.md, the writers and the tests. No `np.random`, no numpy at all, and `demo_expression.json` self-labels as `"kind": "demo_fixture"` with an explicit note that the numbers are not model output, which is the correct handling.

KalinNonchev:
Both blockers are closed at `0714f5f`, along with the `skill-lint` failure that was red at `c042953` and is unrelated to your review.

## 1. The gene CSV now carries provenance

You were right that this was the one output that could travel without its context. Every row now carries the same stamp `result.json` does:

```
gene,expression_log1p_cpm,unit,rank,provenance,model,model_revision
EPCAM,5.82,log1p-CPM,1,demo_fixture,ratschlab/DeepSpotM,86113ee431248c892d25cf55e1f8017cccec2926
```

Two deliberate departures from what you suggested, both small:

- **`expression_log1p_cpm`, not `predicted_expression_log1p_cpm`.** On a `--demo` run nothing was predicted, so `predicted_` would contradict `provenance=demo_fixture` in the next column. The unit was the ambiguity worth fixing; the provenance column carries the rest.
- **The `provenance` column rather than `gene_expression.DEMO.csv`.** You offered either. The column keeps one documented path in SKILL.md and in the reproducibility bundle, and it survives a rename, which a filename does not.

The stamp repeats per row rather than sitting in a header comment, so a `read_csv` that drops comment lines keeps it. `tests:570` now pins the full header, and two new tests assert `demo_fixture` on a demo run and `model_prediction` on a real one.

## 2. `dict(zip(...))` — assumption removed rather than tested

I did not want to pin a contract I cannot enforce, so the assumption is gone instead of asserted. The request is sorted into panel-index order before the call and the caller's order restored afterwards:

```python
canonical = resolve_genes(genes, model.gene_names)
ordered = panel_order(canonical, model.gene_names)
...
values = model.predict_genes(batch, ordered)
scores = [float(v) for v in values.squeeze(0).tolist()]
if len(scores) != len(ordered):
    raise RuntimeError(...)
scored = dict(zip(ordered, scores))
return {gene: scored[gene] for gene in canonical}
```

Handing upstream a list that is already in panel-index order makes "request order" and "panel-index order" the same list, so the mapping is correct under either convention and stays correct if a `deepspotm` 1.x release switches. The length check raises rather than letting `zip` truncate.

On your point that my fake encoded the assumption it was verifying: fair, and it still does, so it is no longer the only test. `test_a_model_returning_panel_index_order_maps_onto_the_right_genes` uses a second fake that gathers by panel index and ignores arrival order — the failure you described, made real. It fails against the old `dict(zip(canonical, scores))` and passes now. `test_upstream_is_asked_in_panel_index_order_whatever_the_caller_typed` is the reversed two-gene round trip you asked for.

## 3. `skill-lint` was red at `c042953`

Separate from your review, and worth flagging because it is a repo-level contract, not a `deepspot-m` one. `agentskills validate` allows six top-level frontmatter keys — `allowed-tools`, `compatibility`, `description`, `license`, `metadata`, `name` — and fails a skill on anything else. So `model_license:` beside `license:` cannot validate, and `deepspot-m` is the first skill to actually use the field.

`CONTRIBUTING.md:124-128` documents the placement that fails. I moved the declaration under `metadata`, taught `normalize_skill_metadata` to read both placements so no existing skill has to move, and corrected `CONTRIBUTING.md`. `catalog.json` regenerates byte-identical apart from the version bump to `0.3.0` for the CSV schema change. Two tests in `tests/test_generate_catalog.py` cover the nested placement.

## Non-blocking items, four taken

- **`--mpp` is now read.** Outside 0.4–0.6 the run warns, on stderr and in the report, and scores anyway. The message states the covered field of view in microns, says the pixel dimensions cannot distinguish the two, and is explicit that the band is a scanner convention for "~20x" rather than a figure from the model card — upstream still publishes none, and `test_the_module_asserts_no_default_microns_per_pixel` still holds. It is kept out of `assessment["warnings"]` on purpose, since `--skip-background` refuses to score on those and this is a caveat, not grounds for withholding a result.
- **Uncertainty is now stated rather than absent.** `result.json` carries `per_gene_uncertainty: null` plus a note that null means none was computed, not that the estimate is certain. It is not a calibration story, but it stops a downstream consumer reading silence as confidence.
- **`requirements.txt` declares `huggingface_hub` and `torch`.** Both are imported directly, and undeclared they surfaced as an `ImportError` an operator reads as a cold weight cache.
- **Stain normalisation** is now named in SKILL.md as something neither this skill nor upstream's loader applies, with unseen stains and scanners called out as a real source of drift.

Two I have left alone, and I would rather say so than quietly drop them. The 28,664-slide atlas sentence sitting under "Key parameters" is a fair reading — it describes the atlas, not the zero-shot base checkpoint this skill pins — but rewriting it is a judgement call about what to claim, and I would rather agree the wording with you than guess. Same for the Hugging Face gate terms: your reading of the concurrent-commercial-affiliation and internal-evaluation clauses is on the record now and bears on ClawBench more than on this skill, so I have not tried to compress it into SKILL.md's Safety section.

## Verification

Skill suite 65 → 77 tests, all green, plus `tests/` at 171. I have **not** seen CI green at this head: the run for `0714f5f` is sitting at `action_required`, which is how fork PRs queue here — the same reason `check-runs` returned 0 at `c042953`. Approving that run is the check I cannot do myself. Locally I reproduced the `skill-lint` frontmatter rule across all 97 skills and none now carry a disallowed top-level key.

KalinNonchev:
@manuelcorpas merge?

manuelcorpas (approved):
Re-audited at head `b544fd72`. Both blocking items are genuinely fixed, and the ordering fix is better than what I suggested.

**Gene ordering.** I proposed a length assertion. You did the stronger thing and removed the assumption instead: `panel_order()` sorts the request into the model's own panel-index order before the call, so request order and panel-index order become the same list and the mapping is correct under either upstream convention. Length equality would not have detected a permutation; this does. The `RuntimeError` on a length mismatch is a good addition on top, with a message that explains the hazard rather than just asserting.

The regression test earns its place. `tests/test_deepspot_m.py:275-300` installs a fake that ignores arrival order entirely and gathers by panel index, so it disagrees with the old implementation rather than agreeing with itself. Traced by hand against the old code: `["MKI67", "EPCAM"]` would have produced `{"MKI67": 0.0, "EPCAM": 4.0}`, the exact permutation, and the assertion fails. The reversed round trip pins the same contract from the caller's side.

**CSV provenance.** Seven columns now, `expression_log1p_cpm` naming its unit, and `provenance` set per row to `demo_fixture` or `model_prediction`, so a demo run is distinguishable on every row rather than in a header comment `read_csv` would drop. The assertion that previously locked the defect in is gone, and the two new tests pin the demo and real cases from opposite directions with hardcoded strings, read through `csv.DictReader` so a renamed column raises.

I also checked the lint change was behaviour-neutral empirically rather than taking the commit message: both parser versions run over all 98 SKILL.md files with PyYAML forced off give exactly one difference, the intended one. `pytest.ini` and `generate_catalog.py` are purely additive, `EXCLUDED_FOLDERS` untouched, and `catalog.json` has one removed line across the whole diff.

All five properties I credited last time still hold: the 40-character `MODEL_REVISION` with its length asserted, `local_files_only=not allow_download` asserted both ways, the resolved-directory-not-repo-id call that stops upstream's ungated loader fetching, safetensors with no `torch.load` anywhere, and the patient-path scrubbing tested against a `PATIENT_12345` directory.

The `--mpp` field-of-view check was a non-blocking suggestion and you built it. Keeping it out of `tile_warnings` so `--skip-background` does not start refusing on it is the right call.

Merging. Three nits for whenever, none blocking:

1. Demo rows still populate `model` and `model_revision` even though no model was loaded. `provenance` disambiguates, but a consumer grouping by `model` without reading `provenance` would attribute ten fixture values to a named revision. Blanking those two on a demo row closes it.
2. The SKILL.md frontmatter dependency block still lists two packages where `requirements.txt`, `PIP_DEPS` and the SKILL.md prose now list four. `catalog.json` is generated from the frontmatter, so it currently advertises an incomplete list for the very omission this commit set out to fix.
3. Nothing pins the SKILL.md column table against `GENE_TABLE_FIELDS`, so docs and writer can drift apart silently.

KalinNonchev:
nice!

Issue #333, opened 2026-08-11
nfcore-rnaseq-wrapper: --pipeline-local rejects valid checkouts (version regex matches custom_config_version)


javrodriguez:
## Summary

`nfcore-rnaseq-wrapper` rejects a valid local pipeline checkout, reporting its version as `master`. The version parser matches `custom_config_version` instead of `manifest.version`, so `--pipeline-local` fails for effectively every nf-core pipeline.

Affects `clawbio==0.6.1` (verified in the published wheel at `clawbio/skills/nfcore-rnaseq-wrapper/pipeline_source.py`).

## Reproduction

```bash
git clone --depth 1 --branch 3.26.0 https://github.com/nf-core/rnaseq.git rnaseq-3.26.0
python nfcore_rnaseq_wrapper.py \
  --input samplesheet.csv --output ./out \
  --pipeline-local ./rnaseq-3.26.0 \
  --genome GRCh38 --check
```

Result:

```
stage:   validation
code:    PIPELINE_SOURCE_INVALID
message: Local nf-core/rnaseq checkout is version 'master', but this wrapper's
         validations are pinned to 3.26.0.
details: {"manifest_version": "master", "contract_version": "3.26.0"}
```

The checkout is genuinely 3.26.0 — `git describe --tags` reports `3.26.0`, and its `nextflow.config` manifest block declares `version = '3.26.0'`.

## Root cause

`pipeline_source.py:26`:

```python
_MANIFEST_VERSION_RE = re.compile(r"(?<![A-Za-z])version\s*=\s*['\"]([^'\"]+)['\"]")
```

The negative lookbehind excludes a preceding **letter**, but the token it needs to exclude is `custom_config_version` — the character immediately before `version` there is `_`, which is not in `[A-Za-z]`. The lookbehind therefore does not fire.

In `nf-core/rnaseq` 3.26.0's `nextflow.config`:

| Line | Content |
|---|---|
| 150 | `custom_config_version = 'master'` |
| 464 | `version = '3.26.0'`  (inside the `manifest { }` block, lines 386-466) |

`re.search` returns the first match, so it reads `master` from line 150 and never reaches the manifest block.

```python
>>> pat = re.compile(r"(?<![A-Za-z])version\s*=\s*['\"]([^'\"]+)['\"]")
>>> pat.findall(open('nextflow.config').read())[:2]
['master', '3.26.0']
```

[...]

PR #335, opened 2026-08-11
fix(rnaseq-wrapper): read manifest.version, not custom_config_version

danilomonge:
## Summary

- The rnaseq wrapper read `custom_config_version = 'master'` instead of `manifest.version`, so **every** local `nf-core/rnaseq` checkout was detected as version `master`.
- The wrapper then rejected that checkout against its pinned `3.26.0` contract with `PIPELINE_SOURCE_INVALID`, which made the sibling-checkout path unusable unless `--allow-pipeline-version-override` was passed.
- Fixed by excluding `_` in the lookbehind, plus a regression test that reproduces the real nf-core config layout.

## Skill affected

`nfcore-rnaseq-wrapper` — `skills/nfcore-rnaseq-wrapper/pipeline_source.py`. It is the only wrapper that defines this regex, so the scrnaseq and sarek wrappers are unaffected.

## Root cause

`_MANIFEST_VERSION_RE` was:

```python
re.compile(r"(?<![A-Za-z])version\s*=\s*['\"]([^'\"]+)['\"]")
```

The lookbehind was added to stop `nextflowVersion` from matching, but `_` is not a letter, so any `*_version` parameter satisfies it too. nf-core pipelines declare those parameters in the `params` block, which sits **above** `manifest`, and `re.search` returns the first hit:

```groovy
params {
    version               = false      // unquoted, no match
    custom_config_version = 'master'   // first quoted match -> returned
}

manifest {
    version = '3.26.0'                 // the intended value
}
```

## Impact

`_read_manifest_version()` returned `master` for every real checkout. Since the wrapper enforces its pinned contract against that value, `resolve_pipeline_source()` failed with:

> Local nf-core/rnaseq checkout is version 'master', but this wrapper's validations are pinned to 3.26.0.

So `clawbio run rnaseq-pipeline` could not use a sibling `rnaseq/` checkout at all — not even a clean one at exactly the pinned tag. The documented `local_checkout_then_remote` source preference silently became "remote only", or failed outright with `--require-local-pipeline`.

## Verification

The new pattern was checked against the released `nextflow.config` of six pinned nf-core versions:

| pipeline | tag | before | after |
|---|---|---|---|
| rnaseq | 3.26.0 | `master` | `3.26.0` |
| rnaseq | 3.25.0 | `master` | `3.25.0` |
| scrnaseq | 4.1.0 | `master` | `4.1.0` |
| scrnaseq | 4.2.0 | `master` | `4.2.0` |
| sarek | 3.8.1 | `115.0-0` | `3.8.1` |
| sarek | 3.9.0 | `115.2-1` | `3.9.0` |

The two sarek values come from `vep_version`, another `*_version` parameter declared above the manifest — the same failure mode, a different key.

End to end, with a clean `nf-core/rnaseq` 3.26.0 checkout placed next to the repo so the `local_checkout_then_remote` path resolves to it: `clawbio run rnaseq-pipeline --check --input <samplesheet> --output <dir> --profile docker --aligner star_salmon --genome GRCh38` exits **1** with `PIPELINE_SOURCE_INVALID` before this change and exits **0** after it. `--check --demo` behaves the same way.

## Checklist

- [ ] SKILL.md is present and complete — n/a: bug fix to an existing skill, no SKILL.md change needed
- [ ] Demo data included for reviewers to test — n/a: the regression test builds its own fixture in `tmp_path`
- [x] Follows [CONTRIBUTING.md](https://github.com/ClawBio/ClawBio/blob/main/CONTRIBUTING.md) conventions

## Test output

Full wrapper suite with the fix applied:

```
$ python -m pytest skills/nfcore-rnaseq-wrapper/tests/ -q
507 passed, 144 warnings in 1.69s
```

The new test is a real regression test — it fails on `main` without the one-line change:

```
$ python -m pytest skills/nfcore-rnaseq-wrapper/tests/test_pipeline_source.py -k custom_config -q
E       AssertionError: assert 'master' == '3.26.0'
1 failed
```

The existing tests stayed green because their synthetic config contains only a `manifest` block, which is why this went unnoticed.

### DIFF
--- a/skills/nfcore-rnaseq-wrapper/pipeline_source.py
+++ b/skills/nfcore-rnaseq-wrapper/pipeline_source.py
@@ -21,9 +21,12 @@ from schemas import DEFAULT_LOCAL_PIPELINE_DIR, DEFAULT_REMOTE_PIPELINE, PIPELIN
 _GIT_TIMEOUT = 10
 
 # Matches `version = '3.26.0'` inside the manifest block of nextflow.config.
-# nextflow.config also carries `nextflowVersion`; the trailing-boundary group
-# (`\b`) prevents `nextflowVersion` from matching the bare `version` key.
-_MANIFEST_VERSION_RE = re.compile(r"(?<![A-Za-z])version\s*=\s*['\"]([^'\"]+)['\"]")
+# The lookbehind must reject every identifier that merely ends in `version`.
+# Excluding letters alone is not enough: nf-core configs declare
+# `custom_config_version = 'master'` in the params block, which sits above the
+# manifest, so that key matched first and every local checkout was reported as
+# version `master`. Excluding `_` as well also keeps `nextflowVersion` out.
+_MANIFEST_VERSION_RE = re.compile(r"(?<![A-Za-z_])version\s*=\s*['\"]([^'\"]+)['\"]")
 
 
 def _path_has_whitespace(path: Path) -> bool:
--- a/skills/nfcore-rnaseq-wrapper/tests/test_pipeline_source.py
+++ b/skills/nfcore-rnaseq-wrapper/tests/test_pipeline_source.py
@@ -77,6 +77,35 @@ def test_local_checkout_manifest_version_ignores_nextflow_version(tmp_path):
     assert result["manifest_version"] == "3.20.0"
 
 
+def test_local_checkout_manifest_version_ignores_custom_config_version(tmp_path):
+    """A real nf-core config declares `custom_config_version` above the manifest block.
+
+    The synthetic fixture used by the tests above does not, which is why the previous
+    regex could read `master` from every real checkout while these tests stayed green.
+    """
+    local = tmp_path / "rnaseq"
+    local.mkdir(parents=True)
+    (local / "main.nf").write_text("// main", encoding="utf-8")
+    (local / "nextflow.config").write_text(
+        "params {\n"
+        "    version                    = false\n"
+        "    custom_config_version      = 'master'\n"
+        "    custom_config_base         = \"https://example.invalid/${params.custom_config_version}\"\n"
+        "}\n"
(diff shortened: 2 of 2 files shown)


manuelcorpas:
Thanks for this, @danilomonge - the diagnosis was correct and the fix genuinely works.

I tested both this PR and #334 empirically against 12 real nf-core `nextflow.config` files (rnaseq, sarek, atacseq, scrnaseq, methylseq, differentialabundance, ampliseq, chipseq, eager, fetchngs, mag, viralrecon). Your regex returns the correct manifest version on **all 12**, including the sarek case where `main` returns `116.0-0` from `vep_version` - a plausible-looking wrong version, which is the nastiest kind because it misdirects debugging rather than looking absurd.

I have merged #334 instead, and I want to be clear that this is not a quality judgement on your patch. #334 scopes the parse to a brace-balanced `manifest { }` block rather than excluding prefixes, which additionally handles two cases where a lookbehind still returns the wrong value:

- `params.version = "9.9.9"` declared before the manifest (`.` is not in the excluded character class)
- a commented-out `// version = "0.0.1-dev"`

The two PRs conflict in both `pipeline_source.py` and the test file, so only one could land.

One thing from your PR worth calling out as the more valuable half: your new test is genuinely load-bearing. I mutation-tested it by reverting the regex to `main` and it goes red, and your fixture is properly nf-core-shaped with `custom_config_version` in a params block above the manifest. That is exactly the fixture discipline whose absence let #333 ship in the first place - the pre-existing shared fixture was an idealised config that could not exist in the wild. #334 carries that lesson forward by putting the adversarial keys into the shared fixture used by every test in the module.

Closing as superseded, with thanks. Please do keep sending patches.

Automated audit by ClawBio pr-audit plus manual verification. Manual review is still recommended.

PR #344, opened 2026-08-14
fix: handle list-valued ClinVar significance

### LINKED ISSUE #328: [Bug] clinical-variant-reporter: live ACMG classification crashes on list-valued clin_sig from VEP
## Skill affected

clinical-variant-reporter

## Describe the bug

In live mode (`clinical_variant_reporter.py --input <vcf>`, without `--demo`), ACMG classification aborts with `AttributeError: 'list' object has no attribute 'lower'` as soon as a variant carries ClinVar annotation.

Ensembl VEP returns `colocated_variants[].clin_sig` as a JSON array. `_extract_evidence_from_vep` stores that array unchanged:

https://github.com/ClawBio/ClawBio/blob/97f178e4790fa2a927222c0a4c64147a274f5ae0/skills/clinical-variant-reporter/clinical_variant_reporter.py#L209-L218

The value is then passed as `clinvar_significance=clinvar_sig` into `VariantEvidence`, where the field is declared `clinvar_significance: str = ""`. `_eval_ps1` calls `.lower()` on it on its first line, before any review-star or consequence gating, so the crash is unconditional for any ClinVar-annotated variant:

https://github.com/ClawBio/ClawBio/blob/97f178e4790fa2a927222c0a4c64147a274f5ae0/skills/clinical-variant-reporter/acmg_engine.py#L232-L240

`_eval_pp5` and `_eval_bp6` dereference the same field the same way.

PR #322 added an `isinstance` guard for `clin_sig_allele` in this same block, which fixed the earlier `'str' object has no attribute 'get'` crash. `clin_sig` on the line above was left un-normalized, so the live path that PR set out to unblock still fails on the identical real payload.

## Steps to reproduce

1. Take any VCF containing a variant with ClinVar records, for example BRCA2 `13:32339267 A>T` or `17:43093464 A>G`.
2. Run `python clinical_variant_reporter.py --input variants.vcf --output out/` (live mode, no `--demo`).
3. The Ensembl response for that region contains `colocated_variants[0].clin_sig` as a list, for example `["uncertain_significance", "pathogenic"]`.

A direct check of the upstream shape, using the same region string the skill builds:

```
curl -s 'https://rest.ensembl.org/vep/human/region/17:43093464-43093464:1/G?content-type=application/json' | jq '.[0].colocated_variants[0].clin_sig'
```

## Expected behaviour

Classification completes and `generate_report` writes `report.md` and the result files, with the ClinVar significance rendered as text.

## Actual behaviour

The run dies before any output is written:

```
Traceback (most recent call last):
  File "clinical_variant_reporter.py", line 352, in run_classification
    result = classify_variant(ev)
  File "acmg_engine.py", line 468, in classify_variant
    ...
  File "acmg_engine.py", line 389, in evaluate_criteria
    ...
  File "acmg_engine.py", line 234, in _eval_ps1
    ...
AttributeError: 'list' object has no attribute 'lower'
```

Because `run_classification` has no handler around `classify_variant` and `main` does not catch it, no `report.md`, no result JSON and no ACMG table are produced, including for variants that had already annotated cleanly. Demo mode is unaffected because `demo_evidence_cache.json` stores `clinvar_significance` as a plain string, which is why the test suite stays green.

## Environment

- OS: Linux
- Python version: 3.11
- ClawBio version/commit: `97f178e` (default branch `main`)

## Input data

Any VCF with at least one ClinVar-known variant. No patient data required, a single line with a known pathogenic position is enough.

## Suggested fix

Normalize the value where it is extracted, so the `str` contract of `VariantEvidence.clinvar_significance` holds:

```python
if isinstance(sigs, list):
    clinvar_sig = ", ".join(str(s) for s in sigs)
else:
    clinvar_sig = sigs or ""
```

Coercing at the `VariantEvidence` boundary instead would also cover any other producer. Worth noting that the test added in PR #322 already uses `"clin_sig": ["pathogenic"]` in its fixture, so extending that test to call `classify_variant` on the resulting evidence would catch this.

_Automated report: this issue was produced and filed automatically, with no human review before posting. Two independent checks agreed it is a real bug, but if it misreads the code please say so and we will close it._

_Found while running [Ito](https://ito.ai) (AI code review that runs your application, free for open source) against recently merged PRs. [Full analysis](https://app.ito.ai/share/298cb64f-d9d5-4e0f-ae69-4e46fc35411b?tab=details&utm_source=oss-outreach&utm_medium=github-issue&utm_campaign=oss-autopilot)._

Sanjays2402:
Closes #328

## Summary

- Normalize ClinVar significance values received as strings or lists into exact terms.
- Treat conflicting pathogenic and benign assertions as neither direction, avoiding both crashes and silent misclassification.

## Skill affected

clinical-variant-reporter

## Checklist

- [ ] SKILL.md is present and complete (YAML frontmatter + methodology)
- [ ] Demo data included for reviewers to test (not needed for this bug fix)

## Test output

`pytest skills/clinical-variant-reporter/tests/test_clinical_variant_reporter.py -q` — 68 passed.

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

### DIFF
--- a/skills/clinical-variant-reporter/acmg_engine.py
+++ b/skills/clinical-variant-reporter/acmg_engine.py
@@ -118,7 +118,7 @@ class VariantEvidence:
     hgvsp: str = ""
     transcript: str = ""
 
-    clinvar_significance: str = ""
+    clinvar_significance: str | list[str] = ""
     clinvar_review_stars: int = 0
 
     gnomad_af: float | None = None
@@ -162,6 +162,23 @@ class ClassifiedVariant:
 # ---------------------------------------------------------------------------
 # Criteria evaluation
 # ---------------------------------------------------------------------------
+def _clinvar_terms(significance: str | list[str]) -> set[str]:
+    """Normalize ClinVar's string and list forms into comparable terms."""
+    values = significance if isinstance(significance, list) else significance.split(",")
+    return {value.strip().lower().replace(" ", "_") for value in values if value.strip()}
+
+
+def _clinvar_directions(significance: str | list[str]) -> tuple[bool, bool]:
+    """Return pathogenic and benign flags, rejecting conflicting assertions."""
+    terms = _clinvar_terms(significance)
+    pathogenic = {"pathogenic", "likely_pathogenic"}
+    benign = {"benign", "likely_benign"}
+    conflicting = "conflicting_interpretations" in terms or (
+        bool(terms & pathogenic) and bool(terms & benign)
+    )
+    return bool(terms & pathogenic) and not conflicting, bool(terms & benign) and not conflicting
+
+
 def _eval_ba1(ev: VariantEvidence) -> EvidenceCriterion:
     """BA1: gnomAD AF > 5% → stand-alone benign."""
     triggered = ev.gnomad_af is not None and ev.gnomad_af > THRESHOLD_BA1_AF
@@ -231,8 +248,7 @@ def _eval_pvs1(ev: VariantEvidence) -> EvidenceCriterion:
 
 def _eval_ps1(ev: VariantEvidence) -> EvidenceCriterion:
     """PS1: Same amino acid change as an established pathogenic variant in ClinVar."""
-    clinvar_lower = ev.clinvar_significance.lower()
-    is_pathogenic_clinvar = "pathogenic" in clinvar_lower and "conflicting" not in clinvar_lower
+    is_pathogenic_clinvar, _ = _clinvar_directions(ev.clinvar_significance)
     triggered = (
         ev.is_missense
         and is_pathogenic_clinvar
@@ -306,8 +322,7 @@ def _eval_pp3(ev: VariantEvidence) -> EvidenceCriterion:
 
 def _eval_pp5(ev: VariantEvidence) -> EvidenceCriterion:
     """PP5: Reputable source reports variant as pathogenic."""
-    clinvar_lower = ev.clinvar_significance.lower()
-    is_pathogenic = "pathogenic" in clinvar_lower and "conflicting" not in clinvar_lower
+    is_pathogenic, _ = _clinvar_directions(ev.clinvar_significance)
     triggered = is_pathogenic and ev.clinvar_review_stars >= CLINVAR_MIN_STARS
     return EvidenceCriterion(
         code="PP5",
@@ -349,8 +364,7 @@ def _eval_bp4(ev: VariantEvidence) -> EvidenceCriterion:
 
 def _eval_bp6(ev: VariantEvidence) -> EvidenceCriterion:
     """BP6: Reputable source reports variant as benign."""
-    clinvar_lower = ev.clinvar_significance.lower()
-    is_benign = ("benign" in clinvar_lower or "likely benign" in clinvar_lower) and "pathogenic" not in clinvar_lower
+    _, is_benign = _clinvar_directions(ev.clinvar_significance)
     triggered = is_benign and ev.clinvar_review_stars >= CLINVAR_MIN_STARS
     return EvidenceCriterion(
         code="BP6",
--- a/skills/clinical-variant-reporter/tests/test_clinical_variant_reporter.py
+++ b/skills/clinical-variant-reporter/tests/test_clinical_variant_reporter.py
@@ -229,6 +229,29 @@ class TestCriteriaEvaluation:
         bp6 = next(c for c in criteria if c.code == "BP6")
         assert bp6.triggered is True
 
+    @pytest.mark.parametrize("significance", [
+        "Pathogenic", ["pathogenic", "pathogenic"],
+    ])
+    def test_clinvar_pathogenic_terms_trigger_pp5(self, significance):
+        ev = VariantEvidence(
+            chrom="chr1", pos=100, ref="A", alt="G",
+            clinvar_significance=significance, clinvar_review_stars=3,
+        )
+        pp5 = next(c for c in evaluate_criteria(ev) if c.code == "PP5")
+        assert pp5.triggered is True
+
+    @pytest.mark.parametrize("significance", [
+        ["benign", "pathogenic"], ["conflicting_interpretations"], ["novel_value"],
+    ])
+    def test_clinvar_conflicting_or_unknown_terms_do_not_trigger(self, significance):
+        ev = VariantEvidence(
+            chrom="chr1", pos=100, ref="A", alt="G",
+            clinvar_significance=significance, clinvar_review_stars=3,
+        )
+        criteria = evaluate_criteria(ev)
+        assert next(c for c in criteria if c.code == "PP5").triggered is False
+        assert next(c for c in criteria if c.code == "BP6").triggered is False
+
 
 # ---------------------------------------------------------------------------
 # Unit tests — SF v3.2 screening
@@ -513,6 +536,27 @@ class TestClinSigAlleleTypeGuard:
         ev = _extract_evidence_from_vep(
             self._vep({"review_status_stars": 3}), self._record())
         assert ev.clinvar_review_stars == 3
+
+    def test_list_clin_sig_classifies_conflicts_without_crashing(self):
+        from clinical_variant_reporter import _extract_evidence_from_vep
+        ev = _extract_evidence_from_vep(
+            {
+                **self._vep({"review_status_stars": 3}),
+                "most_severe_consequence": "missense_variant",
+                "transcript_consequences": [{
+                    "gene_symbol": "BRCA2",
+                    "impact": "MODERATE",
+                    "consequence_terms": ["missense_variant"],
+                }],
+                "colocated_variants": [{
+                    "clin_sig": ["benign", "pathogenic"],
+                    "clin_sig_allele": {"review_status_stars": 3},
+                }],
+            },
+            self._record(),
+        )
+        classified = classify_variant(ev)
+        assert all(code not in classified.triggered_codes for code in ("PS1", "PP5", "BP6"))
 
 
 # ---------------------------------------------------------------------------


manuelcorpas (changes requested):
## Audit: changes requested

Thanks for this. The crash you are fixing is real: VEP returns `clin_sig` as a list and the old code called `.lower()` on it, so this needed doing.

The refactor swaps substring matching for exact set membership, and that changes behaviour on four inputs that worked before. I verified each by executing `acmg_engine.py` at head `7eb0c47`.

### Blocking: the conflicting-interpretations guard no longer fires

`_clinvar_directions` tests `"conflicting_interpretations" in terms`, which is exact membership in a set. The term VEP emits is `conflicting_interpretations_of_pathogenicity`, and since ClinVar's 2024 rename, `conflicting_classifications_of_pathogenicity`. Neither matches, so the guard is bypassed:

```
input=['pathogenic', 'conflicting_classifications_of_pathogenicity']
PS1=True  PP5=True  ->  classification = Likely Pathogenic
```

A variant ClinVar reports as having conflicting classifications comes back as Likely Pathogenic, with both PS1 (strong) and PP5 (supporting) firing. The previous `"conflicting" not in clinvar_lower` caught this correctly, so this is a guard that regressed rather than one that was never there.

The value reaches the function unnormalised: `clinical_variant_reporter.py:227` assigns `clin_sig` straight through, and line 273 passes it to `clinvar_significance`.

### Also blocking: ClinVar's combined values stop matching

`Pathogenic/Likely pathogenic` and `Benign/Likely benign` are common ClinVar clinical significance values. Splitting only on `,` leaves them as single unmatched terms.

| input (string form, worked before) | before | after |
|---|---|---|
| `pathogenic, conflicting_interpretations_of_pathogenicity` | blocked | triggers |
| `pathogenic, conflicting_classifications_of_pathogenicity` | blocked | triggers |
| `Pathogenic/Likely pathogenic` | triggers | no match |
| `Benign/Likely benign` | triggers | no match |

### Why the tests did not catch it

The new parametrised test at `tests/test_clinical_variant_reporter.py:244` uses `["conflicting_interpretations"]`. That short form is not a value VEP produces. Substituting either real term turns the test red.

### Suggested fix

Both small:

- match `conflicting` as a prefix or substring over each term rather than by exact equality;
- split on `/` as well as `,` before normalising.

Then please re-parametrise the test on `conflicting_interpretations_of_pathogenicity`, `conflicting_classifications_of_pathogenicity` and `Pathogenic/Likely pathogenic`, so the values under test are ones VEP actually emits.

Everything else here is an improvement: the `str | list[str]` annotation, the extraction-layer test, and routing PS1, PP5 and BP6 through a single helper.

Worth noting that all eight CI checks pass on this PR, and our static audit scanner returned no findings. The gap is in the fixtures, not in the pipeline.

manuelcorpas (changes requested):
The bug you found is real and severe, and worth saying plainly: `_extract_evidence_from_vep` assigns VEP's `clin_sig` straight through, VEP REST returns that field as a **list**, and every `_eval_*` on `main` starts `ev.clinvar_significance.lower()`. So live mode was raising `AttributeError: 'list' object has no attribute 'lower'` on essentially every ClinVar-annotated variant. The test helper at `tests/test_clinical_variant_reporter.py:524` already passed `["pathogenic"]` but never called `evaluate_criteria`, so a green suite was hiding a completely broken live path. Thank you for catching it.

The abstain-on-conflict design is also the right call. Of the four ways to reduce a list to a decision, zeroing both flags on co-occurrence and letting the variant fall to VUS is the correct one.

Requesting changes because the normaliser swapped **substring** matching for **exact set membership**, and that regresses the scalar path that previously worked. Two defects, pointing in opposite directions, so no single end-to-end count would surface either.

**1. Slash compounds now match nothing.** `acmg_engine.py:165-168` splits on `,` but never `/`:

```python
values = significance if isinstance(significance, list) else significance.split(",")
return {value.strip().lower().replace(" ", "_") for value in values if value.strip()}
```

`Pathogenic/Likely_pathogenic` is a **single** ClinVar term, not two. It normalises to the one token `pathogenic/likely_pathogenic`, which is in neither `pathogenic` nor `benign` at `:174-175`, so PS1, PP5 and BP6 all silently drop.

Concrete: BRCA1 missense, ClinVar `Pathogenic/Likely_pathogenic`, 3 review stars, absent from gnomAD, CADD 28, SIFT deleterious, PolyPhen probably_damaging.
- `main`: PS1 + PP5 + PM2 + PP3, `ps>=1 and pm>=1`, **Likely Pathogenic**.
- this PR: PS1 and PP5 drop, `pm=1, pp=1`, no rule matches, **Uncertain Significance**.

The mirror case holds too: `Benign/Likely_benign` at 2 stars goes Likely Benign to VUS. `main`'s `"pathogenic" in clinvar_lower` caught both. The repo already knows these are single terms: `skills/variant-annotation/variant_annotation.py:57` enumerates `"pathogenic/likely_pathogenic"` and `vcf-annotator` emits `"Benign/Likely benign"`. No test in the suite uses a slash term, so nothing goes red.

**2. The conflicting-term literal matches nothing real.** `acmg_engine.py:176`:

```python
conflicting = "conflicting_interpretations" in terms or (
    bool(terms & pathogenic) and bool(terms & benign)
)
```

`in` on a set is exact membership. The real values are `conflicting_interpretations_of_pathogenicity` (VEP <=111) and, since ClinVar's 2024 data-model rename, `conflicting_classifications_of_pathogenicity`. Neither equals the literal, so that clause is dead code and `main`'s `"conflicting" not in clinvar_lower` was catching both.

Concrete: `clin_sig = ["conflicting_classifications_of_pathogenicity", "pathogenic"]`, 2 stars, missense, absent from gnomAD, CADD 28. `terms & benign` is empty and the literal does not match, so `conflicting` is False, PS1 and PP5 both fire, `ps=1, pm=1, pp=2`, **Likely Pathogenic** on a variant ClinVar itself refuses to classify. On `main` the scalar equivalent gave VUS.

**3. The test named for that guard passes with the guard deleted.** `tests/test_clinical_variant_reporter.py:243-253`: for `["conflicting_interpretations"]` and `["novel_value"]` both intersections are already empty, so PP5 and BP6 are False regardless of `conflicting`. Only `["benign", "pathogenic"]` is load-bearing, and it exercises the co-occurrence clause. The fixture also invents a term that exists nowhere in ClinVar, which is precisely why the real term's mismatch went unnoticed.

Requested changes, all small:

1. Add the compound terms to both sets, ideally reusing `variant-annotation`'s `CLINVAR_PATHOGENIC_TERMS` as a shared vocabulary. Split on `,`, `|` and `&`, never on `/`.
2. Replace the literal with a prefix test, `any(t.startswith("conflicting") for t in terms)`, so both the pre-2024 and post-2024 names match.
3. Rewrite the conflicting test with real terms, and confirm each new parameter goes red when the guard is deleted. Add parameters for `pathogenic/likely_pathogenic`, `benign/likely_benign`, `Likely pathogenic`, and one asserting PP5 **does** fire for `["pathogenic", "likely_pathogenic"]`.
4. Add a scalar regression test pinning `Pathogenic/Likely_pathogenic` at 3 stars to Likely Pathogenic, so the downgrade cannot recur.
5. Join the value for display at `clinical_variant_reporter.py:609` and `:721`. Now that lists survive classification, the report renders `- **ClinVar**: ['benign', 'pathogenic'] (stars: 3)`. Worth stating the abstention explicitly rather than leaving a clinician to infer why PP5 is absent.

Worth noting for a follow-up but **not** yours to fix here: `_clinvar_terms(None)` still reaches `None.split(",")`, reachable from a cache JSON with an explicit null; the conflict rule is applied to `clin_sig`, which is rsID-level rather than allele-level, so a multi-allelic dbSNP entry suppresses PS1/PP5 on a genuinely pathogenic allele (fails safe, but it is a missed call and your new test at `:552` locks it in); and PS1 and PP5 fire on nearly the same predicate, so one ClinVar record scores both strong and supporting with no shared-evidence cap. I will open separate issues for those.

Verified and unchanged by this PR, for the record: the combining rules at `:417-474` match Richards 2015 Table 5 exactly, all eight Pathogenic and all six Likely Pathogenic combinations, with no off-by-one at the boundary; and `CLINVAR_MIN_STARS = 2` correctly stops a 0-star single-submitter assertion carrying weight.

manuelcorpas (changes requested):
## pr-audit: changes requested

Thanks for this. The intent is right and so is the design: suppressing both directions when ClinVar disagrees with itself is the correct clinical policy, and it is a real improvement on the old substring test. Two defects in the *detection* make it unsafe to merge as it stands, and both are invisible to the current suite.

### 1. CRITICAL, `skills/clinical-variant-reporter/acmg_engine.py:176`

The conflicting sentinel is a token that neither ClinVar nor VEP ever emits.

```python
conflicting = "conflicting_interpretations" in terms or (
```

`terms` is a set, so this is an exact-equality test. ClinVar VCF `CLNSIG` emits `Conflicting_interpretations_of_pathogenicity`; VEP REST emits `conflicting_interpretations_of_pathogenicity`. Neither equals `conflicting_interpretations`. A record that ClinVar explicitly flags as conflicting, but whose submitter set includes a Pathogenic assertion, therefore passes the guard and fires PS1 (strong) plus PP5. Through `classify()` at lines 459 to 460, that promotes a VUS to **Likely Pathogenic**.

### 2. CRITICAL, `acmg_engine.py:167`

Splitting only on `,` means the standard ClinVar aggregate forms no longer resolve at all.

```python
values = significance if isinstance(significance, list) else significance.split(",")
```

`/` and `|` are both live ClinVar multi-value delimiters, and neither is split, so exact set membership can never match. Measured against the pre-PR behaviour:

| Value | pre-PR | at head | effect |
|---|---|---|---|
| `Conflicting interpretations of pathogenicity, Pathogenic` | `(False, False)` | `(True, False)` | over-call, fires PS1 + PP5 |
| `Pathogenic/Likely pathogenic` | `(True, False)` | `(False, False)` | P/LP downgraded to VUS |
| `Pathogenic/Likely_pathogenic` | `(True, False)` | `(False, False)` | P/LP downgraded to VUS |
| `Pathogenic\|drug_response` | `(True, False)` | `(False, False)` | P/LP downgraded to VUS |
| `Benign/Likely benign` | `(False, True)` | `(False, False)` | loses BP6 |

So the change over-calls on the conflicting case and under-calls on the compound case. Both directions are clinically wrong, and the second is a regression on input the old substring test handled correctly.

### 3. WARNING, `tests/test_clinical_variant_reporter.py:244`

```python
["benign", "pathogenic"], ["conflicting_interpretations"], ["novel_value"],
```

The conflicting case is parametrised on a token that exists in neither vocabulary, so the test vocabulary was written against the implementation rather than against the data source. That is why neither CRITICAL surfaced. Note also that deleting the literal-membership half of the guard leaves every test green: only the both-directions half is load-bearing, so the branch this PR exists to add is half untested, and the untested half is the broken one.

### Suggested fix

- Match conflicting by prefix rather than equality, for example `any(t.startswith("conflicting") for t in terms)`.
- Split the string path on `[,/|]`, and each list element on `[/|]`.
- Re-parametrise with real vocabulary: `conflicting_interpretations_of_pathogenicity`, plus `Pathogenic/Likely_pathogenic` and `Benign/Likely_benign` as happy cases.
- Confirm the guard is load-bearing by deleting it and watching a test go red.

### Minor

`clinical_variant_reporter.py:721` now writes `ev.clinvar_significance` unconverted into `acmg_classifications.tsv`, so that column will carry a Python list repr such as `['benign', 'pathogenic']` in a clinical deliverable. The same value is interpolated at report line 609 and at engine lines 262, 332 and 374.

One note on scope: the repo's own `demo_evidence_cache.json` contains no compound and no conflicting value, so running the skill's demo cannot surface either CRITICAL. Worth adding both to the cache.

manuelcorpas (changes requested):
### ClawBio automated audit

**Verdict: REJECT**  

Thanks for this. The bug you are fixing is real: `clinical_variant_reporter.py:211-213` assigns VEP's `clin_sig` straight through, and VEP returns that field as a JSON list, so `ev.clinvar_significance.lower()` on main (`acmg_engine.py:234`, `:309`, `:352`) raises `AttributeError` on ordinary annotation output. Widening the field type at `acmg_engine.py:121` is the right call, collapsing three copies of the same ad hoc check into one helper is an improvement, and treating a simultaneous pathogenic and benign assertion as non actionable fixes a real asymmetry in the old BP6 (`acmg_engine.py:353`), which suppressed BP6 on any string containing "pathogenic" but had no matching guard on PS1 and PP5.

I am rejecting on the vocabulary used for the conflicting check.

**1. `conflicting_interpretations` is not a term ClinVar or VEP emits (`acmg_engine.py:176`).**

The guard is an exact set membership test against the normalised token `conflicting_interpretations`. The terms actually produced are:

- VEP JSON `clin_sig`, from the Ensembl clinical significance enum: `conflicting_interpretations_of_pathogenicity`, and `conflicting_data_from_submitters`.
- VEP 111 and later, tracking the 2023 ClinVar rename: `conflicting_classifications_of_pathogenicity`.
- ClinVar `CLNSIG` directly: `Conflicting_classifications_of_pathogenicity`, previously `Conflicting_interpretations_of_pathogenicity`.

None of these normalise to `conflicting_interpretations`, so on real input that clause never evaluates true. The only surviving protection is the pathogenic-and-benign intersection on the same line. That leaves a concrete over-call: a variation whose aggregated `clin_sig` is `["pathogenic", "conflicting_interpretations_of_pathogenicity"]`, which Ensembl produces when a variation carries several ClinVar records with differing assertions, now returns `(True, False)` from `_clinvar_directions`. With stars at or above `CLINVAR_MIN_STARS` that fires PS1 (`acmg_engine.py:251`), strong pathogenic evidence, and PP5 (`acmg_engine.py:325`) on a variant ClinVar reports as conflicting. The same path fires BP6 (`acmg_engine.py:367`) for `["benign", "conflicting_..."]`. The substring test on main, `"conflicting" not in clinvar_lower`, caught exactly this. So for the string input path this is a behavioural regression, and for the list path it converts a loud crash into a silent wrong classification, which is the worse of the two failure modes in an ACMG engine.

Suggested fix: match the family rather than one literal, for example `any(t.startswith("conflicting") for t in terms)`, or an explicit set of the four real tokens above. Please keep both the `_interpretations_` and `_classifications_` spellings, since VEP cache vintages in the wild emit both.

**2. The tests confirm the implementation, not the data source (`tests/test_clinical_variant_reporter.py:244`).**

The conflicting case is parametrised as `["conflicting_interpretations"]`, which is the invented token, so it passes by construction. The `["benign", "pathogenic"]` case passes through the intersection clause instead, and `["novel_value"]` passes because any unrecognised token yields no trigger. Delete the `"conflicting_interpretations" in terms` clause entirely and only the invented-token case fails. No case built from a string VEP or ClinVar really returns would notice. The same applies to the integration test at `tests/test_clinical_variant_reporter.py:540`, which asserts the conflict is caught but supplies `clin_sig` as `["benign", "pathogenic"]`, again exercising the intersection rather than the conflicting term. Please add a case using `conflicting_interpretations_of_pathogenicity` with `review_status_stars` at 2 or above and assert PS1, PP5 and BP6 all stay untriggered.

**3. Splitting on the comma alone drops genuine pathogenic assertions (`acmg_engine.py:167`).**

The string branch splits on `,` only. ClinVar's combined aggregate labels use a solidus, so `"Pathogenic/Likely pathogenic"` normalises to the single token `pathogenic/likely_pathogenic` and matches neither set, and `"Benign/Likely benign"` behaves the same way. Those are among the commonest aggregate values in ClinVar, and the old substring code handled them. VEP's tabular and VCF `CLIN_SIG` output uses `&` as its delimiter, with `|` appearing in ClinVar VCF `CLNSIG`. The direction of this error is under-call rather than over-call, so it is the less dangerous of the two, but it is silent. Splitting on `[,&|/]` would cover all the observed forms. The cached values in `example_data/demo_evidence_cache.json` are all single simple terms, so the existing suite cannot detect this.

**4. Minor, from the widened type.** A list now reaches `clinical_variant_reporter.py:440` and `:552` unchanged, so the markdown report and the `clinvar_significance` column of `tables/acmg_classifications.tsv` will render a Python list repr such as `['benign', 'pathogenic']`. Worth joining on ", " at those two call sites.

On the static scan: it reported no findings on this PR, so there is nothing for me to record as a false positive. Nothing above was flagged by it. I also checked the diff for credential material, network calls and any movement of variant or phenotype data off the machine, and for instruction-like text addressed at a reviewer. There is none; the change is confined to local classification logic.

The structure of this refactor is right and points 1 and 3 are a small edit to `_clinvar_terms` and `_clinvar_directions`. Happy to look again once the conflicting terms match what the annotator emits and a test covers a real one.

No human has reviewed this pull request. This is an automated domain review.

---
Changes requested.

manuelcorpas:
Thank you for taking #328 on. Closing this one as superseded rather than asking for another round.

Main fixed the same bug on 2 September in 51a58538 ("parse ClinVar significance into terms"), refined in f7c8bac1. That version splits on `, ; / | &`, enumerates the three real conflicting tokens (`conflicting_interpretations_of_pathogenicity`, `conflicting_classifications_of_pathogenicity`, `conflicting_data_from_submitters`) and gates both `supports_pathogenic` and `supports_benign` on `is_conflicting`. Issue #328 is closed.

For the record, the two points that kept this PR open are still present at head 7eb0c47a: `acmg_engine.py:176` matches the literal `conflicting_interpretations`, which neither ClinVar nor VEP emits, and `:167` splits on commas only, so `Pathogenic/Likely_pathogenic` becomes one unrecognised token. The tests at `test_clinical_variant_reporter.py:244` use that invented token, which is why they stayed green.

If you want to contribute again, the wanted-skills list in CONTRIBUTING.md is the best place to start.

PR #345, opened 2026-08-15
Add skill: my-awesome-skill

dhjnam:
SKILL.md + demo data + runnable --demo command

### DIFF
--- a/skills/my-awesome-skill/SKILL.md
+++ b/skills/my-awesome-skill/SKILL.md
@@ -1,0 +1,265 @@
+---
+name: my-awesome-skill
+version: 0.1.0
+author: Your Name <you@example.com>
+domain: genomics           # or: pharmacogenomics, clinical, population-genetics, multi-omics
+description: One-line description of what this skill does
+license: MIT
+
+inputs:
+  - name: input_file
+    type: file
+    format: [vcf, csv, tsv, txt]
+    description: Input data file
+    required: true
+
+outputs:
+  - name: report
+    type: file
+    format: md
+    description: Analysis report
+
+dependencies:
+  python: ">=3.11"
+  packages:
+    - pandas>=2.0
+    - numpy>=1.24
+
+tags: [genomics, variants, your-tags-here]
+
+demo_data:
+  - path: demo_input.txt
+    description: Synthetic test data with 50 variants
+
+endpoints:
+  cli: python skills/my-awesome-skill/my_skill.py --input {input_file} --output {output_dir}
+---
+
+# 🦖 Skill Name
+
+You are **[Skill Name]**, a specialised ClawBio agent for [domain]. Your role is to [core function in one sentence].
+
+## Domain Decisions
+
+Document the scientific choices your skill makes. These are the decisions
+a domain expert would make that an AI agent should NOT override.
+
+- **Threshold for significance**: p < 0.05 after Bonferroni correction
+- **Reference database**: ClinVar (release 2025-12)
+- **Population frequency filter**: exclude variants with gnomAD AF > 0.01
+
+## Safety Rules
+
+- Never report a clinical diagnosis; always include the research-use disclaimer
+- Flag DPYD*2A carriers with an urgent warning (fluorouracil toxicity risk)
+- Do not extrapolate results across ancestries unless the model was trained on that population
+
+## Agent Boundary
+
+The agent (LLM) dispatches and explains. The skill (Python) executes.
+The agent must NOT override thresholds, invent gene-drug associations,
+or skip safety warnings defined in this SKILL.md.
+
+## Trigger
+
+> **This is the most important section.** It determines whether the agent discovers
+> and fires this skill. Be loud, explicit, and list every plausible phrasing.
+
+**Fire this skill when the user says any of:**
+- "exact phrase 1"
+- "exact phrase 2"
+- "keyword or synonym"
+- "another way a user might ask for this"
+
+**Do NOT fire when:**
+- [Describe similar-sounding requests that should route elsewhere]
+- [Edge case that belongs to a different skill]
+
+**Design notes:** The trigger must be loud, not subtle. Models skip subdued
+descriptions. Use exact phrases, domain-specific terms, and multiple synonyms.
+If two skills sound similar, the trigger is where you disambiguate.
+
+## Why This Exists
+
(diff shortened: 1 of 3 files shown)


manuelcorpas (changes requested):
## Audit: changes requested

Thanks for having a go, and welcome.

This is `templates/SKILL-TEMPLATE.md` copied across with the placeholders still in place, so I do not think it is ready to merge as a skill. Listing the specifics in case any of it is unintentional:

- `author: Your Name <you@example.com>`, `description: One-line description of what this skill does` and `tags: [genomics, variants, your-tags-here]` are still the template defaults.
- `skills/my-awesome-skill/tests/test_my_skill.py` is 0 bytes at head.
- There are no `metadata.openclaw.trigger_keywords`, so the orchestrator has no way to route to the skill.
- SKILL.md lines 127 and 131 document `python skills/my-awesome-skill/my-awesome-skill.py`. That file does not exist; the script is `my_skill.py`. The frontmatter on line 35 has the right name.
- `parse_input` uses `sep=" "` while the shipped `demo_input.txt` is space-aligned, so it misparses instead of failing. Running it against the skill's own demo data gives:

```
{'total_variants': 3, 'chromosomes': [nan], ...}
```

The chromosome values land in a column called `Unnamed: 3`, and the column actually named `chromosome` is empty. It reports success with the wrong answer, which is worth fixing in whatever this becomes, because that kind of failure does not announce itself.

If you are working towards a real skill, the checklist in `templates/SKILL-TEMPLATE.md` is the place to start, and picking one specific task with a real input file tends to work better than filling in the generic scaffold. Happy to look again whenever you want.

manuelcorpas (changes requested):
Thanks for having a go at this, and sorry to bounce it.

This PR looks like the scaffold rather than a finished skill. It is `templates/SKILL-TEMPLATE.md` with the name swapped in. The author is still `Your Name <you@example.com>`, the trigger phrases are `"exact phrase 1"` and `"exact phrase 2"`, the three gotchas are the bracketed examples, and `analyse()` in `my_skill.py:23-31` is the stub with `# Add your analysis here` still in it. `tests/test_my_skill.py` is 0 bytes (git blob `e69de29b`), which is why the `test-changed-skills` job is red: pytest collects nothing and exits 5.

My guess is `git add` ran straight after `cp templates/SKILL-TEMPLATE.md`, which is an easy mistake given the order in CONTRIBUTING.

Please do open another PR when you have a real skill. If you already have a task in mind, say what it is and I will point you at the closest existing skill to work from. Four things worth knowing before you start:

- **Frontmatter has moved on** since whichever copy of the template you used. Everything now nests under `metadata:`, and `metadata.openclaw.trigger_keywords` is what the orchestrator routes on. Without at least three keywords there, nothing can reach your skill. Yours is currently the only skill of 99 with no `metadata:` block at all, so the catalogue would ship it with `trigger_keywords: []`. Take a fresh copy from `templates/` on current `main`.
- **The output tree in Output Structure is a contract.** Only list files the script always writes. Yours documents `result.json`, `tables/results.csv` and a reproducibility bundle that the code never produces; it writes `report.md` and `figures/summary.png`, and the documented figure name (`plot.png`) does not match either.
- **Your demo would not have worked.** `demo_input.txt` is space-aligned but `parse_input` reads it with `sep=" "` (`my_skill.py:14`), which shifts the columns so `chromosome` comes through as all-NaN. The required-columns check passes anyway, because the header names survive the shift even though the data does not, so the report would print `Chromosomes covered: nan` rather than failing. The tab version is sitting commented out on the line below. Worth running `--demo` and reading the report before pushing.
- **Please delete the template's example Domain Decisions and Safety Rules** unless your skill genuinely implements them. As written, `SKILL.md:42-56` tells the agent that this skill pins ClinVar release 2025-12, applies a gnomAD AF > 0.01 filter and flags DPYD\*2A for fluorouracil toxicity. None of that is in the code, and we treat SKILL.md as executable methodology. In a clinical catalogue that is the line we most need to hold.

Also worth flagging: `numpy>=1.24` is declared and never imported, while `matplotlib` is imported at module scope and never declared, so anyone installing from your declared dependency set hits `ModuleNotFoundError: matplotlib` on every run.

Happy to review a draft early next time if that helps.

Two of these are our bugs, not yours, and I am fixing them separately: `generate_catalog.py` awards the `tested` maturity tier on a filename glob, so a zero-byte test file earns a public quality signal; and we have no lint rule rejecting template sentinel strings, which would have caught this in seconds with a clear message instead of costing you a review round.

manuelcorpas:
One concrete detail to add, since the red tick has two separate causes:

- `test (3.12)`: pytest collects nothing from the 0-byte `tests/test_my_skill.py` and exits 5.
- `skill-lint`: `SKILL.md:12` has `format: [vcf, csv, tsv, txt]`. `agentskills validate` rejects YAML flow sequences, so it wants a block sequence instead:

```yaml
      format:
        - vcf
        - csv
```

That line comes straight from the older template you copied, which is more evidence for taking a fresh copy from `templates/` on current `main`. The current template uses block syntax throughout and nests everything under `metadata:`.

manuelcorpas:
Thanks for giving ClawBio a go, and welcome.

It looks like this PR contains the worked example from `docs/tutorial-build-a-skill.ipynb` rather than a skill of your own, so I am going to close it, but please do not read that as discouragement. The scaffold is still a placeholder in every substantive section: `analyse()` still carries `# Add your analysis here`, the Trigger section still lists `"exact phrase 1"`, and Gotcha 1 still reads `The model tends to [bad assumption]`.

When you have a genomics task you would like to automate, copy that scaffold to your own skill name, fill in the Trigger and Gotchas sections and the `analyse()` function, and open a fresh PR. We would be glad to review it.

Two things that tripped you up were our fault, not yours, and we are fixing them:

- The tutorial's frontmatter uses inline flow style such as `format: [vcf, csv, tsv, txt]`, which our `skill-lint` validator rejects. That is why the check went red, and it would go red for anyone following the tutorial today.
- The tutorial also omits the `metadata.openclaw.trigger_keywords` block that the orchestrator needs in order to route to a skill at all.

For future reference, the other thing to remember is that adding a skill means regenerating `skills/catalog.json`, which is what the `test_checked_in_catalog_is_current` failure was about.

Thanks again for taking the time to work through the tutorial.

PR #346, opened 2026-08-18
Add claim-provenance-check skill

eladmint:
## What

A citation-binder + coverage-report skill: given a case file of retrieved evidence rows, per-class
source outcomes, and drafted claims, it binds every `[class:index]` citation tag in a claim to a
retrieved row and reports which evidence classes a subject actually reached vs `NO COVERAGE`.

No partial credit — a claim with an invented citation, a missing citation, or an 8+ word verbatim
run copied from a cited row is refused in full, not scored down.

Standard-library-only Python (`re`, `dataclasses`, `enum`, `typing`), zero network calls, zero
model calls.

## Why

An LLM-drafted claim carrying a plausible-looking citation like `[trials:2]` reads as evidence to
anyone who doesn't re-run the retrieval themselves. This is a mechanical, free, deterministic gate
that runs before any model judge and catches exactly that failure mode — an invented citation —
which arithmetic catches perfectly and a paid judge cannot be trusted to catch reliably.

## Provenance

Ported from a citation-binder/coverage-report engine built for the ClawBio + Nebius Berlin
hackathon (2026-08-18), itself adapted from Sippar's production `reportQuality.ts`
(sippar.network). The core logic (`provenance_core.py`) is unchanged in behaviour from the
hackathon source; this PR is the from-scratch ClawBio-shaped packaging of it (`SKILL.md`, CLI
wrapper, synthetic demo data, tests).

## Checklist (CONTRIBUTING.md / CLAUDE.md 17-check conformance)

- [x] `SKILL.md` follows `templates/SKILL-TEMPLATE.md`: Trigger (fire / do-not-fire), Scope,
      Workflow (numbered steps), Example Output (real rendered sample), Gotchas (5 entries),
      Safety (references the ClawBio disclaimer), Agent Boundary, plus Domain Decisions grounded
      in what the code actually hardcodes (the 8-word verbatim threshold, the 2-class SUPPORTED
      floor, no-partial-credit).
- [x] `uv run agentskills validate skills/claim-provenance-check/` → `Valid skill`
- [x] Synthetic demo data only (`demo_input.txt`) — no real patient/trial data, clearly labelled
      synthetic in the file itself.
- [x] `python skills/claim-provenance-check/claim_provenance_check.py --demo --output <dir>` runs
      clean and writes `report.md` (with the safety disclaimer) + `result.json`.
- [x] 21 tests in `tests/test_claim_provenance_check.py`, all passing — pins the refusal behaviour
      (invented citation, missing citation, verbatim copying, no-partial-credit, NO COVERAGE vs
      CONTESTED vs SUPPORTED) plus CLI/end-to-end checks. `pytest skills/claim-provenance-check/tests/ -v`
- [x] Registered as `provcheck` in `clawbio.py` / `skills/catalog.json` via
      `python scripts/generate_catalog.py` (clean diff, no unrelated churn); confirmed with
      `python clawbio.py list` and `python clawbio.py run provcheck --demo`.

## Test plan

```bash
uv run agentskills validate skills/claim-provenance-check/
uv run pytest skills/claim-provenance-check/tests/ -v
python skills/claim-provenance-check/claim_provenance_check.py --demo --output /tmp/demo
python clawbio.py run provcheck --demo
```

### DIFF
--- a/clawbio/cli.py
+++ b/clawbio/cli.py
@@ -1220,6 +1220,13 @@ SKILLS = {
         "no_input_required": True,
         "accepts_genotypes": False,
     },
+    "provcheck": {
+        "script": SKILLS_DIR / "claim-provenance-check" / "claim_provenance_check.py",
+        "demo_args": ["--demo"],
+        "description": "Bind every citation in a claim to a retrieved evidence row and report which evidence classes were reached, refusing invented citations with no partial credit.",
+        "allowed_extra_flags": set(),
+        "accepts_genotypes": False,
+    },
 }
 
 try:
--- a/skills/catalog.json
+++ b/skills/catalog.json
@@ -1,7 +1,7 @@
 {
   "version": "1.0.0",
   "generated_by": "scripts/generate_catalog.py",
-  "skill_count": 96,
+  "skill_count": 97,
   "galaxy_tool_count": 8182,
   "skills": [
     {
@@ -626,6 +626,54 @@
         "cell-type-specific expression",
         "expression specificity",
         "marker gene specificity"
+      ],
+      "chaining_partners": [],
+      "license": "MIT",
+      "model_license": "",
+      "data_license": ""
+    },
+    {
+      "name": "claim-provenance-check",
+      "cli_alias": "provcheck",
+      "description": "Bind every citation in a claim to a retrieved evidence row and report which evidence classes were reached, refusing invented citations with no partial credit",
+      "version": "0.1.0",
+      "status": "planned",
+      "maturity_tier": "cli-registered",
+      "maturity_evidence": {
+        "has_skill_md": true,
+        "has_script": true,
+        "has_tests": true,
+        "has_demo": true,
+        "cli_registered": true,
+        "ci_tested": false,
+        "benchmark_validated": false
+      },
+      "has_script": true,
+      "has_tests": true,
+      "has_demo": true,
+      "demo_command": "python clawbio.py run provcheck --demo",
+      "dependencies": [
+        "python>=3.9"
+      ],
+      "tags": [
+        "citation-verification",
+        "provenance",
+        "coverage-report",
+        "hallucination-detection",
+        "evidence-binding",
+        "report-qa"
+      ],
+      "trigger_keywords": [
+        "check citations",
+        "citation binder",
+        "verify claim citations",
+        "coverage report",
+        "no coverage",
+        "invented citation",
+        "hallucinated citation",
+        "does this claim have evidence",
+        "bind claim to evidence",
+        "unbindable claim"
       ],
       "chaining_partners": [],
       "license": "MIT",
--- a/skills/claim-provenance-check/SKILL.md
+++ b/skills/claim-provenance-check/SKILL.md
@@ -1,0 +1,336 @@
+---
+name: claim-provenance-check
+description: Bind every citation in a claim to a retrieved evidence row and report which evidence classes were reached, refusing invented citations with no partial credit
+license: MIT
+metadata:
+  version: 0.1.0
+  author: Sippar (Nuru-AI) <elad@sippar.network>
+  domain: clinical
+  tags:
+    - citation-verification
+    - provenance
+    - coverage-report
+    - hallucination-detection
+    - evidence-binding
+    - report-qa
+  inputs:
+    - name: input_file
+      type: file
+      format:
+        - json
+      description: >-
+        Case file (conventionally named demo_input.txt or *.json): a JSON object with
+        "subject" (string), "rows" (retrieved EvidenceRow objects: tag, source_class, text,
+        url?, cohort?, n?), "outcomes" (SourceOutcome objects: source_class, ok, rows?,
+        reason?, route?, cost_usd?), and "claims" (strings containing "[class:index]"
+        citation tags).
+      required: false
+    - name: demo
+      type: flag
+      description: Run on the bundled synthetic demo_input.txt instead of --input
+      required: false
+  outputs:
+    - name: report
+      type: file
+      format:
+        - md
+      description: Coverage map + per-claim verdict (SUPPORTED / CONTESTED / NO COVERAGE) with refusal reasons and the safety disclaimer (report.md)
+    - name: result
+      type: file
+      format:
+        - json
+      description: Machine-readable coverage map and claim verdicts (result.json)
+  dependencies:
+    python: ">=3.9"
+  demo_data:
+    - path: demo_input.txt
+      description: Synthetic evidence case with one SUPPORTED claim, one CONTESTED claim, one uncited claim, and one claim with an invented citation tag -- exercises every verdict path
+  endpoints:
+    cli: python skills/claim-provenance-check/claim_provenance_check.py --input {input_file} --output {output_dir}
+    cli_demo: python skills/claim-provenance-check/claim_provenance_check.py --demo --output {output_dir}
+  openclaw:
+    requires:
+      bins:
+        - python3
+    always: false
+    emoji: "🔗"
+    homepage: https://github.com/ClawBio/ClawBio
+    os:
+      - darwin
+      - linux
+    trigger_keywords:
+      - check citations
+      - citation binder
+      - verify claim citations
+      - coverage report
+      - no coverage
+      - invented citation
+      - hallucinated citation
+      - does this claim have evidence
+      - bind claim to evidence
+      - unbindable claim
+---
+
+# 🔗 Claim Provenance Check
+
+You are **Claim Provenance Check**, a specialised ClawBio skill for verifying that scientific
+claims cite real, retrieved evidence. Your role is to mechanically bind every `[class:index]`
+citation tag in a claim to a retrieved evidence row, and to report which evidence classes a
+subject actually reached versus `NO COVERAGE` -- so an agent's drafted narrative can be checked
+before it ships, not trusted on faith.
+
+## Trigger
+
+**Fire this skill when the user says any of:**
+- "check citations" / "verify these citations" / "citation binder"
+- "does this claim have evidence" / "is this citation real"
+- "coverage report" / "what evidence classes did we reach"
+- "invented citation" / "hallucinated citation" / "no coverage"
+- "bind claim to evidence" / "unbindable claim"
+- "did every citation in this report resolve"
+- "run the claim provenance check"
+
+**Do NOT fire when:**
+- The user wants the skill to *retrieve* evidence itself (e.g. "find me trials for BRCA1"). This
+  skill only binds claims to rows it is *given*; route retrieval to `clinical-trial-finder` or
+  `lit-synthesizer` first, then bring the retrieved rows back here as a case file.
+- The user wants a judgement on whether a claim is scientifically *true*. This skill checks
+  whether a citation resolves and how many independent evidence classes support it -- never the
+  correctness of the underlying evidence itself.
+- The user has no case file (subject + rows + outcomes + claims) and no intention of assembling
+  one -- there is nothing to bind against.
+
+## Why This Exists
+
+- **Without it**: an agent drafts a claim with a citation tag that looks plausible -- `[trials:2]`
+  -- and nobody mechanically checks whether row `trials:2` was ever actually retrieved. The tag
+  reads as evidence to anyone who does not re-run the retrieval themselves.
+- **With it**: every citation is checked against the rows that were actually retrieved, in
+  milliseconds, before a report ships. A claim with one invented tag is refused outright.
+- **Why ClawBio**: this is arithmetic, not judgement -- a fixed-cost, deterministic gate that
+  catches the specific failure mode (an invented citation) that a model judge is neither
+  guaranteed to catch nor cheap to run for.
+
+## Scope
+
+**One skill, one task.** This skill binds already-retrieved evidence to already-drafted claims
+and reports evidence-class coverage for a subject. It does not retrieve evidence, does not draft
+claims, and does not judge scientific correctness. If you need retrieval too, chain with
+`clinical-trial-finder` / `lit-synthesizer` first and feed their output in as `rows`.
+
+## Core Capabilities
+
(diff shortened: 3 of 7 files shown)


manuelcorpas (changes requested):
Thanks for this. The engineering is genuinely good: standard-library only, no network surface, deterministic, real negative tests that go red when the guards are deleted, and the first output contract I have reviewed here where SKILL.md, the tests and the code all agree. The failure mode you are targeting is a real one and worth a skill.

I am requesting changes on what the artefact **says**, not on how it is built. The fixes are renaming and wiring, not redesign.

**The verdict never reads the evidence row.** `provenance_core.py:243-247`:

```python
cited_classes = {t.split(":", 1)[0] for t in tags}
if len(cited_classes) >= 2:
    return ClaimState.SUPPORTED, []
```

Using your own `demo_input.txt` rows unmodified, a claim negated to contradict every row it cites still returns `SUPPORTED`:

```
"Expression is REDUCED against matched normal tissue [expression:0], and no
 independent group replicates the direction [literature:0]."
```

`SKILL.md:97-99` correctly says the skill does not judge truth, so you clearly understand the boundary. But the emitted token `SUPPORTED` and the disclaimer at `claim_provenance_check.py:36-38` ("SUPPORTED / CONTESTED describe what the cited evidence rows say") do not honour it. Cited-but-contradicted is stamped identically to cited-and-correct, in the rendered `report.md`, which is exactly the document a downstream reader trusts instead of re-running retrieval.

**Class independence is computed from the tag prefix, never from `source_class`.** `SKILL.md:178-179` states the rule in terms of `source_class` values, but the code reads `t.split(":", 1)[0]` off the claim string, and nothing validates that the two agree. Two rows both marked `"source_class": "pubmed"` but tagged `a:0` and `b:0` return `SUPPORTED`, i.e. "2+ independent evidence classes" for two rows from one source. `SKILL.md:251` calls this a load-bearing Domain Decision; right now it is inferred from an unvalidated substring of a string the drafting model wrote.

**`CONTESTED` means two unrelated things, and the one a reader assumes is unreachable.** `disagreeing_tags` is not in `REQUIRED_CASE_KEYS` (`claim_provenance_check.py:45`) and not in the documented input schema (`SKILL.md:22-26`), so via `--input` or `--demo` the only route to `CONTESTED` is line 247, "exactly one cited class". Your own shipped demo shows this: claim 2 correctly reports a failed trial, cites the one row saying the trial failed, and renders as `**[CONTESTED]**`. `test_disagreement_is_contested` passes because it calls `state_for_claim` directly with the kwarg, so the suite reports a feature the CLI cannot reach.

**The `NO COVERAGE` disclaimer is false for the verbatim path.** `claim_provenance_check.py:38-40` asserts `NO COVERAGE` means the claim "could not be bound to any retrieved evidence row", but `state_for_claim` returns it on *any* binder issue, including verbatim overlap, where the citations bound perfectly. Ordinary biomedical prose trips it:

```
row:   "No benefit was seen in patients with metastatic non small cell lung cancer ..."
claim: "The study reported no signal in patients with metastatic non small cell lung cancer [trials:0]."
-> ('NO COVERAGE', ['verbatim-run: in patients with metastatic non small cell lung'])
```

Any disease name of four or more words plus a short shared stem hits the eight-word threshold. `SKILL.md:276-279` restates the same false equivalence to the agent.

**The verbatim guard is calibrated backwards on both tails.** It is a contiguous-run check, so one interior edit defeats it. Your own demo claim 2 lifts eleven of twelve content words from `trials:0` in order and scores zero hits, because "early" was deleted and the two-word opener swapped. So it fires on honest paraphrase and is close to free to evade, while `SKILL.md:296` calls no-partial-credit "the load-bearing safety property".

**There is no provenance field.** `EvidenceRow.url` is optional and there is no required identifier, retrieval timestamp or checksum. A case file of entirely invented rows returns `SUPPORTED`. `SKILL.md:258-260` states the tags-are-retrieval-layer-only rule, but it is enforced only by asking the model not to cheat. For a skill named "provenance check" this is the central gap. It verifies internal consistency of a document with itself, which is a real and useful property, just not provenance.

Suggested changes:

1. Rename the states to what is actually computed. `CITED` / `SINGLE-SOURCE` / `UNBOUND` would be true statements. Reserve `SUPPORTED` for a check that reads the row.
2. Split the verbatim failure out of `NO_COVERAGE` into its own state or issue class, and correct the disclaimer.
3. Compute independence from `EvidenceRow.source_class`, and validate in `__post_init__` that the tag prefix equals it, or drop the independence claim.
4. Either thread `disagreeing_tags` through the case file and `REQUIRED_CASE_KEYS`, or delete the parameter and its documentation.
5. Change `domain: clinical` (`SKILL.md:8`); there is no biomedical logic in the 455 lines of Python, no PMID, DOI, accession, gene or variant awareness. Declaring it as report QA for other skills' output is honest and the chaining story with `lit-synthesizer` and `clinical-trial-finder` is coherent.
6. Narrow the `coverage report` trigger keyword. In a genomics repo "coverage" means read depth; someone asking for a coverage report on a BAM should not land here. No collision with other skills' declared keywords, so "evidence coverage report" fixes it.
7. Optionally, require at least one non-null provenance field per row so the name is earned.

Smaller: `SKILL.md:199` labels the Example Output "unedited" but the disclaimer is elided with `[...]` and the inner fences are escaped. Everything else in that sample does match what the code produces, and I verified the demo's headline numbers reproduce (3/5 classes, $0.0062, 1 SUPPORTED, 1 CONTESTED, 2 NO COVERAGE). No fabricated output, no `np.random`, no hardcoded allowlists anywhere.

With the renaming and wiring above this is a merge.

manuelcorpas (commented):
## pr-audit: advisory

Security surface is essentially nil, and unusually so: no network, no subprocess, no eval, no third-party imports, pure stdlib. The demo data is exemplary for this skill category, with every row labelled synthetic and no real-looking PMIDs or DOIs to mistake for real ones. The Trigger, Scope and Agent Boundary sections are disciplined and correctly refuse the truth-adjudication role.

The findings are all semantic, which in a provenance skill is where the damage lives.

### CRITICAL, `provenance_core.py:135` with `claim_provenance_check.py:37`

`SUPPORTED` is computed without ever comparing the claim to the cited evidence's content, yet the shipped disclaimer tells the reader it did.

`row.text` is used in exactly one place in the whole core:

```python
135:        row_words = _words(row.text)
```

and that call site is `check_no_verbatim`, which *penalises* the claim for resembling its source. So the only reading of the evidence text is one that rewards divergence from it. State selection never touches it:

```python
243:    cited_classes = {t.split(":", 1)[0] for t in tags}
244:    if len(cited_classes) >= 2:
245:        return ClaimState.SUPPORTED, []
```

A claim asserting the exact opposite of the row it cites ships as `SUPPORTED`; a claim that accurately restates it is refused as `verbatim-run`. Meanwhile every `report.md` carries:

```python
37:    "judgement. SUPPORTED / CONTESTED describe what the cited evidence rows say, not whether "
```

SKILL.md is careful about this elsewhere at lines 97-99 and 116-119, which makes the disclaimer the one place a reader is actively misinformed. Small fix: say that `SUPPORTED` means the cited tags resolve and span two or more classes, with content agreement explicitly unchecked.

### Warnings

- `provenance_core.py:240-247`. `CONTESTED` conflates "evidence disagrees" with "only one source", and both paths return the same enum with an empty issue list. The demo's own output at SKILL.md:221 renders a single-source, perfectly valid trial finding as `[CONTESTED]` with no explanation, which any reader takes to mean disputed. Suggest a distinct `SINGLE_SOURCE` state or a reason field.
- `claim_provenance_check.py:77` against SKILL.md:141 and 180. `disagreeing_tags` is documented as caller-supplied but is unreachable from the only entry point, and `load_case` never reads such a key. The genuine-disagreement half of `CONTESTED` is therefore dead in production.
- `tests/test_claim_provenance_check.py:87-88`. `test_short_claims_do_not_trip_verbatim` passes with the guard it names deleted: remove `if len(claim_words) < run: return []` at `provenance_core.py:128-129` and `range(2 - 8 + 1)` is empty anyway, so nothing can match.
- `provenance_core.py:171-173`. Verbatim lifting from a row the claim did not cite is not detected, since `cited = [r for r in rows if r.tag.lower() in set(tags)] or list(rows)` only widens the net when nothing matched at all.
- SKILL.md:336 and `provenance_core.py:21-23`. The only source cited is Sippar's `reportQuality.ts` in `Nuru-AI/sippar`, which SKILL.md itself says is private and which returns 404 under my token. The load-bearing constant `VERBATIM_RUN_WORDS = 8` is justified as "chosen empirically upstream" against that same unreachable source. Under this skill's own rules that is `citation-invalid`. Either publish the derivation or drop the empirical claim and call 8 a chosen default.

### Correcting our own tooling

pr-audit reported "no corresponding test file found for `provenance_core.py`". That is a **false positive** from filename matching, and I am noting it so it does not count against you. The test file imports seven symbols from `provenance_core` at lines 27-36, and 14 of its 20 tests exercise the core. `test_invented_citation_is_caught`, `test_uncited_claim_is_caught`, `test_verbatim_copying_is_caught` and the end-to-end refusal tests all fail if the verification logic is removed. You did the work.

manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

## Domain review: claim-provenance-check

Thank you for this. The skill is genuinely self contained and the refusal-first framing is the right one for a citation gate.

**What is right.** Imports are standard library only at `skills/claim-provenance-check/provenance_core.py:30` to `:33` and `skills/claim-provenance-check/claim_provenance_check.py:19` to `:31`, so the zero-network claim at `skills/claim-provenance-check/SKILL.md:289` holds; nothing leaves the machine. Malformed input raises rather than returning a plausible-looking result, at `claim_provenance_check.py:56` and `:61`, so there is no silent-failure path. I traced the bundled demo by hand and the example block at `SKILL.md:215` is what the code actually produces: coverage 3/5, cost 0.0062 from `demo_input.txt:26` to `:28`, one SUPPORTED, one CONTESTED, one `citation-floor`, one `citation-invalid`. The CONTESTED claim at `demo_input.txt:44` shares a 6 word run with its row text at `demo_input.txt:21`, correctly under the 8 word threshold at `provenance_core.py:104`. CLI registration is correct: `run_skill` appends `--input` and `--output` itself at `clawbio/cli.py:1507` and `:1521`, so the empty allowlist at `clawbio/cli.py:1227` does not break the documented usage at `SKILL.md:155`. The catalogue `status: planned` matches the generator default at `scripts/generate_catalog.py:461`.

Three things to resolve before this ships.

### 1. SUPPORTED is decided by the tag prefix, not by `source_class`

`provenance_core.py:243` computes `cited_classes = {t.split(":", 1)[0] for t in tags}`, taking the evidence class from the text of the citation tag. `SKILL.md:177` and `SKILL.md:251` both state the rule as "2+ distinct `source_class` values". Those are not the same thing. `EvidenceRow` at `provenance_core.py:57` carries `tag` and `source_class` as independent caller-supplied fields, and `__post_init__` at `provenance_core.py:73` only checks that the tag contains a colon; nothing requires the prefix to equal `source_class`.

Consequence: two rows both with `source_class` of `literature` but tagged `pubmed:0` and `europepmc:0` make a claim citing both SUPPORTED. One source class is then presented as independent corroboration, which is the exact over-call this skill exists to prevent, and it errs in the direction that ships a claim rather than refuses it. The reshaping step you describe for `clinical-trial-finder` and `lit-synthesizer` at `SKILL.md:311` is precisely where a hand-chosen prefix would diverge. The rows are already in scope at `provenance_core.py:224`, so either map cited tags back to their rows and read `source_class` from there, or reject a row whose tag prefix differs from its `source_class` in `__post_init__`.

### 2. The tests cannot catch that

Every fixture row at `skills/claim-provenance-check/tests/test_claim_provenance_check.py:39` to `:41` uses a tag whose prefix equals its `source_class` (`trials:0`/`trials`, `literature:0`/`literature`, `ip:0`/`ip`), and so does every row in `demo_input.txt:6`, `:13` and `:19`. `test_two_independent_classes_is_supported` at `tests/test_claim_provenance_check.py:106` passes under either rule, so it does not pin the documented one. A single case with a divergent prefix would.

### 3. `disagreeing_tags` is unreachable through the documented interface

`state_for_claim` accepts `disagreeing_tags` at `provenance_core.py:227`, but `evaluate_claims` calls it without that argument at `claim_provenance_check.py:77`, and neither `REQUIRED_CASE_KEYS` at `claim_provenance_check.py:45` nor the case-file schema at `SKILL.md:22` has a field for it. `SKILL.md:141` and `SKILL.md:179` describe CONTESTED-on-disagreement as part of the workflow, so as shipped that route can never fire from `--input` or `--demo`. `test_disagreement_is_contested` at `tests/test_claim_provenance_check.py:117` calls the core directly and so does not notice. Either plumb the field through the case file or drop it from the documented workflow.

### Minor

- `bind_claim` at `provenance_core.py:172` falls back to checking the claim against every row when no cited tag resolves. That only fires on claims which have already failed, so it is harmless, but it can attach a `verbatim-run` issue to a claim refused for `citation-floor`. Confirm that is intended.
- The frontmatter declares `domain: clinical` at `SKILL.md:10`, yet nothing here touches clinical or genomic data. Whether a general report-QA tool belongs in the catalogue, and under which domain, is a maintainer call rather than a defect.
- `SKILL.md:336` cites a private Sippar repository, so the port claim at `provenance_core.py:21` cannot be verified from the diff. Please confirm you hold the rights to relicense that logic MIT.

### Static scan

The static scan reported no findings on this PR. I re-read the diff against it and have nothing to contest. The high-yield vocabulary class does not apply here: no literal in this diff is matched against ClinVar, VEP, gnomAD, CPIC or PharmGKB output.

Holding at NEEDS_REVIEW on item 1, which is a scientific over-call in the shipping direction, with items 2 and 3 alongside it.

---

manuelcorpas:
Closing for now: no activity since 18 August and the branch now conflicts with main. The three items from the 29 August review still stand (SUPPORTED decided by tag prefix rather than source_class, tests that cannot distinguish the two, and the disagreement path unreachable from the case file). The idea of a refusal-first citation gate is a good one, and if you pick it up again, reopen or open a fresh PR against current main and it will be reviewed promptly. Thank you for the contribution.

PR #347, opened 2026-08-18
skills: add abstention-ledger, and a parser for the legacy SnpEff EFF field

x4x4x:
Built at the Berlin hackathon for Challenge 1. Standard library only — no new dependency, no API key, no GPU.

## The gap this closes is one the docs already name

The Challenge 1 brief warns, verbatim:

> Do not feed the historical quartet directly to `rare-high-impact-variants`: it does not parse the legacy EFF field.

That is accurate. `rare_high_impact_variants.py:126` reads the consequence from `MC`, `Consequence` or `ANN`; classic SnpEff writes `EFF`. So the skill whose report explains the population-frequency blind spot receives an empty consequence for every record in that format.

**The failure mode is worse than an error.** Measured, not asserted — `prove_gap.py` runs the unmodified skill three times over the same 68 records and reads the numbers out of its own `result.json`:

| Metric | raw `EFF=` only | `MC=` unmapped | `MC=` mapped |
|---|---|---|---|
| `variants_processed` | 68 | 68 | 68 |
| `carried_variants` | 68 | 68 | 68 |
| `high_impact_carried` | **0** | **29** | **68** |
| exit code | 0 | 0 | 0 |

Column 1 is a false negative that reads as a clean run: processed and carried counts correct, every impact metric zero, nothing on stderr.

Column 2 is the part worth a maintainer's attention. Setting `MC=` to the SnpEff effect name is the obvious one-line fix and it is *worse* than the bug: `_is_high_impact` is an unanchored case-insensitive substring test, and SnpEff's names agree with three of the six LoF terms by coincidence (`STOP_GAINED`, `START_LOST`, `STOP_LOST` match; `FRAME_SHIFT`, `SPLICE_SITE_DONOR`, `SPLICE_SITE_ACCEPTOR` do not). A silent 29 is harder to catch than a silent 0. `eff_to_info.py` maps all six explicitly and raises `UnmappedEffect` rather than guessing.

## What is in the PR

| File | What |
|---|---|
| `legacy_eff.py` | Parses `EFF` into **one record per transcript** instead of collapsing to `max(impact)` |
| `eff_to_info.py` | Adapter to the INFO keys existing skills read — makes `rare-high-impact-variants` work on the legacy corpus |
| `prove_gap.py` | Runs the unmodified target skill three ways and measures the difference |
| `gates.py` | Cohort abstentions and per-record gates |
| `abstention_ledger.py` | CLI, segregation, sample-map resolution, rendering |
| `fetch_evidence.py` | Build-matched GRCh37 evidence layer, cached to disk |
| `lint_claims.py` | Fails the build on a claim the input cannot support |
| `verify.sh` | Tests + demo self-consistency + lint + secret scan, non-zero on failure |

Per-transcript retention is not only a fix, it is a gate: **22 of the 68 records carry a HIGH impact that is not reproduced across other transcripts of the same gene**, so the HIGH label is a transcript-selection decision. A `max(impact)` pipeline reports those as HIGH with no hint that an equally annotated transcript disagrees.

## Reuse rather than duplication

`SECONDARY_FINDING_NO_CONSENT` imports `ACMG_SF_V32_GENES` from `skills/clinical-variant-reporter/acmg_engine.py` rather than keeping a second copy of a curated gene list. There is a provenance-carrying snapshot at `data/acmg_sf_v32_snapshot.json` used **only** when that import misses (the skill is also distributed standalone); the import stays the primary path and every report prints which one ran.

On the challenge data that gate fires **zero times** — all 68 screened, no gene on the list — and the report says so, because a check that ran and found nothing is a result, not an absence of work.

## Registration

Added to `clawbio/cli.py` and `skills/abstention-ledger/INTENTS.json`, so `python clawbio.py run abstention-ledger --demo` works.

Two things deliberately left undone rather than half-done:
- `bio-orchestrator` keeps its own keyword table separate from the CLI registry and still does not list this skill.
- `skills/catalog.json` is **not** regenerated: `scripts/generate_catalog.py` did not complete in this environment (>2 min, no output). Happy to add the entry however you prefer.

## Conformance

All 17 SKILL.md checks pass. `SKILL.md` is 386 lines. Every generated report carries the mandated disclaimer verbatim, asserted by a test rather than left to reviewer discipline. Demo data is synthetic.

```bash
sh skills/abstention-ledger/verify.sh .        # tests, demo, lint, secrets
python3 skills/abstention-ledger/abstention_ledger.py --demo --output /tmp/al
```

## Two bugs found in passing

Not fixed here, since they are outside this skill's scope, but both reproducible in a clean clone:

1. **`clinical-variant-prioritizer` is non-functional.** `data/clinical_panel.json` does not exist — the `data/` directory is absent — so `api.run()` raises `FileNotFoundError` on every call and its test file fails. `prioritize.py` has no `argparse`, no `main()`, no `__main__`, yet `catalog.json` advertises `--demo`.
2. **`bio-orchestrator --demo` does not exist.** `catalog.json` advertises the command; argparse has no such flag.

Also worth passing on: the Ensembl **GRCh37** REST endpoint answers **503 Service Unavailable**, not 400, for parameters it does not support (`mane`, `af_1kg`, `af_gnomade`, `af_gnomadg`). It reads as an outage. `PARAMS` in `fetch_evidence.py` is verified parameter by parameter, with a comment.


manuelcorpas:
Thank you for documenting the legacy `EFF` parsing failure in such concrete terms. I saw that you closed pull request 347 and deleted the branch. I do not want the bug evidence to disappear with the larger event patch. Would you be willing to open a focused issue containing the three measured cases, or a much smaller parser fix? I can help define a reviewable scope.

PR #348, opened 2026-08-18
Add skill: prs-abstain

abotescu20-maker:
## What this adds

`prs-abstain` decides whether a polygenic score percentile may be reported for a given
individual, and withholds it when it cannot be interpreted. It does not infer ancestry
(`ancestry-risk-profiler`, `claw-ancestry-pca`) and does not compute scores (`gwas-prs`).
It sits between them.

Built for Challenge 3, ClawBio + Nebius Berlin, 18 August 2026.

## Why

Every curated score in the library discloses `Reference population: EUR` and still returns a
confident percentile for any genome. Disclosure is not abstention.

While building this we found the reference mean is exactly `Sum(2 * AF_EUR(i) * w(i))` for all
six curated scores (PGS000013 1.1186 vs 1.12; PGS000004 2.8428 vs 2.84; PGS000057 7.1080 vs
7.11). The percentile is centred on a European allele-frequency calculation, so the error
decomposes per variant rather than being a property of the whole score.

## Findings from the existing demo data

- `gwas-prs --demo` returns a breast cancer **and** a prostate cancer percentile for the same
  individual. At least one is inapplicable.
- PGS000001 and PGS000039 each carry two different rsIDs at one coordinate
  (`chr1:114448389` = rs11552449 + rs7072776). Coordinate error or a locus counted twice;
  either corrupts the sum. This blocks the score.
- PGS000013's two largest weights (rs7903146, rs12255372) are one locus, TCF7L2, 50 kb apart
  and 48% of the score. Effective n is 3.5, not 6.
- PGS000057 looks like 147 independent contributions and behaves like 35.9.

## Checks, in order

1. Trait applicability (sex-specific scores)
2. Score integrity (weight coverage, concentration, strand-ambiguous, duplicate positions)
3. Ancestry: markers >= 30 (Kosoy et al. 2009, PMID 18683858), placeable, distance <= threshold
4. Reference-population match

Raw scores are retained in every branch. Only the percentile is withheld: summing weights is
valid for anyone, converting that sum to a percentile needs a reference distribution that fits.

## Validation

- 59 tests, written red/green before implementation
- 17/17 SKILL.md conformance; 10/10 on the submit-page validator
- 120 tests pass across `prs-abstain`, `gwas-prs`, `equity-scorer` (no regressions)
- Stress-tested 10 ways; three regressions found and fixed, all now Gotchas
- `--k-sd 20` is detected and stamped THRESHOLD OVERREACH rather than silently passing
  non-European individuals

## Demo

```bash
python skills/prs-abstain/prs_abstain.py --demo --output /tmp/prs-abstain
```

Three individuals, three verdicts, two of them deliberate refusals. Writes a jargon-free
clinician report and a technical report, both also as PDF.

## Honest limitations

LD is approximated by physical distance, not measured, so the grouped effective_n remains an
overestimate. Admixed individuals are the principal unhandled case. The bundled allele-frequency
table is synthetic and labelled as such in the file. Twelve limitations are listed in the
technical report and in SKILL.md.


manuelcorpas (changes requested):
This is careful work and the hard part is right. Stating that first because most of what follows is a list of problems, and the ratio would otherwise mislead.

The ancestry logic is the part I expected to fail, and it does not. The skill never infers ancestry (PCs are supplied by the caller), never emits a hard super-population label for a query individual, never produces admixture proportions from a thin panel, and refuses rather than imputing missing coordinates to the origin, with `SKILL.md:272-275` explaining precisely why the origin would read as maximally European. The threshold is calibrated (mean + k*sd of within-reference spread, `min_reference_n=10`) rather than asserted, and the `THRESHOLD OVERREACH` detector that voids its own `REPORT` verdicts is a genuinely good idea. I verified k=3 gives 3.4657 on the bundled panel, sitting between `within_max` 3.1356 and `nearest_other` 7.3788. Deleting the abstention guard turns four tests red. Tracing every path that can emit a percentile: there is no try/except, no default ancestry and no closest-population fallback in the gate.

Requesting changes on the tiers around that gate, and on one thing that is our fault rather than yours.

**Not your bug, but it blocks this PR: five of the six bundled PGS accessions carry the wrong trait.** Verified live against the PGS Catalog REST API:

| File header claims | PGS Catalog |
|---|---|
| PGS000001 Breast cancer, Mavaddat 2015, 77 | correct |
| PGS000004 Coronary artery disease, Abraham 2016, 46 | **Breast cancer**, Mavaddat 2018, **313** |
| PGS000011 Atrial fibrillation, Tada 2014, 12 | **Coronary artery disease**, Tada 2015, **50** |
| PGS000013 Type 2 diabetes, Vassy 2014, 8 | **Coronary artery disease**, Khera 2018, **6,630,150** |
| PGS000039 Body mass index, Khera 2019, 97 | **Ischemic stroke**, Abraham 2019, **3,225,583** |
| PGS000057 Prostate cancer, Schumacher 2018, 147 | **Coronary heart disease**, Natarajan 2017, **57** |

Three of the cited PMIDs also point at unrelated papers: `25087147` is a ventilator-associated-pneumonia nursing paper, `27535533` is the ExAC paper, `30239722` is Pulit et al. on body fat distribution.

Your files are byte-identical to `skills/gwas-prs/data/*.txt`, and the wrong map lives in `skills/gwas-prs/curated_scores.json` on `main`. You copied what the repo told you was authoritative, so this is ours to fix and I have opened a separate issue for it.

It blocks here because of what this PR changes: `gwas-prs` prints a terminal report, `prs-abstain` produces a typeset A4 PDF headed "Summary for clinicians", designed to be handed over. Tracing your demo for `EUR_001`, three percentiles are released: Type 2 diabetes 14th, Atrial fibrillation 98th, Coronary artery disease 8th. The underlying accessions are a CAD score, a CAD score and a breast cancer score. A mislabelled 98th-percentile atrial fibrillation result in a handover document is the harm this skill exists to prevent. Either wait for the `gwas-prs` fix, or read from a single corrected location rather than copying.

Relatedly, please withdraw the "real PGS Catalog files, real PMIDs" claim at `DEMO.md:122` and `SUBMISSION.md:31`. For a skill whose thesis is that disclosure is not abstention, an unverified provenance claim about its own inputs is the one thing it cannot afford.

**The score-integrity gate is silently skipped without `--genotype`.** `prs_abstain.py:1308-1314`:

```python
lds[pid] = ld_audit(sdef, window_kb=args.ld_window_kb)
if genotype:
    audits[pid] = audit_score(sdef, genotype)
    integrity[pid] = integrity_verdict(...)
```

`--genotype` is optional. Without it `integrity` stays empty, so at `gate_scores:646-649` `integ is None` and no integrity reason is ever added, with no warning that the tier was skipped. Duplicate genomic positions are a property of the scoring file alone; `ld_audit` computes them one line above and needs no genotype. `SKILL.md:337` states that rule as blocking. `--demo` sets `--genotype` at `:1275`, so no test can see this. The same conditional also disables the 0.90 weight-coverage refusal.

**The gate fails open on absent metadata, twice.** At `:654` a score whose `reference_population` is missing adds no reason, so the percentile is released, and the note at `:668-670` then asserts a provenance never established: "Reported against the EUR reference distribution." Workflow step 9 promises the opposite. `test_score_reference_population_mismatch_is_flagged` covers only the explicit `"EAS"` case. Separately, the sex-applicability gate at `:223` keys on free text against seven hardcoded keywords, so an absent, null or unusually phrased trait is declared applicable. For a gate whose purpose is refusing to show a prostate score to a woman, unrecognised should fail closed.

**`load_score_definitions` cannot read a real PGS Catalog file.** `:171-185` ignores the header row and hardcodes positional indices. The authentic `PGS000001_hmPOS_GRCh37.txt` has column order `rsID, chr_name, effect_allele, other_allele, effect_weight, locus_name, OR, hm_source, hm_rsID, hm_chr, hm_pos`, with no `chr_position` at index 2 and no allele-frequency column at all. Applying your indices gives `float('CCND1')` and an unhandled `ValueError`, after the output directory has already been created. Where a numeric column happens to sit at index 5, it fails silently instead. Parse by header name and add one genuine harmonised download as a fixture. Related: `:1318` does not guard `expected_mean` returning `None`, so `raw - None` raises whenever a scoring file carries no allele frequencies, which real ones frequently do not.

Smaller items:

- `SKILL.md:225-231` Example Output does not reproduce: I recompute EUR_001 = 1.7413 and AFR_001 = 10.3770 from your own bundled PCs, against the 1.44 and 9.87 shown. Your `DEMO.md:57-59` prints the correct values, so SKILL.md is just stale. The note string `Withheld: REFUSE_DISTANT.` also does not exist in the code. Everything else reproduces exactly, including the threshold 3.47, `within_max` 3.14, `nearest_other` 7.38, the PGS000057 effective-n drop 122.5 to 35.9, the TCF7L2 pair at 48% of the score, and the AFR shift of 1.28 sd with rs7903146 contributing 92%.
- `n_markers_shared` (`:482-483`) is trusted from a CSV column and never cross-checked, not even against the genotype file when one is supplied. Absent or non-numeric correctly fails closed, but an inflated value makes the marker gate pass unconditionally. Worth saying so in `KNOWN_LIMITATIONS`.
- `audit_score:263` counts a variant as covered on rsID presence alone; the call itself is never inspected, so a `--` no-call counts as full weight. Both that and the unchecked allele compatibility push `weight_coverage` up, the direction that lets a score through the 0.90 gate. Your own `SKILL.md:288-290` states the intended behaviour.
- Please delete `DEMO.md`, `HOWTO.md`, `PR_BODY.md` and `SUBMISSION.md`. No other skill of 96 carries these. `SUBMISSION.md:6` still says to post a link in `#berlin-demos` before 16:10 and contains a `<YOUR-FORK>` placeholder, and `DEMO.md:146` points at a sandbox path. The flag table at `HOWTO.md:60-77` is genuinely useful and should fold into SKILL.md's CLI Reference.
- Regenerate `skills/catalog.json` with the same `ensure_ascii` setting the catalogue bot uses. The regenerated file re-escapes every non-ASCII character to a `\uXXXX` sequence, so 20 of the 35 changed lines are punctuation churn in unrelated skills' descriptions. That will conflict on rebase.
- `pdf_report.py:159` imports PIL and the tests import pypdf, neither declared anywhere.

Two boundary tests would be worth adding: `decide` uses `dist > cal.threshold` and `n < min_markers`, so exactly-at-threshold reports and exactly-30-markers passes, and neither edge is exercised.

On the two duplicate-position defects you found: both are real, but they are not lift-over artefacts. rs7072776 is on chromosome 10 (GRCh38 10:21,744,013), yet your `PGS000001` line 77 places it at chr1:114448389, which is rs11552449's correct GRCh37 coordinate. Same shape for rs3101336 and rs2815752 in `PGS000039`. In both pairs one variant is correct and the other has been handed a copy of it, which is a signature of synthetic file construction rather than a coordinate error. Good catch either way.

Data movement is clean: no http, s3, gs, ftp, urllib, requests, socket or os.system anywhere. The PDF safety work is well done, in particular that `_write_clinician_report` never renders `risk_category` and that the "not evidence of low risk" sentence is asserted present in the rendered output.

manuelcorpas:
Filed the scoring-file provenance problem separately as #356, since it lives on `main` in `skills/gwas-prs` and predates this PR. Nothing for you to fix there; this PR just needs to not inherit it before the clinician PDF ships.

manuelcorpas:
CI has now run (I approved it) and `skill-lint` fails on one line, separately from the review findings:

```
Validation failed for skills/prs-abstain:
  - Unexpected fields in frontmatter: data_license. Only ['allowed-tools',
    'compatibility', 'description', 'license', 'metadata', 'name'] are allowed.
```

`data_license` needs to move under `metadata:` rather than sitting at the top level. Small fix, and worth doing in the same pass as the rest.

abotescu20-maker:
Thank you for the review — its depth made this straightforward to act on. Three commits address it point by point. All 72 skill tests pass (59 original + 13 new), 133 across prs-abstain, gwas-prs and equity-scorer.

**Provenance claim** — withdrawn. DEMO.md, HOWTO.md, PR_BODY.md and SUBMISSION.md are deleted (they were hackathon artefacts; the flag table from HOWTO now lives in SKILL.md's CLI Reference), and the "real PGS Catalog files, real PMIDs" assertion went with them.

**Integrity tier without `--genotype`** — now runs unconditionally. Duplicate positions and weight concentration are checked from the scoring file alone; when no genotype is supplied the verdict records exactly which genotype-dependent checks were skipped, instead of skipping the tier silently. A test asserts the duplicate-position reason reaches the gate with no genotype at all.

**Fail-open on absent metadata** — both closed. A score with no `reference_population` now withholds the percentile ("a percentile of unknown provenance cannot be released"), and the report note can no longer assert "Reported against the EUR reference distribution" for a score that never declared one. Sex-applicability fails closed on an absent or unreadable trait; the residual keyword-list limitation is stated in KNOWN_LIMITATIONS.

**Parser** — rewritten to resolve columns by header name (hm_* preferred), never by position. A genuine PGS000001 harmonised download from the EBI FTP is now a test fixture and parses correctly, including `expected_mean` returning None on its missing AF column instead of raising. A malformed effect_weight raises naming file:line rather than skipping silently.

**Coverage counting** — a `--` no-call and an allele-incompatible call no longer count toward weight coverage (a strand-complement call still does). Fail-closed direction, since coverage feeds the 0.90 release gate.

**skill-lint** — `data_license` moved under `metadata:`; verified locally with the exact CI command (`uv run agentskills validate skills/prs-abstain` passes).

**Smaller items** — stale SKILL.md example output fixed (1.74 / 10.38, real note string); `n_markers_shared` trust boundary added to KNOWN_LIMITATIONS in your wording; reportlab/pillow declared as an optional dependency group and pypdf in the dev group; catalog.json regenerated with the catalogue bot's `ensure_ascii=False` — the diff against main is now only the new entry; both boundary tests added (exactly-at-threshold reports, exactly-30-markers passes).

**On the #356 dependency** — rather than waiting for the corrected files, we'd propose reading scoring files from a single shared location (`skills/gwas-prs/data`) so the fix lands once and prs-abstain inherits it. Happy to sequence it the other way if you prefer; the demo currently still runs on the bundled copies, now labelled as inherited pending #356.

And thank you for tracing the duplicate-position pairs to their real cause — we had read them as coordinate errors; one variant carrying a copy of the other's coordinates is a cleaner explanation and consistent with what #356 found about the files' construction.

manuelcorpas (changes requested):
Re-audit at `d69d52a1`. Three of your four headline claims hold up under checking, and two of the fixes are genuinely good work.

**Verified as properly fixed:**

- **Parsing (`0b173c8d`).** `_resolve_columns:172-187` maps by header name with `hm_*` preferred and raises naming the missing field. The authentic 77-variant fixture is exactly the right move: it asserts `hm_chr`/`hm_pos` resolution, the full-precision weight `0.16220387987485377`, `af_reference is None` and `expected_mean(s) is None`. That is testing against a real artefact rather than against what you imagined the format to be.
- **Coverage.** `_call_usable:330-346` rejects any call with no ACGT base and requires the bases to be a subset of the declared pair or its strand complement. Palindromic sites resolve the same both ways so they count as covered rather than being silently mis-resolved, warned above 10 percent and documented. That is a defensible position, clearly reasoned. Three new tests, all red on revert.
- **Both fail-open defaults.** Absent `reference_population` now appends a reason and the note can no longer assert an origin that was never established; `check_applicability:287-292` fails closed on absent, empty, `unknown`, `none`, `null`, `nr` and `na`, and `:293-301` on a sex-specific trait with absent sex. The tests cover the absent cases specifically, not just explicit mismatch.
- **Items 4, 7, 8, 9.** `KNOWN_LIMITATIONS:1297-1299` now states verbatim that `n_markers_shared` is never cross-checked "not even against the genotype file when one is supplied" and that an inflated value passes unconditionally, which is what was asked. All four scaffolding files are gone, not two. The catalog diff is now one entry plus `skill_count`. Frontmatter parses to exactly the four allowed keys with `data_license` intact inside `metadata`.
- **Everything I credited last time still holds:** PCs supplied rather than ancestry inferred, no super-population label for an individual, no admixture, missing coordinates refused rather than imputed to the origin, threshold calibrated (I reproduced 3.4657 independently), THRESHOLD OVERREACH detector firing in all three surfaces, "Research tool. Not a medical device." on every PDF page, and `risk_category` never rendered.
- **Example Output now reproduces.** I recomputed from your own `demo_reference_pcs.csv`: EUR_001 = 1.7413, AFR_001 = 10.3770, threshold = 2.0142 + 3 x 0.4838 = 3.4657. Your 1.74 / 10.38 / 3.47 are right, as are 3.14, 7.38 and the three AF expectation figures. The fabricated `Withheld: REFUSE_DISTANT.` string is gone.

Requesting changes on two things, one of which is the same class the round set out to close.

**1. "Every gate fails closed" is not yet true. The integrity tier still fails open on the command your own docs call standard.** `48c0997f` correctly moved the computation out of `if genotype:` and made `integrity_verdict` record a skip warning. But the consumer at `prs_abstain.py:755-758` was left unchanged:

```python
integ = (integrity or {}).get(pid)
if integ is not None:
    reasons.extend(f"Score integrity: {r}" for r in integ.reasons)
```

`--scores` is optional and is absent from the required list at `:1398-1402`, so `score_defs` is `{}`, `integrity` is `{}`, every `pid` resolves to `None`, and the tier contributes no reason, no warning and no caveat. `SKILL.md:169-172` gives exactly that invocation as "Standard usage" with no `--scores`.

Concretely, on your own bundled demo files via the documented standard command: EUR_001 is female, 480 markers, distance 1.7413, verdict REPORT. PGS000001 carries the duplicate-position pair `rs11552449` / `rs7072776` at chr1:114448389, which I confirmed. Under `--demo` that score is blocked. Under the standard command it is released at percentile 94.2 with the note "Reported against the EUR reference distribution." and no caveat. `SKILL.md:351` says duplicate positions "block rather than warn". The same silent skip hits any `pgs_id` in `prs_results.json` with no matching scoring file even when `--scores` is given.

Treat a missing `IntegrityVerdict` the way `integrity_verdict` already treats a missing audit: a recorded reason, or at minimum a caveat, never a pass.

**And reverting the fix you did make leaves all 72 tests green.** `test_integrity_tier_runs_without_genotype` calls `integrity_verdict(None, ld=ld)` directly and `test_integrity_reasons_reach_the_gate_without_genotype` hand-builds the dict; neither goes through `main()`. All 22 CLI invocations in the suite pass `--demo`, which always sets both `--genotype` and `--scores`. So moving `integrity[pid] = ...` back inside `if genotype:` breaks nothing. The defect was in the wiring and the tests sit on the units either side of it. A CLI test with no `--genotype`, and one with no `--scores`, asserting the PGS000001 block, closes both this and the item above.

**2. Rebase onto `main`.** Your branch predates the #357 merge, and the six bundled scoring files are byte-identical to the pre-#357 `gwas-prs` blobs. They still carry `#pgs_id=PGS000013` and `#pgs_name=PRS for type 2 diabetes (Vassy et al. 2014)`, which #357 replaced with `#clawbio_panel_id=CLAWBIO-T2D-8` and an explicit "This is NOT a PGS Catalog scoring file ... Do not cite this file as that accession". None of that is your fault, it was a defect on `main` filed as #356, but this PR duplicates the pre-fix files into a second location and propagates the accessions into a document headed "Summary for clinicians".

Rebasing is necessary but not sufficient: `load_score_definitions:246` falls back to `fp.stem.split("_")[0]`, so it would re-derive `PGS000013` from the filename even once the header key is gone. Worth reading `#357` to see the shape of the fix.

Two smaller ones, not blocking:

- `prs_abstain.py:1442-1443`: a `z_score` of exactly 0.0, entirely plausible for someone at the mean, makes `z` falsy, so `sd_ref` is `None` and the denominator silently becomes `1.0`. The technical report then prints that quotient under a column headed "shift (sd)" and `gate_scores:776` raises a caveat off it. Same class as the reference-population defect you just fixed: a quantity presented in units whose provenance was never established. Propagate `None` and say so.
- Catalog entry still records `"version": "0.1.0"` and `"data_license": ""` against `0.3.0` and `CC0-1.0` in SKILL.md. Regenerating fixes both.

**CI, and this one is quick.** Five jobs are red and they all trace to a single cause: `lockfile-check` reports `The lockfile at uv.lock needs to be updated, but --check was provided`, and `test (3.11/3.12/3.13)` plus `benchmark` all die at the *Install dependencies* step with `--locked`, in 9 to 14 seconds each, so nothing actually ran. The trigger is a good change on your part, declaring `reportlab`, `pillow` and `pypdf` in `pyproject.toml` after I flagged them as undeclared. Editing `pyproject.toml` just needs `uv lock` run and the result committed. `skill-lint` passes, so the `data_license` move landed.

abotescu20-maker:
Round 2 pushed, at `add4fca`. Point by point against your review, then something your review's method caught that you haven't seen yet.

**1. "Every gate fails closed" — the integrity tier through `main()`.** `gate_scores` no longer lets a missing `IntegrityVerdict` pass. Two cases, deliberately different: definitions were loaded but none declares the score's id → withheld, recorded reason ("never inspected ... fails closed"); `--scores` not given at all → every released percentile carries an explicit "Score integrity was not verified" caveat in `caveats`, in `note`, and now in the clinician report (status cell + a plain-language section explaining what wasn't checked and how to close the gap). Your exact scenario re-run: the documented standard command releases the breast-cancer panel at 94.2 **with** the caveat on every surface; with `--scores` and no `--genotype`, the chr1:114448389 duplicate pair withholds it. The CLI tests you asked for exist and go through `subprocess` `main()`: no `--genotype` (asserting the duplicate-position block), no `--scores` (asserting the caveat on every released percentile), and `--scores` missing one score's file (asserting fail-closed). I mutation-tested each: re-inlining `integrity[pid]` under `if genotype:` or reverting the consumer turns them red.

**2. Rebase + id provenance.** Rebased onto `main` at `5ca937c` (includes #357 and #363). The six bundled files are byte-identical to the post-#357 `gwas-prs/data` files, `demo_prs_results.json` and every SKILL.md surface use the `CLAWBIO-*` panel ids, and `load_score_definitions` takes the id from `#pgs_id` or `#clawbio_panel_id` only — a file with neither is refused with an error saying an identifier is never derived from a filename. The one intact `PGS000001` is the authentic-format harmonised fixture, which keeps its real accession header.

**3. z of exactly 0.0.** `af_shift` accepts `sd=None` and propagates `shift_sd=None`; nothing divides by a fabricated 1.0. The technical report prints `n/a (sd underivable)` in the shift column and the gate raises a caveat stating the shift in raw-score units with the reason the sd could not be derived. Tested at the unit *and* through `main()` with a results file carrying `z_score: 0.0` — the through-`main()` variant exists because reverting the fix left the unit tests green, which is your round-2 lesson repeating at a smaller scale.

**4/5. Catalog and lockfile.** Catalog regenerated (`0.4.0`, `CC0-1.0`, byte-identical to `generate_catalog.py` output on this tree); `uv lock` run and committed (`reportlab`, `pypdf` resolved), `uv lock --check` green locally.

**What the audits found beyond your list.** Before pushing I ran three independent adversarial audits against my own revision, primed with your defect class. The headline finding: **I had recommitted it.** My loader keyed curated panels by `CLAWBIO-*`, but the documented producer — `gwas-prs`, whose `prs_results.json` this skill's frontmatter names as its input — still emitted only the legacy `pgs_id` file label. The real chain would have withheld all six scores with a false "never inspected" reason, and every green test sat on prs-abstain's own regenerated fixtures. Fixed at both ends: `gwas-prs` now carries `curated_demo_panel` and `curated_panel_id` into `prs_results.json` (which I read as the durable-artefact intent of your #356 comment — happy to split into its own PR if you'd rather keep this one single-skill), `score_id()` prefers the panel id, and a chain test runs `gwas-prs --demo` and feeds its actual output through prs-abstain's CLI.

Also from the audits, all fixed: a `--scores` directory yielding no usable definitions used to release everything with a caveat claiming `--scores` was never supplied — `main()` now refuses to run it (exit 2, clean message), and loader errors no longer escape as tracebacks (a stray `README.txt` is skipped with a warning naming it, safe because unmatched ids fail closed); the note field embedded a literal `| Caveats:` pipe that split every markdown table row and made `pdf_report.py`'s row-clipping drop the caveats from the PDFs entirely — no pipe now, an invariant test guards it, and the technical report prints full notes below the table instead of truncating at 160 chars mid-sentence; the clinician translation of the new fail-closed reason no longer misreads as "too little of the score was measured".

One thing I touched on `main`'s side of the fence deliberately: four of the six curated panel headers carry `#weights=... Vassy 2014 ... against 8 and 46 loci here`, which is true only for the T2D and CAD panels — the BC/AF/BMI/PC files cite the wrong papers and counts. Since this PR's subject is provenance accuracy and the files are bundled here byte-identically, I replaced that sentence in both copies with one that's true by construction. Flagging it explicitly since it edits #357's files.

prs-abstain's suite (85 tests) is now wired into CI next to gwas-prs's, and the catalog records the chaining both ways. Suites: 85 + 106 passing, `agentskills validate` clean, both PDFs render.

manuelcorpas (changes requested):
All six items from the last round are genuinely fixed at `56b8d5c`, verified against the diff rather than the commit messages. CI is now green on all nine jobs. This round of work is unusually disciplined and I want to be specific about why, because most of it is the hard version rather than the easy one.

**The integrity tier now fails closed on every path.** `prs_abstain.py:793-807` has three branches where there was one, and the third is the one that mattered:

```python
elif integrity:
    reasons.append("Score integrity: none of the supplied scoring files declares the id ...")
else:
    warnings.append("Score integrity was not verified: no scoring files were supplied (--scores) ...")
```

Traced through `main()` for the exact scenario I raised: with `--scores`, `integrity_verdict` now runs outside `if genotype:`, so CLAWBIO-BC-77's duplicate pair at chr1:114448389 blocks for EUR_001. Without `--scores`, the caveat propagates into `caveats`, the note, the clinician table cell and a dedicated section. A caveat is the honest ceiling when there is no file to inspect. You also added two further fail-closed paths I had not asked for: exit 2 when `--scores` yields no definitions, and outright withholding when a result's id matches no supplied file. And you fixed the root cause in the docs by adding `--scores` to the standard command and to the declared inputs.

**The tests now drive `main()`.** That was the real defect last time: the guards were covered by unit tests either side of the wiring, so reverting the fix left all 72 green. Eight of the thirteen new tests shell out through the CLI without `--demo`. I traced three reverts and each goes red, and they assert the reason *text*, so a guard firing for the wrong reason is caught too.

**`sd_ref` no longer fabricates.** `:1541-1543` now propagates `None` on `z == 0` and `af_shift` returns `None` rather than `nan`, with the technical report printing `n/a (sd underivable)` and an explicit caveat naming the raw-units shift. The test fixture is well chosen: I recomputed `Sum(2*AF*w)` for CLAWBIO-T2D-8 as 1.1186, so `z_score = 0.0` with `raw_score = 1.1186` really does put that individual exactly at the mean rather than merely asserting it.

**The bundled panels are byte-identical to `skills/gwas-prs/data/`**, not merely similar, and the filename fallback is gone: `:259-264` refuses to derive an identifier from a filename and raises naming the missing header. `score_id()` prefers `curated_panel_id` over the legacy label, and `TestProducerConsumerChain` crosses the real boundary by running `gwas_prs.py --demo` and feeding its output through the gate.

**The catalog is a genuine regeneration.** I ran the current generator over your tree and diffed against your committed `skills/catalog.json`: byte-identical. The diff is scoped to `skill_count`, one chaining entry and your own block, with no collateral churn. That matters more than it looks: a hand-edited catalog is what put #329 into an unresolvable merge conflict, and you avoided it.

**`uv.lock` is correct.** `provides-extras`, the optional-dependency group, `requires-dist` with matching markers, and hashed blocks for reportlab 4.5.1 and pypdf 6.16.1, with pillow correctly not re-added since matplotlib already locked it. `lockfile-check` passes.

Also checked and still holding: no ancestry inference, no hard super-population label, no admixture, missing coordinates refuse rather than imputing to the origin, calibrated threshold, disclaimer on every PDF page, `risk_category` rendered nowhere. Every documented number reproduces from the bundled data, including EUR_001 = 1.7413, AFR_001 = 10.3770 and threshold 3.4657, which I recomputed independently.

Two things before merge, both small.

**1. The clinician PDF is the one report that omits the THRESHOLD OVERREACH warning.** `threshold_exceeds_nearest_other` fires at `:1003` (report.md), `:1246` (technical) and `:1630` (stderr). `_write_clinician_report` receives `cal` and reads it zero times.

Failure scenario: `--demo --k-sd 20` gives a threshold of 39.7 against a nearest non-EUR panel member at 7.3788, so AFR_001 (true distance 10.3770) returns `REPORT` and its percentiles are released. `report.md` and the technical report both say "Any REPORT verdict below is uncalibrated". The document headed "Summary for clinicians" prints AFR_001's traits as "Released" with no indication the abstention rule was disabled. That is the one surface where the omission costs most, and it is the same shape as the defect you just fixed elsewhere: the JSON stays honest while the human-facing document does not.

This is pre-existing rather than a round-2 regression, but it is worth closing now.

**2. The three new clinician-report rendering branches have no test.** `:1145-1152`, `:1165-1172` and `:1204-1205` only render when `--scores` is absent, and every clinician-report test drives `--demo`, which always supplies it. The underlying data is well covered at `result.json` level by `test_no_scores_means_no_silent_integrity_pass`, so the gap is narrow and specific: the JSON keeps the caveat while the PDF silently stops carrying it. One test without `--scores` closes it.

While you are there, `tests/test_prs_abstain.py:515-518` (`test_skill_still_runs_without_reportlab`) asserts only `hasattr(pa, "_write_pdfs")`, which measures nothing about the documented guarantee. Deleting the guarantee leaves it green.

**One thing needing my decision rather than a change from you.** The PR now edits `skills/gwas-prs/` (the script, a test, and four data files) and `.github/workflows/ci.yml`. I checked all of them and they are correct: the `gwas_prs.py` change adds two keys set unconditionally at the single construction site so the direct indexing cannot `KeyError`, and the data-file edits remove a sentence about Vassy and Abraham that had been copy-pasted into four panels citing neither, correctly left in place on the two where it belongs. The `ci.yml` step is benign and, I now think, load-bearing rather than redundant: it duplicates the changed-skill step on PRs, but that step does not run on push-to-main, and `TestProducerConsumerChain` shells out to gwas-prs, so a future gwas-prs-only merge would break prs-abstain and only the hardcoded step would catch it. Accepted, and flagged only so the cross-skill scope is on the record rather than implicit.

abotescu20-maker:
Both items closed at `a08bb02`, plus the nit.

**1. Threshold overreach now reaches the clinician.** `_write_clinician_report` reads `cal` and, whenever `threshold_exceeds_nearest_other`, opens the document with a plain-language "Read this first" section — placed before any Released row can be read as calibrated — saying the safeguard was effectively switched off, every Released result is uncalibrated, and to re-run with the default threshold. Your exact scenario is the test: `--demo --k-sd 20` through the CLI, asserting AFR_001's verdict really is `REPORT` (so the failure scenario is live, not hypothetical), that the warning text appears, and that it precedes the `## Results` section. A companion test asserts the default threshold prints none of it.

**2. The three no-scores clinician branches are pinned.** One CLI test runs the standard command without `--scores` and asserts all three: the marked status cell, the dedicated note section, and the remedy sentence. It also counts: the number of "score file not checked" rows in the clinician report must equal the number of released scores carrying the integrity caveat in `result.json`, so the JSON and the clinician surface cannot drift apart silently again — which was the shape of both of this round's findings.

**The reportlab nit.** `test_skill_still_runs_without_reportlab` now measures the documented guarantee instead of `hasattr`: it runs the CLI in a subprocess whose `import reportlab` is forced to fail via a PYTHONPATH shim, and asserts exit 0, markdown reports written, no PDFs. Mutation-checking it taught me something: the guarantee turns out to live at three layers (`_write_pdfs`'s import guard, `pdf_report._available`, and the per-job `except`), so deleting one or even two leaves the behavior intact and the test rightly green; deleting all three turns it red. The test guards the behavior, not any single guard's existence.

Suite is 88 passing; nothing outside `skills/prs-abstain/` changed this round.

On the cross-skill scope you accepted: agreed on the record, and thanks for tracing the `ci.yml` step to its load-bearing case — the push-to-main hole for `TestProducerConsumerChain` was exactly why it's hardcoded rather than left to the changed-skills step.

manuelcorpas (commented):
## pr-audit: advisory, no blocking findings

This is the strongest skill PR in the current queue. It is built specifically to defend the failure classes this repo keeps getting bitten by, and in several places it defends them better than the surrounding codebase does. Detail, since a review that only lists complaints is not much use:

- The skill never infers ancestry and never emits a population label. It consumes a declared population plus supplied PC coordinates and returns only `REPORT`, `REFUSE_DISTANT` or `REFUSE_UNDETERMINABLE`. Marker sufficiency is checked before distance in `decide()` at `:702-718`, missing coordinates are explicitly never imputed as the origin at `:720-730`, and abstention genuinely fires in the shipped demo, since `DEMO_PATIENT` carries `n_markers_shared=0`.
- No PRS is computed, no OR is multiplied, no prevalence applied, and `KNOWN_LIMITATIONS` states outright that absolute risk is not computed. The `ld_audit` and `effective_n_ld` machinery actively downgrades the effective number of independent contributions for physically clustered variants, which is a real guard against correlated-loci double counting.
- The four modified `gwas-prs/data/PGS*.txt` files are a correction, not a provenance downgrade. The "this is NOT a PGS Catalog scoring file" notes already existed at base; the only edit removes a boilerplate sentence about Vassy 2014 and Abraham 2016 that was factually inapplicable to those four files, whose `loci_reference` lines are Mavaddat, Tada, Locke and Schumacher. The sentence is correctly retained in the two files it does describe. `load_score_definitions` refuses to fall back to the filename at `:264`.
- `DISCLAIMER` is written into all three markdown reports, the JSON and stdout, and `NOT_REASSURANCE` ("a withheld percentile is not evidence of low risk") is emitted on every refusal. That distinction is easy to get wrong and it is right here.
- The three added CI lines are a test invocation only, with no new triggers, no `pull_request_target` and no secrets.

### Four small things

**1. `prs_abstain.py:1640-1653`. Threshold overreach warns but does not enforce.**

```python
if cal.threshold_exceeds_nearest_other:
    print(f"  WARNING: THRESHOLD OVERREACH ...", file=sys.stderr)
```

When a user raises `--k-sd`, the radius can exceed the distance to the nearest non-reference individual, at which point the gate admits other populations and, in the code's own words, is no longer an abstention rule. The condition is computed at `:72-74` and surfaced everywhere, but it never changes a verdict: `REPORT` still issues and `main()` returns 0. This is the one place where abstention is advisory rather than enforced. Consider refusing to release percentiles under overreach unless an explicit override flag is passed.

**2. `prs_abstain.py:39-41`. The Kosoy threshold is characterised slightly more strongly than the paper supports.**

The citation is correct, which is more than can be said for several past cases: PMID 18683858 is genuinely Kosoy et al., Hum Mutat 30(1):69-78. But the abstract states that 128 AIMs and subsets as small as 24 suffice for continental assignment, so the validated floor is 24, not 30. Your 30 is more conservative and fails safe, but "the lower bound validated in Kosoy" is not quite what the paper says. Suggest rewording to "above the 24-AIM subset validated in Kosoy et al. 2009".

**3. `prs_abstain.py:1389-1391`. `n_markers_shared` is never cross-checked.** Your own limitations list says it: an inflated value passes the marker gate unconditionally. Since the demo path already passes `--genotype` at `:1488`, the count could be validated against actual overlapping calls. Cheap fix, and it closes the only remaining route to the low-marker over-call class.

**4. `clawbio/cli.py:453-463` omits `accepts_genotypes` while allowing `--genotype`.** Not a bypassed guard: I traced the key repo-wide and it is never read at runtime. Cosmetic, flagged only so it is not mistaken for something worse.

### Unrelated, for a separate issue

`skills/ancestry-risk-profiler/ancestry_risk_profiler.py:54-55` cites the same Kosoy paper as "Hum Genet", which is the wrong journal. Pre-existing and not this PR's problem; this PR gets it right.

manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Thank you for this. I have read the whole diff and the skill as it would run, including the demo data, both report writers and the test suite.

## What is right

- **Identifier provenance is handled properly.** `load_score_definitions` takes the score id from `#pgs_id` or `#clawbio_panel_id` and raises rather than deriving one from the filename (`skills/prs-abstain/prs_abstain.py:260-265`). The bundled panels declare `#clawbio_panel_id=CLAWBIO-BC-77` with an explicit note that the file is not the PGS Catalog accession (`skills/prs-abstain/examples/scores/PGS000001_hmPOS_GRCh37.txt:2,14-19`). Four of the six bundled files are byte-identical blobs to `skills/gwas-prs/data/`, so the "byte-for-byte copy" claim in `SKILL.md` is verifiable from the diff itself.
- **The parser is tested against vocabulary the source really emits**, not only the bundled dialect. `skills/prs-abstain/tests/fixtures/PGS000001_hmPOS_GRCh37_authentic.txt:1-20` carries the genuine harmonised header block and the `hm_rsID`, `hm_chr`, `hm_pos`, `hm_inferOtherAllele` columns, and `test_authentic_pgs_catalog_file_parses_by_header_name` (`skills/prs-abstain/tests/test_prs_abstain.py:592`) pins resolution by header name, full-precision weights, `af_reference is None`, and `expected_mean` returning None rather than crashing. That is the check most often missing on this repository.
- **Fail-closed behaviour is consistent.** An absent `reference_population` withholds instead of defaulting to EUR (`prs_abstain.py:838`), duplicate genomic positions block rather than warn (`prs_abstain.py:492`), an unusable `--scores` directory exits 2 instead of releasing with a false caveat (`prs_abstain.py:1559`), and threshold overreach is a hard error unless explicitly overridden (`prs_abstain.py:1542`). Marker sufficiency genuinely precedes distance (`prs_abstain.py:711`), and missing coordinates never become the origin.
- **Citations check out.** Kosoy et al. 2009, Hum Mutat 30:69-78 is PMID 18683858; Martin 2019 is 30926966; Duncan 2019 is 31346163 (`skills/prs-abstain/SKILL.md:427-433`). No network calls appear anywhere in `prs_abstain.py` or `pdf_report.py`, and the bundled allele-frequency table is labelled synthetic in the file itself (`skills/prs-abstain/examples/demo_population_af.tsv:1-2`).
- **The `#weights=` edit to the four gwas-prs panels is a real provenance fix.** The removed sentence named Vassy 2014 and Abraham 2016, which describe the T2D and CAD panels, not breast cancer, atrial fibrillation, BMI or prostate (`skills/gwas-prs/data/PGS000001_hmPOS_GRCh37.txt:9`). The sentence correctly survives in the T2D file.

## What needs an answer before merge

1. **One `prs_results.json` is applied to every individual.** `main()` gates the same `scores` list against each row of the individuals CSV (`prs_abstain.py:1604-1608`) and then writes one row per sample_id and pgs_id into `tables/gated_scores.csv` (`prs_abstain.py:1621`). `gwas-prs` emits a flat list of score records with no sample identifier, so there is no join key, and nothing warns. The symptom is visible in your own documentation: `SKILL.md:255` shows AFR_001 carrying raw score 0.8000, which is the demo patient's score, and a released percentile for a passing individual would come from the same file. For a skill whose stated purpose is to stop a number being attached to a person it does not describe, silent cross-attribution should be a refusal. The single `--genotype` assumption is disclosed at `prs_abstain.py:1415-1419`; this one is not disclosed at all. Please either require a sample identifier in the results file, or refuse when the individuals CSV carries more than one row.

2. **The allele-frequency caveat is asserted unconditionally.** `gate_scores` appends "Re-centring on AFR allele frequencies would move the reference mean by X sd, i.e. this percentile is off by roughly that much before any individual-level error" to every score carrying a shift (`prs_abstain.py:857-861`), including released percentiles for the EUR individual who passed the gate. For that individual the sentence is not true: the AFR shift is a property of the score under a hypothetical re-centring, not an error in their percentile. The direction of the error is conservative, so this is wording rather than an unsafe release, but it should be conditioned on the individual's population or reworded.

3. **Sign inversion in the module header comment.** `prs_abstain.py:98` says the mean moves by `Sum(2 * (AF_ref - AF_pop) * w)`, while the implementation (`prs_abstain.py:558`), `SKILL.md` and the technical report all use `AF_pop - AF_ref`. The code is right, the comment is not.

4. **Documented demo does not match shipped demo data.** `SKILL.md:255-257` and the unit tests use 480 shared markers, while `examples/demo_query_individuals.csv:2-3` declares 400. The demo genotype file holds roughly 415 calls, so 480 would now trip the new credibility cap you added. Please align the worked example with the data that ships.

5. **Nothing has been executed.** CI is awaiting maintainer approval and no check has run on this head, so every claim about demo output is unverified. That matters more than usual here because `skills/gwas-prs/gwas_prs.py:1515` reads `r["curated_demo_panel"]` and `r["curated_panel_id"]` directly, and those keys are produced upstream of this diff. `skills/gwas-prs/tests/test_score_provenance.py:349` and `TestProducerConsumerChain` (`skills/prs-abstain/tests/test_prs_abstain.py:885`) would both catch a mismatch, but only once the workflow runs.

## Scanner findings re-checked

- **ci_signal WARNING**: correct. `statusCheckRollup` is empty and the run is pending approval.
- **code_quality INFO, `pdf_report.py` has no corresponding test file**: literally true, misleading as a coverage claim. `TestPDF` (`skills/prs-abstain/tests/test_prs_abstain.py:489-546`) renders both PDFs, parses them with pypdf, asserts page count and that the key sentence survives extraction, and forces an ImportError to prove the run still exits 0 with the markdown written.
- **dangerous_calls INFO, four findings**: all are `subprocess.run([sys.executable, <the skill's own script>])` inside tests. Not findings.
- **No prompt injection.** The imperative "The model will want to X. Do not." lines in the SKILL.md Gotchas address the dispatching agent and only tighten its behaviour. Nothing in the diff addresses a reviewer or claims approval.

No secrets, no telemetry, no variant or phenotype content leaving the machine. The scientific framing of transferability, LD attenuation and effective_n is sound and honestly bounded in `KNOWN_LIMITATIONS` (`prs_abstain.py:1411`). Point 1 is the one I would want fixed before this merges.

---

manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Strong work, and the domain read holds up. The fail-closed direction is consistent throughout: missing PC coordinates never become the origin (`prs_abstain.py`, `decide`), an absent trait refuses rather than assumes (`check_applicability`), and a score id that no supplied file declares is withheld rather than passed (`gate_scores`). I checked the vocabulary rather than trusting it: `"estimated (allele freq)"` is a string `skills/gwas-prs/gwas_prs.py:676` really emits, and `tests/fixtures/PGS000001_hmPOS_GRCh37_authentic.txt` carries genuine PGS Catalog `hm_*` columns whose weights are true ln(OR). Kosoy 2009 (PMID 18683858), Martin 2019 and Duncan 2019 all check out, and admixture is named as the unhandled case on every surface rather than papered over.

Two things before this can go in.

1. In `skills/gwas-prs/gwas_prs.py` the new lines index `r["curated_demo_panel"]` and `r["curated_panel_id"]` directly. Please confirm every result record reaches that point carrying both keys, including the non-curated `--pgs-id` path; the new test in `skills/gwas-prs/tests/test_score_provenance.py` drives only `--demo`. A `.get()` would settle the question outright.

2. CI has never run here. The workflow is awaiting maintainer approval, so 5,600 added lines carry no signal at all. That approval is on us, not you.

**Non-blocking suggestions**

The scanner note that `pdf_report.py` has no test file is a false positive: `TestPDF` in `tests/test_prs_abstain.py` covers it, including the missing-reportlab path.

---

manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Reviewed the whole diff, not just the new skill.

What is right, specifically: the gate order (markers, then placeability, then distance) is enforced in `decide()` and pinned by a test; a missing PC never becomes 0.0; `--scores` supplied but unusable exits 2 rather than releasing with a false caveat; and `TestProducerConsumerChain` feeds real `gwas-prs` output through the gate instead of a hand written fixture, which is the only way `estimated (allele freq)` and the panel ids get checked against the tool that actually emits them. Ids come from the file header, never the filename. Citations hold: Kosoy 18683858, Martin 30926966, Duncan 31346163, and the PMIDs on the bundled panels. Dropping the Vassy/Abraham sentence from the four panels that cite Mavaddat, Tada, Locke and Schumacher is a correction, not a weakened caveat.

Merging. The points below are for later.

**Non-blocking suggestions**

1. `prs_abstain.py:1157` states unconditionally that "Every score reviewed here was built and calibrated in people of European ancestry". Under `--ref-pop` anything other than EUR that is false, in the clinician document. Derive it from `cal.reference_population`.
2. `prs_abstain.py:1299` maps every "Score integrity" reason to "too little of the score was measured", so a duplicate position block (the demo's CLAWBIO-BC-77) reaches the clinician with the wrong reason. A branch on "more than one scored variant" would fix it.
3. The scan note "No corresponding test file found" for `pdf_report.py` is a false positive: `TestPDF` covers rendering, PDF validity and the reportlab absent path.

CI has not run yet on this fork, so it will need approval and a green run before merge.

---

manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Thanks, this is careful work. I rechecked the demo arithmetic by hand: the EUR centroid over PC1 to PC4 puts EUR_001 at 1.74 and the nearest non-EUR individual (EAS_004) at 7.38, so the 3.47 threshold really does sit in an empty gap and the demo cannot overreach silently. The z scores in `examples/demo_prs_results.json` reproduce from the gwas-prs curated means and sds, `EUR` and `estimated (allele freq)` are strings gwas-prs actually emits (`gwas_prs.py:649,676`), the fixture column names are genuine PGS Catalog harmonised ones, and PMIDs 18683858, 30926966 and 31346163 all check out. Nothing blocks this on the science.

**Non-blocking suggestions**

1. `prs_abstain.py:4404` and the `audits` dict: one `--genotype` is applied to every queried individual, so with several rows in the individuals CSV the weight coverage printed per person in `report_technical.md` belongs to whoever owns the genotype. You already refuse this cross-attribution for scores; consider refusing `--genotype` on multi-individual runs, or binding it to a `sample_id`.
2. `prs_abstain.py:4395`: `sd_ref` inverts `z` against `expected_mean`, which assumes `allelefrequency_effect` is the population the percentile was centred on. True for the curated panels, and it fails closed to `None` on most real files, so a note only.
3. This branch sits on the unmerged #356/#357 work: on main, `skills/gwas-prs/tests/test_score_provenance.py` does not exist and nothing sets `curated_panel_id`, so the patch will not apply and the producer-consumer test cannot pass. It needs a rebase before it can merge; GitHub already flags the conflict.

Scanner: its warning (no CI on this head) is accurate. The "no test file for `pdf_report.py`" note is a false positive; `TestPDF` exercises it end to end.

---

Strong, careful work. The fail-closed defaults are consistent throughout: missing PCs are never imputed as the origin, the marker check runs before distance (`prs_abstain.py:746`), an absent `reference_population` withholds rather than releases, and threshold overreach is a hard error unless explicitly overridden. The citations check out (Kosoy 2009 PMID 18683858, Martin 2019, Duncan 2019), `estimated (allele freq)` really is a string gwas-prs emits (`gwas_prs.py:676`), and the score loader keys on the header instead of the filename. The tests read as behaviour tests, not as mirrors of the implementation.

Two things I would like you to confirm before this merges.

1. **One `--genotype` speaks for everyone.** Weight coverage (`prs_abstain.py:1678`) and the marker cap (`:1700`) come from a single genotype file and then gate every row of the individuals CSV, so EUR_001's percentile can be released on the strength of DEMO_PATIENT's DNA. The bundled `--demo` runs exactly that shape with three people. The results file already refuses this without `sample_id` (`:1724`); could `--genotype` follow the same rule?

2. **Reference sd provenance.** `sd_ref = abs((raw - expected_mean) / z)` (`:1691`) assumes the reference mean is Sum(2*AF*w). That holds for the curated panels, but in an authentic PGS Catalog file `allelefrequency_effect` is the source-GWAS frequency, so `shift_sd` can be mis-scaled while printing as a definite "+1.28 sd" in both reports.

**Non-blocking suggestions**
- `examples/scores/` keeps PGS-Catalog-style filenames while the headers say `CLAWBIO-*`. The loader is header-driven so nothing is wrong, but CLAWBIO-named files would remove the trap.
- No CI has run on this head and the branch conflicts with main; a rebase would let the producer/consumer chain test actually run.

---

manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Thanks, this is unusually careful work. Three things I checked specifically and found correct:

1. Vocabulary against the real producer. `estimated (allele freq)` matched in `prs_abstain.py` is exactly the literal `gwas_prs.py` emits, and the PGS Catalog column fallbacks (`hm_rsID`, `hm_pos`, `allelefrequency_effect`) are real harmonised names, exercised against a genuine PGS000001 fixture.
2. Every gate fails closed: absent trait, absent `reference_population`, unmatched score id, unusable `--scores` directory, unbound genotype and unkeyed results across several individuals all refuse rather than release. Missing PCs never become the origin.
3. Citations check out. Kosoy 2009 PMID 18683858, Martin 2019 PMID 30926966 and Duncan 2019 PMID 31346163 are all real and correctly attributed, and the header edits in `skills/gwas-prs/data/` remove a Vassy/Abraham note from panels those papers do not source. Demo AF and genotype data are labelled synthetic, and there are no network calls anywhere.

The scanner's only warning is that CI awaits maintainer approval; that is accurate, not a false positive. Merging still needs green CI.

### Non-blocking suggestions

Merging on these grounds; both are for later.

- `clawbio/cli.py:461` omits `--genotype-sample-id` from `allowed_extra_flags`, so a multi-individual genotype run via the wrapper can only refuse. Fails closed, but worth adding.
- `prs_abstain.py:744`: with a single-population panel `nearest_other` is None, so the overreach check silently cannot run. Elsewhere you record unrunnable checks as caveats; a note here would match.

---

manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Thanks for this. I checked the arithmetic rather than taking the prose on trust: the demo numbers reproduce from the bundled panel (EUR_001 1.74, AFR_001 10.38, nearest non-EUR 7.38, within-EUR max 3.14), so the "empirically empty gap" is measured, not asserted. Every tier fails closed, the raw score is kept while the percentile is withheld, and there are no network calls. The literals match their sources too: `hm_rsID`, `allelefrequency_effect` and `#pgs_id` are genuine PGS Catalog fields, and `estimated (allele freq)` is a string gwas-prs really emits (gwas_prs.py:784). Kosoy PMID 18683858, Martin 30926966 and Duncan 31346163 all check out. This is merging; the points below are for later.

**Non-blocking suggestions**

1. Main already sets `curated_demo_panel` and `curated_panel_id` in the same `jr` literal (gwas_prs.py:1594-1595). Your added block repeats both, so the merged dict carries each key twice. Drop them on rebase.

2. prs_abstain.py:672: matching the frequency table's allele by strand complement is right for A/G, but for a palindromic A/T or C/G variant it accepts the other allele and flips the sign of that variant's `delta_mean`. Consider skipping palindromic variants on that path. It moves a caveat figure only, never a released percentile.

The scanner's one warning is that CI has not run on this fork, which is accurate and unrelated to the code. Its note that `pdf_report.py` has no tests is covered by `tests/test_prs_abstain.py::TestPDF`.

---

---

manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Strong piece of work, and the vocabulary holds up against the real sources rather than against itself. `estimated (allele freq)` really is emitted by gwas-prs (gwas_prs.py:784), `curated_panel_id` and `#clawbio_panel_id` are genuine, and the PGS Catalog column chain in prs_abstain.py:2800 (`hm_rsID`, `hm_inferOtherAllele`, `allelefrequency_effect`) matches the published format. The fixture `tests/fixtures/PGS000001_hmPOS_GRCh37_authentic.txt` is real-format and its weights reconcile with its ORs, so the parser test proves something. Kosoy 18683858, Martin 30926966 and Duncan 31346163 all resolve and match the claims made of them.

The header edit in `skills/gwas-prs/data/` is a real correction, not cosmetic: "8 and 46 loci" only ever described CLAWBIO-T2D-8 and CLAWBIO-CAD-46 and had been copy-pasted into the other four panels.

Every tier fails closed, a missing PC never becomes the origin, and `test_gwas_prs_demo_output_flows_through_the_gate` crosses the producer/consumer boundary instead of testing the parser against its own output. Nothing here can release a percentile it should withhold.

### Non-blocking suggestions

These are for later; the verdict is merging.

1. prs_abstain.py:1038 calls `score_pop.lower()`, but `load_prs_results` type-checks only the numeric fields. A non-string `reference_population` would raise after some outputs are already written. Worth adding to the same validation loop.
2. skills/catalog.json:3529 carries `"status": "planned"` beside `maturity_tier: ci-validated` and `has_tests: true`. Likely a regeneration artefact.
3. For the record, the scanner's "no corresponding test file" note on `pdf_report.py` is a false positive: the PDF path is covered by TestPdf in test_prs_abstain.py, including a run with reportlab absent.

---

manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Careful work, and the parts that usually go wrong here are right. The PGS Catalog column names are real (`hm_rsID`, `hm_chr`, `hm_pos`, `hm_inferOtherAllele`, `allelefrequency_effect`), and `tests/fixtures/PGS000001_hmPOS_GRCh37_authentic.txt` tests the parser against a genuine harmonised file rather than a fixture shaped to fit it. `estimated (allele freq)` and `curated_panel_id` match what `gwas-prs/gwas_prs.py` really emits. The `Sum(2*AF*w)` identity in SKILL.md checks out against `gwas-prs/curated_scores.json` (1.12, 2.84, 7.11), and the example distances reproduce from `examples/demo_reference_pcs.csv`. Missing coordinates, non-finite distances and absent provenance all fail closed, and there are no network calls. No blocking findings; the notes below are for later.

**Non-blocking suggestions**

1. `prs_abstain.py:496` drops score variants whose `chr`/`pos` will not parse, without counting them; every other skipped-input path in this file reports a count on stderr. `clustered_weight_share` also keeps the full weight denominator, so a scoring file with partial coordinates under-warns on LD clustering.
2. The scanner INFO "no corresponding test file" for `pdf_report.py` is a false positive: `TestPDF` in `tests/test_prs_abstain.py` covers the rendered PDF, its key sentence and the no-reportlab path.
3. Practical blocker rather than a review point: the branch is conflicting and no CI has run on 9808645. A rebase onto main would let the pipeline verify it.

---

manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Thorough work, and the fail-closed direction is right throughout: an absent PC coordinate returns `REFUSE_UNDETERMINABLE` rather than the origin (`prs_abstain.py:964`), markers are checked before distance, and threshold overreach is a hard error unless explicitly overridden (`prs_abstain.py:1925`).

Vocabulary checks out, which is where this area usually breaks. `estimated (allele freq)` is exactly the string at `skills/gwas-prs/gwas_prs.py:784`, and `curated_panel_id`, `reference_population` and `z_score` are real fields. `TestProducerConsumerChain` runs `gwas_prs.py` itself rather than a hand-written fixture, so the contract is tested against the source, not your own spelling.

Citations too: Kosoy 2009 (PMID 18683858), Martin 2019 (30926966) and Duncan 2019 (31346163) are real and accurately described. Dropping the Vassy/Abraham sentence from the four panels that do not cite them, while leaving it in CAD-46 and T2D-8, is a genuine fix.

Merging. Two points for later.

**Non-blocking suggestions**

1. `load_population_af` (`prs_abstain.py:316`): a three-column table carries no `_allele`, so the shift assumes the frequency counts the effect allele. A gnomAD export giving ALT frequency where the effect allele is REF would yield a wrong `af_shift_sd`. It gates nothing, but a stderr warning when the allele column is absent would close it.
2. The scanner's "no test file" note on `pdf_report.py` is a false positive: `TestPDF` renders both PDFs and parses them with pypdf. Its CI warning is valid though; checks have not run yet.

---

manuelcorpas:
Full re-read at head a5af8667 against every point from the three rounds and the audit comments since. All twelve are addressed, and I have the line for each: overreach is a hard error at `prs_abstain.py:1925`, the three no-scores clinician branches are tested at `test_prs_abstain.py:1212`, the reportlab shim at 605, integrity fails closed at 1051 to 1065, `z == 0` returns no sd at 662, filename identifiers are refused at 284, the one-results-for-everyone and one-genotype-for-everyone cases exit 2 at 2038 and 1983, the Kosoy wording, the `reference_population` type check at 833, and the `ld_audit` unplaced count at 500. Also checked: the PGS000001 fixture is byte-identical to the EBI FTP file, the six example score files are byte-identical to `skills/gwas-prs/data/`, the pdf module has no network and no subprocess, and no ancestry label or admixture is ever emitted (below 30 markers it refuses at 927 to 944).

I have approved the CI run for this head. It had never executed, so nothing above was machine-verified until now.

**One open question before merge**

Genome build never refuses. `prs_abstain.py:291` stores the score header's `genome_build` and no line reads it back. A GRCh38 genotype against a GRCh37 score passes silently, and the only disclosure is the `KNOWN_LIMITATIONS` line at 1803. Either check it against a `--build` argument and refuse on mismatch, or keep it as a documented limitation and say so in the report the clinician sees, not only in the technical one. Tell me which you prefer.

**Non-blocking**

- `skills/catalog.json` has `"status": "planned"` beside `"maturity_tier": "ci-validated"`.
- `main()` creates `tables/` and `reproducibility/` at 1912 before any `return 2`, so a refusal leaves a partial tree.
- `load_population_af` at 351 says nothing when the AF table has no allele column.

Green CI on this SHA plus a decision on the build check, and this merges.

manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Thank you, this is unusually careful work and I checked the numbers rather than trusting them.

What holds up: `Sum(2*AF*w)` for CLAWBIO-T2D-8 comes to 1.1186, exactly as `SKILL.md` claims; the demo distances 1.74 and 10.38 and the nearest non-EUR individual at 7.38 all reproduce from `examples/demo_reference_pcs.csv`. Kosoy 2009 (18683858), Martin 2019 (30926966) and Duncan 2019 (31346163) are all real and correctly described. The PGS Catalog vocabulary in `prs_abstain.py:181` (`hm_rsID`, `chr_name`, `allelefrequency_effect`, `#pgs_id`) is what the Catalog really emits, and it is tested against a genuine PGS000001 fixture rather than one written to match the parser. Marker sufficiency before distance, missing PCs never imputed as origin, and the refusal to derive `shift_sd` without established provenance are the right calls. Merging.

### Non-blocking suggestions

1. `prs_abstain.py:358` warns on stderr when the allele-frequency table names no allele, but `af_shift` then trusts the orientation silently for any non-conflicting rsID and `prs_abstain.py:766` adds no note. A gnomAD export counting the other allele would flip the sign of the printed shift. Worth carrying that stderr warning into `sd_note`.
2. `prs_abstain.py:291` reads `#genome_build`, but harmonised Catalog files keep the original there (`NR` in your own fixture) and declare the coordinate build in `#HmPOS_build`. Since you prefer `hm_pos`, `build_check` at `prs_abstain.py:495` checks a different build from the one it uses. Impact is confined to LD clustering, as matching is by rsID.
3. Scanner note that `pdf_report.py` has no test file is not right: `TestPDF` exercises it end to end.

CI has not run yet (awaiting approval on a first-time fork), so this still needs a green run before it lands.

---

manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Thanks, this is an unusually careful skill. What I checked and found right:

- The decision order in `prs_abstain.py` (markers, then placeability, then distance) fails closed at each step, and a missing PC is never imputed as the origin.
- The scoring-file vocabulary is real: `hm_rsID`/`hm_pos` preferred over the submitted columns, `HmPOS_build` used instead of `genome_build` when harmonised coordinates are in play, and `estimated (allele freq)` is a string `gwas-prs` actually emits (`skills/gwas-prs/gwas_prs.py:784`).
- The `sample_id` join, plus the refusal to gate one results file against several people, closes the cross-attribution path.
- Citations check out (Kosoy PMID 18683858, Martin PMID 30926966, Duncan PMID 31346163), and dropping the Vassy/Abraham sentence from the BC, BMI, PC and AF panel headers fixes a real copy-paste error.

This is merging. The points below are for later.

**Non-blocking suggestions**

1. `af_shift`'s `sd_note` only reaches a report when `shift_sd` is None (`skills/prs-abstain/prs_abstain.py:1153`, `:1720`). With a frequency table that names no allele (`:383`, including the bundled demo table) a derivable sd prints a signed "+x.xx sd" while the orientation caveat stays on stderr. Appending `sd_note` on both branches would keep the warning next to the number, as the comment there intends.
2. The static scanner's only warning is that CI has not run yet; its "no test file for pdf_report.py" note is stale, since the PDF path is covered end to end in `tests/test_prs_abstain.py`.
3. `prs_abstain.py:322` carries a Romanian docstring; English would match the rest of the file.

---

Thanks for this: it is unusually careful work. The fail-closed ordering in `decide()`, refusing to impute a missing PC as the origin, and the sample_id join that refuses cross-attribution are all right. I checked the literals against the producers: `curated_panel_id`, `"estimated (allele freq)"`, `#clawbio_panel_id`, `#HmPOS_build` and the `hm_*` columns are strings those sources really emit, and `examples/scores/` is byte for byte the updated `skills/gwas-prs/data/` files. Kosoy 2009, Martin 2019 and Duncan 2019 all resolve.

One thing to settle before merge.

**The orientation caveat never reaches a report when the shift is quantified.** `prs_abstain.py:1159` and `:1720` surface `sd_note` only when `shift_sd is None`. `examples/demo_population_af.tsv` has no allele column, so `orientation_unverified` is set at `prs_abstain.py:806`; the curated panels do have a derivable sd, so the demo prints a signed "+X sd" shift with the "would reverse the sign" sentence dropped. Your own comment at `prs_abstain.py:380` says the warning must travel with the number. `tests/test_prs_abstain.py:1595` asserts only that the string is on the dataclass field, and it passes sd=1.0, which is exactly the case where it is dropped.

**Non-blocking suggestions**
- The `AFTable` docstring at `prs_abstain.py:322` is in Romanian; the rest of the file is English.
- CI has not run here at all (fork PR awaiting approval), so none of the 1,785 test lines have executed yet. The absent checks are absent, not green.

---

manuelcorpas:
Delta read, a5af8667 to 8c74f978. The genome-build question from the 8th is resolved: score side reads `HmPOS_build` before `genome_build` (`prs_abstain.py:295`), genotype side is the explicit `--genotype-build` argument (`:1945`), mismatch prints "Genome build mismatch" and returns 2 (`:2053-2058`), unknown is disclosed in the clinician report (`:1571-1577`) and in `KNOWN_LIMITATIONS` (`:1875`). Each branch has a test that goes red if its guard is deleted (`test_prs_abstain.py:462` to `:504`, `:1629`). The `.bak` removal commit is deletions only, and moving the `tables/` and `reproducibility/` mkdir to `:2189` closes the partial-tree note. Thank you, and I have re-approved CI for this head.

Two things remain, then this merges.

1. **`NR` becomes a mismatch, not unverified.** `norm()` at `:535` returns any unknown string unchanged, so a non-harmonised Catalog file with `#genome_build=NR` and no `HmPOS_build`, run with `--genotype-build GRCh37`, exits 2 with "is NR, genotype is GRCh37" (`:546`). NR means not reported and belongs in the unverified branch. No test covers it.
2. **Orientation caveat still detaches from the number.** `:1159` and `:1720` gate `sd_note` on `shift_sd is None`. The bundled `examples/demo_population_af.tsv` has no allele column, so `orientation_unverified` fires at `:383` and the demo prints a signed shift with the "reverse the sign" sentence on stderr only. Your own comment at `:380` says the warning has to travel with the number. Append `sd_note` on both branches; `:1595` currently tests exactly the dropped branch with `sd=1.0`.

Then rebase: the branch conflicts on `skills/catalog.json` (#329 and #390 merged since) and `.github/workflows/ci.yml` (both sides add a step after "Run GWAS PRS tests"; keep both). Regenerate the catalog rather than hand-editing it.

Also, the `AFTable` docstring at `:322-327` is in Romanian. Fine by me, but the next reader may not agree.

For the record, the automated comments on the 9th and 10th both say "merging". They were wrong to: the branch was conflicting and CI had never run on it. That is a bug on our side, not a judgement on your work.

manuelcorpas:
Status check at head 8c74f978, nothing new from your side needed beyond what is below.

1. **This branch now conflicts with main** in `.github/workflows/ci.yml`, `pyproject.toml` and `uv.lock`. Main has moved about 186 commits since the merge base. Please merge or rebase on main, then regenerate `uv.lock` rather than hand-resolving it.
2. **CI was red on this head** because `skills/catalog.json` was not regenerated for the new skill: `tests/test_generate_catalog.py::test_checked_in_catalog_is_current` saw 97 against 98. After the rebase, run `python scripts/generate_catalog.py`, and update the skill counts in `README.md`, `llms.txt`, `CITATION.cff`, `.zenodo.json` and the two `.claude-plugin` JSON files, which `tests/test_public_claims.py` checks.
3. The two items from 10 September are still open at `prs_abstain.py`: `norm()` (`:528-535`) returns `NR` unchanged, so `#genome_build=NR` is reported as a build mismatch instead of unverified; and `sd_note` is still attached only when `shift_sd is None` (`:1159`, `:1720`).

manuelcorpas (commented):
**Verdict: NEEDS REVIEW** — mechanically blocked, science is sound (audited at 8c74f978)

**First, a correction to the record.** The 17 September note on this PR says `catalog.json` was not regenerated for the new skill. That is wrong and the contributor is owed the correction. Checking the trees directly:

| Ref | `skills/` dirs | catalog entries |
|---|---|---|
| merge base `1bd402ef` | 97 | 96 |
| this head `8c74f978` | 98 | 97 |
| `main` | 98 | 97 |

The constant off-by-one is `EXCLUDED_FOLDERS`. This head's 97-entry catalog is exactly what the generator produces for this head's tree, so the catalog **was** regenerated. The `assert 97 == 98` failure is stale-branch drift: `claw-amplicon-qc` landed on main after the 7 September merge base, and CI tests the merge. The fix is still on your side (rebase, then regenerate) but the step was not omitted.

**The science is the strongest in the current queue.** Abstention is enforced rather than advisory, the threshold is calibrated from within-reference spread rather than asserted, and marker sufficiency is checked before distance. I recomputed the gate from the bundled panel independently:

```
n_EUR 22  mean 2.0142  sd 0.4838  threshold(k=3) 3.4657
within_max 3.1356   nearest_other 7.3788
EUR_001  dist 1.7413  -> REPORT
AFR_001  dist 10.3770 -> REFUSE_DISTANT
```

The threshold sits inside a genuinely empty gap. No ancestry is inferred, no admixture proportions are emitted, missing PC coordinates are refused rather than treated as the origin, and there is no nearest-population fallback anywhere in the gate. No odds ratios, no prevalence, no absolute-risk conversion; a percentile is released only when a reference population is declared. No `np.random`, no network surface, no hardcoded effect coefficients, and every bundled input is labelled synthetic. All four items from the 26 August review are fixed, including the overreach flag now hard-erroring and the Kosoy floor corrected to 24.

Three things to resolve:

**Blocking, mechanical.** The branch conflicts in `ci.yml`, `pyproject.toml` and `uv.lock` and is 27 commits behind. Rebase, then regenerate the catalog and lockfile.

**WARNING** `norm()` returns an unrecognised build verbatim, so PGS Catalog's `NR` ("not reported") lands in `mismatched` and the refusal message asserts a build conflict that was never established. Harmonised files are handled correctly and the direction of error is safe, but `NR` belongs in `unverified`.

**WARNING, third time raised.** The orientation caveat detaches from the number: `sd_note` is appended only when `shift_sd is None`, so when the sd is derivable the report prints a signed shift while "reverse the sign" stays on stderr, which the code's own comment says must not happen. Worse, the test for it calls `af_shift` with `sd=1.0`, so it exercises the branch that never reaches a report and stays green with both append sites deleted. Please append on both branches and drive the gate rather than the helper.

Minor: the `AFTable` docstring is still in Romanian, and `examples/demo_prs_results.json` carries `pgs_id: PGS000013` (a real Khera CAD score) beside a T2D panel id. The resolved id is what reaches reports, so nothing leaks, but the latent wrong-trait label is worth clearing.

abotescu20-maker:
Thanks Manuel — all four points from R16/R17 are in, plus the merge with main. Pushed as `8864e32` (fixes) and `14957a6` (merge).

**1. `sd_note` on both branches, and the test now drives the gate.**

You were right three times over, and the third time you named the exact mechanism: the test called `af_shift(..., sd=1.0)` and asserted on the helper's `sd_note`, which is populated on every branch, so it stayed green with both append sites deleted. It was testing the helper, not the behaviour.

The fix itself: `sd_note` reached the clinician only when the reference sd could *not* be derived. The derivable branch is precisely the one that prints a signed sd-unit shift, and an unverified table orientation reverses that sign, so the caveat now travels with the number on both branches. In the technical report it is printed whether or not the sd was derivable, for the same reason.

The test is rewritten to run `gate_scores` on both branches and assert on the caveat text the clinician actually reads. I checked it the way you would: deleted both append sites, the test goes red; restored them, green.

**2. `build_check` was refusing on a not-reported build.**

`"NR"` was compared verbatim, so it differed from any declared build and a file stating *no* build was refused with a message asserting a build conflict that had never been established. `NR`, `NA`, `n/a`, `none`, `null`, `unknown`, `unspecified` and `.` now map to `None` and reach the unverified branch, on either side of the comparison. New test, also mutation-checked.

**3. `PGS000013` on the T2D demo entries.**

Correct on both counts — `score_id()` resolves to `curated_panel_id` so nothing leaked into a report, and the latent wrong-trait label was still worth clearing. The legacy labels in `examples/demo_prs_results.json` are now the panel file stems, which is what that field is: a file label, not an accession. Verified on generated reports — only `CLAWBIO-*` ids appear. The remaining `PGS000001` is the authentic fixture and is correctly labelled.

**4.** `AFTable`'s docstring was left in Romanian. Translated.

**Merge with main.** The PR had gone conflicting. I merged rather than rebased, to avoid replaying the same three conflicts across 28 commits and to keep your review threads anchored.

- `ci.yml`: both steps kept, `wgs-prs` then `prs-abstain`.
- `pyproject.toml`: main's new extras and the `[tool.uv]` liknorm override untouched; the `prs-abstain` extra sits with the other extras.
- `uv.lock`: taken from main and regenerated with `uv lock` rather than merged by hand — it adds `reportlab` and `pillow` under the extra and `pypdf` under dev, nothing else.
- `skills/catalog.json` regenerated: 98 skills, 93 with demo data, 51 CLI-registered. The counts that quote the catalog follow it in `README.md`, `llms.txt`, `CITATION.cff`, `.zenodo.json` and the two `.claude-plugin` manifests.

One thing I deliberately left alone: the test totals already disagree across the repo (5,040 in `README.md`, 4,730 in `CITATION.cff`, `.zenodo.json` and `llms.txt`). Reconciling them is your call, not this PR's, so I bumped only the skill counts, which are derived from the catalog.

prs-abstain is 157/157. Repo `tests/` is 328 passing here; the 4 failures under `tests/benchmark/` are the `sushie` extra missing from my local environment, unrelated to this branch.

abotescu20-maker:
Good catch by CodeQL, and it was more than a lint: that branch dropped an unrenderable figure from the PDF with no trace, so anyone comparing the PDF against the markdown would find a figure missing and nothing saying why.

Narrowed to the three failures the branch can actually produce (`ImportError` when Pillow is absent, `OSError` on an unreadable or truncated image, `ValueError` on a zero-width image, which the aspect-ratio division would hit) and it now warns on stderr, naming the file and saying the markdown still references it. Test added and mutation-checked against the old `except Exception: pass`. 158 pass.

PR #350, opened 2026-08-18
refactor: nutrigx repro bundle delegates to clawbio.common

camlloyd:
## Summary

- Replaces nutrigx's hand-rolled reproducibility bundle (own `sha256_file`, inline `commands.sh`/`environment.yml`/`checksums.txt` writers) with the shared `clawbio.common.reproducibility` + `clawbio.common.checksums` layer already used by 12 other skills
- Bundle artefacts now follow the ClawBio-wide convention: written to `<output_dir>/reproducibility/` with a sha256sum-compatible `checksums.sha256`; `provenance.json` keeps its schema and moves into the same directory
- Updates SKILL.md and the report footer to reflect the new bundle location

## Behaviour changes (beyond the refactor)

Called out per review:

- **Checksum manifest is outputs-only and anchored.** `checksums.sha256` now lists `nutrigx_report.md` labelled relative to the output dir (`write_checksums(..., anchor=output_dir)`), so `cd <output_dir> && sha256sum -c reproducibility/checksums.sha256` passes. The `commands.sh` verify step was updated to that form. The input file and SNP panel — whose labels could never resolve from the manifest's location — are attested as `input_sha256`/`panel_sha256` fields in `provenance.json` instead.
- **A missing report aborts the bundle instead of being silently omitted.** The old `FILE_NOT_FOUND` tombstone is gone, but rather than inheriting the shared layer's silent skip, `create_reproducibility_bundle` now raises `FileNotFoundError` if `nutrigx_report.md` is absent, so a bundle can never validate while the artefact it attests is missing.
- **`environment.yml` channels are now `conda-forge` only.** The `defaults` channel from the old hardcoded literal is intentionally dropped: all four conda deps resolve from conda-forge, and this avoids Anaconda's `defaults` ToS. This matches the shared helper's default.
- **`provenance.json` routes through `write_text_lf`** like its three siblings (LF on every OS), and gains the two input-attestation fields above.
- **Version reconciled to 0.2.0** across `repro_bundle.VERSION`, `nutrigx.py` (`result.json` now uses the shared constant), and SKILL.md frontmatter — previously `provenance.json` said 0.1.0 while `result.json` said 0.2.0 for the same run.

## Skill affected

nutrigx

## Checklist

- [x] Demo data included for reviewers to test — existing `tests/synthetic_patient.csv`

## Test output

Red/green TDD: the tests in `tests/test_repro_bundle.py` were written first and failed against the prior implementation (5 failed on the review-fix pass, 6 on the original refactor), then passed after each change.

```
$ python -m pytest skills/nutrigx/tests -q
20 passed in 0.59s
```

End-to-end verified, including the checksum verification flagged in review:

```
$ python skills/nutrigx/nutrigx.py --input skills/nutrigx/tests/synthetic_patient.csv --output /tmp/nutrigx_repro_check
[NutriGx] Done. Report: /tmp/nutrigx_repro_check/nutrigx_report.md
$ ls /tmp/nutrigx_repro_check/reproducibility
checksums.sha256  commands.sh  environment.yml  provenance.json
$ cd /tmp/nutrigx_repro_check && shasum -a 256 -c reproducibility/checksums.sha256
nutrigx_report.md: OK
```

Note: `clawbio/common/tests/test_report.py::test_generate_report_header_requires_skill_version` fails on main before this change — pre-existing, unrelated.


manuelcorpas (changes requested):
Checked the delegated hash line by line: the removed `sha256_file` used the identical binary-mode 8192-byte-chunk loop, the line format is preserved exactly (`f"{digest}  {name}"`, two spaces, `p.name` labels, trailing newline), and routing through `write_text_lf` is a genuine improvement, LF on every OS. Call ordering is safe, since `nutrigx.py:125-140` writes the report before building the bundle, and no consumer anywhere in the repo reads `checksums.txt` or the old top-level paths. JSON serialisation is unchanged, so no hash-relevant drift.

The SKILL.md change is correct and complete, which is worth calling out because it is the part most often missed: `SKILL.md:199-207` now matches `repro_bundle.py` file for file including `provenance.json`, `checksums.sha256` replaces `checksums.txt`, and `generate_report.py:220` was updated in the same pass. Docs, tests and code agree.

Two things to fix.

**A missing output is no longer recorded, it is erased.** On `main`, `sha256_file` caught `FileNotFoundError` and returned the string `"FILE_NOT_FOUND"`, so a missing artefact left a visible tombstone line in the manifest. `write_checksums` at `clawbio/common/reproducibility.py:152-154` does `if not p.exists(): continue` instead. So if `generate_report` fails to write `nutrigx_report.md` (bad output path, figure generation raising, disk full), the new bundle simply has two lines instead of three, `sha256sum -c` reports `2 OK`, and `docs/reproducibility.md:100` explicitly blesses that outcome. The bundle validates while the artefact it was supposed to attest is absent.

I have deliberately not called this critical, and the reason matters: silent-skip is the documented shared behaviour, `clinical-trial-finder/writers.py:698` already does the same, and on the only shipped call path the report always exists. You are bringing nutrigx into line with the repo, not deviating from it. But it is the one semantic change to an integrity artefact hiding inside a PR titled "refactor", and `test_checksums_skip_missing_files` at `tests/test_repro_bundle.py:71-84` asserts `"FILE_NOT_FOUND" not in content`, which locks it in as intended behaviour rather than flagging it. Please give it a line in the PR body, and ideally add an explicit existence check on `nutrigx_report.md` in `nutrigx.py` before the bundle is built.

**The generated `commands.sh` instructs a verification that cannot succeed from any directory.** `repro_bundle.py:50`:

```python
f"# 3. Verify checksums\n"
f"sha256sum -c checksums.sha256",
```

Before this PR, `checksums.txt` sat in `output_dir` next to `nutrigx_report.md`, so `cd output_dir && sha256sum -c checksums.txt` at least resolved that entry. Now `checksums.sha256` lives in `output_dir/reproducibility/` while its labels are bare basenames of files one level up, so from `reproducibility/` all three entries fail to resolve, and from `output_dir` the filename in the instruction is wrong. `docs/reproducibility.md:92-95` gives the correct form: `cd <output_dir> && sha256sum -c reproducibility/checksums.sha256`. Fixing the string and passing `anchor=output_dir` makes the `nutrigx_report.md` line verify properly. The input file and panel path will still not resolve, which is an argument for restricting the manifest to outputs as the docs describe.

Three smaller notes, none blocking:

- `provenance.json` at `:73-75` is the only file in the new `reproducibility/` directory that bypasses `write_text_lf`, using plain `.write_text(json.dumps(...))`. `clawbio/common/textio.py`'s own docstring names provenance JSONs in the list of artefacts that must be byte-identical regardless of OS, and calls itself "the single choke-point all bundle writers route through". Carried over unchanged from `main`, so not a regression, but this PR is what moves the file into the shared bundle directory and makes the inconsistency visible. One-line fix.
- The new `VERSION = "0.1.0"` at `:26` agrees with the SKILL.md frontmatter but disagrees with `nutrigx.py:4` ("v0.2.0") and `nutrigx.py:150` (`version="0.2.0"`), so every bundle ships a `provenance.json` claiming 0.1.0 next to a `result.json` claiming 0.2.0 for the same run. This predates the PR; hoisting it into a named constant is an improvement and a good moment to reconcile the three.
- The `defaults` channel silently disappears from the generated `environment.yml`: `main` hardcoded `[conda-forge, defaults]` and the helper defaults to `["conda-forge"]` only. All four conda deps are on conda-forge so this is functionally fine and arguably desirable given Anaconda's terms, but it is a change to a file the docs call a pinned conda environment, and no test would catch it.

Also cosmetic: `SKILL.md:243-247` lists the contents of `tests/` and was not updated to include `test_repro_bundle.py`, and the committed demo report at `examples/output/nutrigx_report.md:264` still says "in the output directory".

`test_commands_sh_in_reproducibility_dir` asserting the executable bit is a nice addition, since the old code never made `commands.sh` executable. Nothing in the diff introduces a scientific issue, and the privacy claim at `SKILL.md:258-260` still holds.

manuelcorpas (commented):
## pr-audit: advisory

Two of the four declared changes are clean, but one undeclared change breaks the bundle this PR ships, so please hold the merge until it is fixed.

### CRITICAL, `skills/nutrigx/repro_bundle.py:61-64`

```python
write_checksums(
    [input_file, panel_path, output_dir / "nutrigx_report.md"],
    output_dir,
)
```

`write_checksums` is called without `anchor`, so the labels stay bare filenames while the manifest itself moved down one directory into `reproducibility/`. `nutrigx_report.md` previously verified from `output_dir` and now resolves to nothing, so the bundle's own instruction at line 50, `sha256sum -c checksums.sha256`, fails on every entry. #352 gets this right with `anchor=output_dir`; this one does not. No test opens the manifest from `output_dir` and resolves the labels, which is why the suite stays green.

### Undeclared behaviour changes in a PR framed as a refactor

- `:53-59` omits `channels`, so the shared helper defaults to `["conda-forge"]` alone and the `defaults` channel present in the deleted `CONDA_ENV` literal is silently dropped from the pinned environment.
- `:73-75` writes `provenance.json` with a bare `write_text`, bypassing `write_text_lf`, which `clawbio/common/textio.py` documents as the single choke-point all bundle writers route through. It will emit CRLF on Windows while its three siblings emit LF.
- `:32-34` and `:62` change missing files from an explicit `FILE_NOT_FOUND` line to silent omission, so a run where the report failed to generate now yields a manifest that looks complete. Defensible, since it is `sha256sum -c` compatible, but it is not in SKILL.md.

### Warnings

- `:30` and `:66-72` put a wall-clock timestamp into both `commands.sh` and `provenance.json`, and `args` records absolute input and output paths verbatim, so the reproducibility bundle is not reproducible by construction. Pre-existing rather than introduced here, but #352 fixed exactly this class in the same batch.
- `generate_report.py:220` has no test coverage: `tests/test_nutrigx.py` asserts nothing about the reproducibility paragraph, and the new `test_repro_bundle.py` never imports `generate_report`.

### Clean

The SKILL.md change is confined to lines 202-206 inside the Reproducibility Bundle section. The nutrigenomics safety language is untouched: the "not diagnostic" caveat at line 186, the Disclaimer bullet at 197, and Limitations and Disclaimer at 269-271 are all byte-identical to base. No other nutrigx module imports `repro_bundle.sha256_file`, so removing the `FILE_NOT_FOUND` sentinel has no external consumer.

camlloyd:
Addressed both audit reviews in 6277508:

- **Anchor/verify (critical):** manifest is now outputs-only with `anchor=output_dir`, and `commands.sh` step 3 is `( cd "$(dirname "$0")/.." && sha256sum -c reproducibility/checksums.sha256 )`, which resolves regardless of where the script is invoked from. Verified end-to-end: `cd <output_dir> && shasum -a 256 -c reproducibility/checksums.sha256` → `nutrigx_report.md: OK`. Input and panel hashes moved to `provenance.json` (`input_sha256`, `panel_sha256`) rather than sitting in the manifest as unresolvable entries.
- **Missing report:** went further than the requested call-site check — `create_reproducibility_bundle` itself raises `FileNotFoundError` if `nutrigx_report.md` is absent, so the bundle can never validate while the report is missing. `test_checksums_skip_missing_files` was replaced by `test_missing_report_raises`.
- **`write_text_lf` for provenance.json:** done, with a byte-level LF assertion in the test.
- **Version reconciliation:** 0.2.0 everywhere; `nutrigx.py` now imports the shared `VERSION` constant for `result.json`, locked by `test_version_is_consistent`.
- **Channels:** keeping `conda-forge`-only deliberately (all four deps resolve there; avoids the Anaconda `defaults` ToS) — now declared in the PR body along with the other behaviour changes.
- **Cosmetics:** SKILL.md tests listing includes `test_repro_bundle.py`, checksum wording updated, and the committed demo report footer now says `reproducibility/` subdirectory. `generate_report.py:220` already had the correct wording; it now has a test.

`skills/nutrigx/tests`: 20 passed. The timestamp/absolute-paths warning (portable `commands.sh` à la #352) is left as-is per the "pre-existing" framing — happy to do that as a follow-up if wanted.

manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

## Domain review: nutrigx

Thanks for this. The refactor is faithful: every artefact the old bundle produced is still produced, and the delegation targets match the shared layer's real signatures.

**What I checked**

- `skills/nutrigx/repro_bundle.py:18` and `:19` import `sha256_file`, `write_checksums`, `write_commands_sh` and `write_environment_yml`. All four exist in the checkout at `clawbio/common/checksums.py:9`, `clawbio/common/reproducibility.py:130`, `:223` and `:169`, and the call sites match their signatures. `write_checksums(paths, output_dir)` at `repro_bundle.py:61` is positional and lines up with `reproducibility.py:130`; the keyword form at `repro_bundle.py:53` matches `reproducibility.py:169`.
- `_PROJECT_ROOT = Path(__file__).resolve().parents[2]` at `repro_bundle.py:14` resolves to the repo root from `skills/nutrigx/`, and `parents[3]` at `tests/test_repro_bundle.py:15` resolves to the same root from `skills/nutrigx/tests/`. Both are correct.
- The output contract moved to `<output_dir>/reproducibility/`, and `SKILL.md:200` onwards plus `generate_report.py:220` were both updated to say so, including the `checksums.txt` to `checksums.sha256` rename. Documentation and code agree.
- No scoring, panel or genotype logic is touched. `score_variants.py` and `data/snp_panel.json` are untouched, so no nutrient risk call changes. Nothing in the diff reaches the network, and `provenance.json` at `repro_bundle.py:73` stays on local disk exactly as before.

**Tests**

These are real tests, not shape assertions. `test_repro_bundle.py:71` compares the recorded digest against `common_sha256_file` applied to the actual fixture on disk, so a wrong or truncated digest would fail. `test_repro_bundle.py:52` asserts the executable bit that `reproducibility.py:240` sets. `tests/synthetic_patient.csv` and `data/snp_panel.json` both exist in the checkout, so `make_bundle` at `test_repro_bundle.py:27` runs against real files.

**Notes, none blocking**

1. Behaviour change in the missing-file path. The old helper returned the sentinel `FILE_NOT_FOUND` and still emitted a line; the shared `write_checksums` skips missing files silently (`clawbio/common/reproducibility.py:153`), which `test_repro_bundle.py:86` now codifies. Consequence: if `nutrigx_report.md` was never written, `checksums.sha256` omits it and `sha256sum -c` still passes on an incomplete bundle. This is the repo-wide convention already in `main`, so adopting it here is right, but it is a quieter failure than before.
2. `repro_bundle.py:73` writes `provenance.json` into `reproducibility/` without creating the directory itself. It works today only because the three preceding calls each `mkdir(parents=True, exist_ok=True)`. Worth a defensive `mkdir` if the calls are ever reordered or made conditional.
3. `repro_bundle.py:53` no longer passes `channels`, so the helper defaults to `conda-forge` alone (`reproducibility.py:209`), dropping the `defaults` channel the old inline recipe carried. numpy, pandas, matplotlib and seaborn all resolve on conda-forge, so the environment still builds. Flagging it only so the drop is deliberate.
4. Pre-existing, not introduced here, but the line was reformatted at `repro_bundle.py:32`: `--no-figures` is `action="store_true"` (`nutrigx.py:52`), so when the flag is set the generated command reads `python nutrigx.py --no-figures True`, which argparse rejects. The emitted `commands.sh` is then not runnable for that one flag combination. Worth a separate fix.
5. Also pre-existing: `VERSION = "0.1.0"` at `repro_bundle.py:25` disagrees with `version="0.2.0"` in `write_result_json` at `nutrigx.py:152`, so `provenance.json` and `result.json` label the same run differently. The constant makes it easy to align later.

**Static scan**

The scanner reported no findings on this PR, so there is nothing to contest.

No scientific, clinical or safety defect found. Note that no human has reviewed this.

---

Thanks for moving nutrigx onto the shared reproducibility layer; the direction is right and most of the mechanics are correct.

What I checked and found sound: no scoring, panel or recommendation logic is touched, so no nutrigenomic call changes; the SNP panel and the report text are untouched apart from the pointer at `skills/nutrigx/generate_report.py:219`, which now correctly says `reproducibility/`. The `sys.path` anchors resolve to the repo root in both files (`skills/nutrigx/repro_bundle.py:14`, `skills/nutrigx/tests/test_repro_bundle.py:20`). No network calls, no telemetry, no input data leaves the machine; the bundle still writes locally only. `write_commands_sh` emits the shebang itself, so the generated script is still valid. Ordering is safe today: `nutrigx.py:124` writes the report before `nutrigx.py:135` builds the bundle, so `nutrigx_report.md` exists when it is hashed. The static scan reported no findings on this PR and I found nothing it missed on the security side.

Blocking issue, checksum manifest is no longer verifiable.

`skills/nutrigx/repro_bundle.py:61-64` calls `write_checksums` with no `anchor`, so every label is a bare basename (`clawbio/common/reproducibility.py:158-159`). Two of the three entries, the input file and `snp_panel.json`, live outside `output_dir` entirely, and the third, `nutrigx_report.md`, now sits one directory above the manifest. The generated script at `skills/nutrigx/repro_bundle.py:50` tells the user to run `sha256sum -c checksums.sha256`, and `commands.sh` now lives in `reproducibility/` alongside the manifest, so from that directory all three entries fail to resolve. Run from `output_dir` instead, only `nutrigx_report.md` resolves. Before this change the manifest sat next to the report in `output_dir`, so that one entry did verify; this is a regression, not a pre-existing wart.

This is already a settled convention in the repo. See `skills/nfcore-scrnaseq-wrapper/provenance.py:226-228`, which records digests of files outside the bundle in `provenance.json` precisely "because bare-basename entries in checksums.sha256 would make `sha256sum -c` fail", and `provenance.py:321-322` plus `skills/bioqc-mcp/bioqc_mcp.py:1051`, which pass `anchor=output_dir`. Suggested fix: pass `anchor=output_dir` and restrict the manifest to files under `output_dir`, then carry the input and panel digests in `provenance.json` where they belong.

Smaller points, please address or tell me why not:

1. `skills/nutrigx/repro_bundle.py:73` writes `provenance.json` into `output_dir/reproducibility/` without creating the directory. It works only because `write_commands_sh` created it earlier in the same function. An explicit `mkdir(parents=True, exist_ok=True)` would stop a future reordering turning this into a `FileNotFoundError`.
2. `skills/nutrigx/repro_bundle.py:18` imports `sha256_file` but nothing in the module uses it. Its only consumer is the identity assertion at `skills/nutrigx/tests/test_repro_bundle.py:40-41`, which tests an import statement rather than behaviour. Either drop both, or use the helper in the module so the re-export is real.
3. Missing files are now dropped from the manifest silently (`clawbio/common/reproducibility.py:153-154`), and `skills/nutrigx/tests/test_repro_bundle.py:76-87` locks that in. The old code emitted a visible `FILE_NOT_FOUND` marker. A short manifest that looks complete is worse than a loud one; please log a warning when an expected artefact is absent.
4. The new tests compare digests recomputed in Python against digests written by the same helper (`skills/nutrigx/tests/test_repro_bundle.py:65-73`), so they cannot catch the label problem above. A test that runs `sha256sum -c` from the intended working directory, as `skills/nfcore-scrnaseq-wrapper/tests/test_bundle_portability.py:156-188` does, would.
5. The File Structure block in `skills/nutrigx/SKILL.md:243-245` still lists only `synthetic_patient.csv` and `test_nutrigx.py` under `tests/`. Please add `test_repro_bundle.py` while you are editing that file. The bundle description at `skills/nutrigx/SKILL.md:202-206` is otherwise correctly updated.
6. Not blocking, just flagging the behaviour change: the `defaults` channel present in the old inline `CONDA_ENV` is gone, since `write_environment_yml` defaults to `conda-forge` only and `skills/nutrigx/repro_bundle.py:53-59` does not pass `channels`. If that was deliberate, say so in the PR description; if not, the listed deps all resolve on conda-forge, so it should be fine.

---

manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Thanks, this is a tidy delegation to `clawbio.common` and it lines up end to end.

What I checked specifically:

- `repro_bundle.py:34` now takes `sha256_file` from `clawbio.common.checksums`, which drops the old `FILE_NOT_FOUND` sentinel that would have made `sha256sum -c` fail on a truncated run.
- The environment is unchanged in substance: `write_environment_yml` is called with python 3.11, numpy>=1.26, pandas>=2.2, matplotlib>=3.8, seaborn>=0.13 and `clawbio==0.1.0`, matching the old `CONDA_ENV` literal.
- The rename to `checksums.sha256` is consistent across the generated `commands.sh`, `SKILL.md:205` and `generate_report.py:220`, so the report no longer points readers at a path that does not exist.

No scoring, panel or interpretation code is touched, and nothing leaves the machine. The static scan reported no findings and I found nothing it missed. Merging.

**Non-blocking suggestions**

1. `write_checksums` silently skips absent files, so a run that never produced `nutrigx_report.md` yields a manifest that still looks complete. Your `tests/test_repro_bundle.py:258` documents that behaviour honestly; a warning on the skipped path would be a nice follow-up.
2. `repro_bundle.py:73` writes `provenance.json` into `reproducibility/` and relies on the three earlier calls having created it. An explicit `mkdir(parents=True, exist_ok=True)` would survive any future reordering.
3. `SKILL.md:68` still lists the bundle artefacts without the new `reproducibility/` location; worth a one line touch up next time you are in there.

---
Approved and merged.

PR #351, opened 2026-08-18
refactor: delegate sha256 helpers to clawbio.common.checksums

camlloyd:
## Summary

- Removes two private copies of the SHA-256 file-hashing loop that already lives in `clawbio.common.checksums.sha256_file`: `methylation_clock.py` and clinical-trial-finder's `writers.py` now import the shared function
- `writers.py` gains the standard project-root `sys.path` shim; `methylation_clock.py` already had one for its `common.reproducibility` imports and just adds the checksum import to it
- No behaviour change — identical digests, identical chunk size; bundle layouts untouched

## Skill affected

methylation-clock, clinical-trial-finder

## Checklist

- [x] SKILL.md is present and complete (YAML frontmatter + methodology) — unchanged, no methodology change
- [x] Demo data included for reviewers to test — existing skill demo data unchanged

## Test output

Red/green TDD: one delegation test per skill (asserting the module helper `is` the shared function), both watched failing against the old local helpers, then passing after the swap. The methylation-clock test stubs `pyaging` so it runs even where that optional heavy dependency cannot be imported.

```
$ python -m pytest skills/methylation-clock/tests skills/clinical-trial-finder/tests -q
165 passed, 1 skipped, 1 warning in 8.04s
```

(The 1 skip is `test_methylation_clock.py`'s pre-existing module-level `pytest.importorskip("pyaging")` on machines where pyaging is unavailable.)

### DIFF
--- a/skills/clinical-trial-finder/tests/test_sha256_delegation.py
+++ b/skills/clinical-trial-finder/tests/test_sha256_delegation.py
@@ -1,0 +1,17 @@
+"""Checksum helper must delegate to the shared clawbio.common layer."""
+
+import sys
+from pathlib import Path
+
+SKILL_DIR = Path(__file__).resolve().parent.parent
+PROJECT_ROOT = SKILL_DIR.parents[1]
+for p in (SKILL_DIR, PROJECT_ROOT):
+    if str(p) not in sys.path:
+        sys.path.insert(0, str(p))
+
+import writers  # noqa: E402
+from clawbio.common.checksums import sha256_file  # noqa: E402
+
+
+def test_sha256_delegates_to_common_layer():
+    assert writers._sha256 is sha256_file
--- a/skills/clinical-trial-finder/writers.py
+++ b/skills/clinical-trial-finder/writers.py
@@ -7,14 +7,19 @@ an output directory, creates the file, and returns its Path.
 
 import argparse
 import csv
-import hashlib
 import html as _html
 import json
+import sys
 from collections import Counter, defaultdict
 from datetime import datetime, timezone
 from pathlib import Path
 
-from constants import (
+_PROJECT_ROOT = Path(__file__).resolve().parents[2]
+if str(_PROJECT_ROOT) not in sys.path:
+    sys.path.insert(0, str(_PROJECT_ROOT))
+
+from clawbio.common.checksums import sha256_file as _sha256  # noqa: E402
+from constants import (  # noqa: E402
     CSV_COLUMNS,
     DEFAULT_PAGE_SIZE,
     DISCLAIMER,
@@ -37,16 +42,6 @@ from constants import (
 def count_recruiting(trials: list[dict]) -> int:
     """Count trials with RECRUITING status."""
     return sum(1 for t in trials if t["status"] == "RECRUITING")
-
-
-def _sha256(path: Path) -> str:
-    """Return hex SHA-256 digest of a file.  Reads in 8KB chunks to handle
-    large files without loading everything into memory."""
-    h = hashlib.sha256()
-    with open(path, "rb") as f:
-        for chunk in iter(lambda: f.read(8192), b""):
-            h.update(chunk)
-    return h.hexdigest()
 
 
 # ---------------------------------------------------------------------------
--- a/skills/methylation-clock/methylation_clock.py
+++ b/skills/methylation-clock/methylation_clock.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import argparse
-import hashlib
 import json
 import sys
 from datetime import datetime, timezone
@@ -18,6 +17,7 @@ _PROJECT_ROOT = Path(__file__).resolve().parents[2]
 if str(_PROJECT_ROOT) not in sys.path:
     sys.path.insert(0, str(_PROJECT_ROOT))
 
+from clawbio.common.checksums import sha256_file as _sha256  # noqa: E402
 from clawbio.common.reproducibility import (  # noqa: E402
     ReproCommand,
     ReproPath,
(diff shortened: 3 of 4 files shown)


manuelcorpas (changes requested):
The hash delegation itself is correct, and I checked it line by line rather than trusting the diff. Both removed helpers are character-for-character the same algorithm as `clawbio/common/checksums.py:9-15`: `open(path, "rb")`, `iter(lambda: f.read(8192), b"")`, `hexdigest()`. No normalisation, no sorting, no path-string hashing, no symlink handling on either side, and lowercase hex by definition. Missing-file behaviour is preserved too: the shared helper's bare `open()` raises `FileNotFoundError`, which is exactly what `writers.write_checksums` catches at `:696-699`. Only the type hint widens. No circular import. So there is no silent digest change here, which was my main worry going in.

Requesting changes on one thing, and it is about the import rather than the hash.

**`clinical-trial-finder` is a stdlib-only skill, and this makes it require opentelemetry, numpy and pandas.** `writers.py:17-21`:

```python
from clawbio.common.checksums import sha256_file as _sha256  # noqa: E402
```

Python imports parent packages before submodules, so that statement executes `clawbio/common/__init__.py`, which eagerly imports `report` (reaching `clawbio/common/audit.py:18`, `from opentelemetry import context`) and `scrna_io` (reaching `import numpy as np` and `import pandas as pd` at module scope). All three are top-level `clawbio` dependencies, not extras.

Every other module in this skill is pure stdlib, and its `SKILL.md:82-86` declares `openclaw.requires.bins: [python3]` with no `install:` block at all, where equity-scorer declares five. The file you are editing says so itself at `:298-305`: "matplotlib is optional, the skill works without it."

The concrete failure is the skill's own `demo.sh`. Its step 5 guards numpy with `if python3 -c "import numpy"` and an explanatory else branch, so it was written for a machine with python3, a git clone and nothing else. After this PR, steps 1 to 4 die before printing a line with `ModuleNotFoundError: No module named 'opentelemetry'`, raised from `audit.py` while the user is trying to search ClinicalTrials.gov. CI cannot see it, because the `test` job runs `uv sync --locked --all-extras`, so green is guaranteed and uninformative.

That is a large trade for replacing a six-line `hashlib` function. Three options, in my order of preference:

1. Make `clawbio/common/__init__.py` lazy via a module-level `__getattr__` re-export, so importing a leaf module does not drag `report` and `scrna_io` in. This also benefits the other ~35 skills that import `clawbio.common`, so I am happy to do it separately and ahead of this PR if you would rather not widen the scope.
2. Take only the methylation-clock half of this PR, which has no such constraint.
3. Declare the dependencies in `clinical-trial-finder/SKILL.md`. I would not recommend it: it converts a zero-dependency skill into a numpy/pandas/OTel skill for a checksum helper.

Two smaller notes:

**The new tests are identity assertions** (`assert writers._sha256 is sha256_file`). They go red if the delegation is reverted, which is what they claim to test, but they cannot go red if the shared helper's byte semantics change. Compounding that, there is no `clawbio/common/tests/test_checksums.py` anywhere in the repo, and `sha256_file` is only ever exercised indirectly through `test_reproducibility.py`, where every assertion compares its output against itself. So today, if someone added newline normalisation to `sha256_file`, every checksum in every skill would change and the entire suite would stay green. One golden-value test against a known digest would close that, and `clawbio/common/tests/` is a better home for it than either skill. That is our gap, not something to fix in this PR, but it is why I would not lean on the identity tests as coverage.

**`skills/methylation-clock/tests/test_sha256_delegation.py:18`** does `sys.modules.setdefault("pyaging", MagicMock())` at module scope, which leaks into the whole pytest session. Default collection is alphabetical, so `test_methylation_clock.py` is imported first and its `pytest.importorskip("pyaging")` at line 9 runs before the stub lands, which is why this is green. Under an explicit file ordering or any shuffling plugin, the stub lands first, `importorskip` succeeds against a `MagicMock`, and the real suite runs against a mock that returns a mock for everything. `monkeypatch.setitem(sys.modules, ...)` inside a fixture would contain it. The docstring explaining why the file is separate is good practice and worth keeping.

manuelcorpas (commented):
## pr-audit: advisory, no blocking findings

Verified behaviour-preserving byte for byte. Both deleted local `_sha256` bodies are character-for-character equivalent to `clawbio/common/checksums.py:9-15`: same `hashlib.sha256`, same `open(path, "rb")`, same `iter(lambda: f.read(8192), b"")`, same `.hexdigest()`. Same chunk size, same binary mode, hex rather than base64, streaming rather than read-whole-into-memory, and 64 characters untruncated. The hash covers identical bytes, so no previously published checksum is invalidated.

The one non-obvious contract also survives:

```python
# skills/clinical-trial-finder/writers.py:696-699
try:
    lines.append(f"{_sha256(p)}  {rel}")
except FileNotFoundError:
    pass  # Skip outputs that weren't generated
```

The shared helper's bare `open()` raises `FileNotFoundError` exactly as the deleted local one did, so the skip still fires. `methylation_clock.py:304-309` guards with `path.is_file()` and never relied on the exception.

### One request

Both new tests assert object identity only:

```python
assert writers._sha256 is sha256_file
assert methylation_clock._sha256 is sha256_file
```

That proves the wiring but nothing about the digest. If `sha256_file` were silently truncating, hashing the filename, or returning base64, both tests would still pass. Across #350, #351 and #352 there are twelve new test assertions on digests and not one compares against a known-good constant. Could you add one pinned digest of a committed fixture per skill? That converts the whole set from wiring checks into correctness checks, and it is two lines each.

Also checked and clean: signature widening from `Path` to `str | Path`; `sys.path.insert(0, _PROJECT_ROOT)` shadowing, since no repo-root `constants.py` exists and `from constants import ...` at `writers.py:22` still resolves to the skill-local module; `parents[2]` arithmetic; no new paths written; no patient data movement; no scoring logic touched.

manuelcorpas (approved):
Verified behaviour-preserving byte for byte: same algorithm, chunk size, binary mode, hex encoding and streaming read, so the digest covers identical bytes and no previously published checksum is invalidated. The FileNotFoundError skip contract in writers.py:696-699 survives the delegation.

Merging. The pinned known-good digest is worth adding in a follow-up, but it should not hold this up.

PR #352, opened 2026-08-18
feat: equity-scorer writes reproducibility bundle via clawbio.common

camlloyd:
## Summary

- equity-scorer's SKILL.md has always documented a `reproducibility/` output directory (`commands.sh`, `environment.yml`, `checksums.sha256`) that the code never actually produced — this implements it
- Both pipelines (VCF and ancestry CSV) now write the bundle via the shared `clawbio.common.reproducibility` layer; no hand-rolled writers
- `commands.sh` is a portable replay script (`write_portable_commands_sh` with `$CLAWBIO_ROOT`/`$OUTPUT_DIR` anchors, quoted paths, full-precision weights); `checksums.sha256` covers the derived artefacts under `tables/` and `figures/`, labelled relative to the output directory so `cd <output_dir> && sha256sum -c reproducibility/checksums.sha256` passes, including after a replay. Timestamped outputs (`report.md`, `result.json`) and out-of-tree inputs are deliberately excluded from the manifest; `result.json` already records the input digest as `input_checksum`.
- Note on scope of the provenance claim: `run_csv_pipeline` scores against literature-based heterozygosity estimates (labelled `het_source = "literature_estimate"` in its outputs), not values computed from the input. The bundle wraps those runs too — replaying reproduces the scoring, not a genotype-derived measurement.

## Skill affected

equity-scorer

## Checklist

- [x] SKILL.md is present and complete (YAML frontmatter + methodology) — unchanged; the code now matches its existing Output Structure section
- [x] Demo data included for reviewers to test — existing `examples/demo_populations.vcf` + `examples/demo_population_map.csv`

## Test output

Red/green TDD both rounds: the original 5 bundle tests were written first and failed before implementation; the review-driven rewrite (verification-from-output-dir, replay determinism, full-precision weights, pinned environment) again went red (6 failures) before the fix.

```
$ python -m pytest skills/equity-scorer/tests -q
43 passed in 5.51s
```

End-to-end verified, including the documented verification command and a replay via the generated script:

```
$ python skills/equity-scorer/equity_scorer.py --input examples/demo_populations.vcf \
    --pop-map examples/demo_population_map.csv --output "$OUT"
$ (cd "$OUT" && shasum -a 256 -c reproducibility/checksums.sha256)
tables/fst_matrix.csv: OK
... (9 artefacts, all OK)
$ bash "$OUT/reproducibility/commands.sh" && (cd "$OUT" && shasum -a 256 -c reproducibility/checksums.sha256)
... all OK after replay
```

### DIFF
--- a/skills/equity-scorer/SKILL.md
+++ b/skills/equity-scorer/SKILL.md
@@ -127,8 +127,15 @@ equity_report/
 └── reproducibility/
     ├── commands.sh          # Commands to re-run
     ├── environment.yml      # Conda export
-    └── checksums.sha256     # Input file checksums
+    └── checksums.sha256     # Output artefact checksums (tables/ + figures/)
 ```
+
+Verify with `cd <output_dir> && sha256sum -c reproducibility/checksums.sha256`.
+The manifest names the run's derived tables and figures only: `report.md` and
+`result.json` embed wall-clock timestamps and would never re-verify, so they
+are deliberately outside the envelope (every number in `result.json` is still
+covered via the byte-identical `tables/heim_score.json`). Input files are not
+listed — input provenance is recorded as `input_checksum` in `result.json`.
 
 ## Example Report Output
 
--- a/skills/equity-scorer/equity_scorer.py
+++ b/skills/equity-scorer/equity_scorer.py
@@ -39,9 +39,11 @@ from clawbio.common.parsers import parse_vcf_matrix
 from clawbio.common.checksums import sha256_file as _sha256_file
 from clawbio.common.report import write_result_json, DISCLAIMER as _DISCLAIMER
 from clawbio.common.reproducibility import (
+    ReproCommand,
+    ReproPath,
     write_checksums,
-    write_commands_sh,
     write_environment_yml,
+    write_portable_commands_sh,
 )
 
 # ---------------------------------------------------------------------------
@@ -78,6 +80,22 @@ POP_COLOURS = {
 # REPRODUCIBILITY BUNDLE (delegates to shared library)
 # ===================================================================
 
+def _repro_path(value: Path, *, output_dir: Path) -> ReproPath:
+    """Classify a path for portable reproducibility command rendering."""
+    value = value.resolve()
+    try:
+        value.relative_to(_PROJECT_ROOT)
+        return ReproPath(value, anchor="repo_root")
+    except ValueError:
+        pass
+
+    try:
+        value.relative_to(output_dir.resolve())
+        return ReproPath(value, anchor="output_dir")
+    except ValueError:
+        return ReproPath(value, anchor="auto")
+
+
 def _write_repro_bundle(
     output_dir: Path,
     input_path: Path,
@@ -86,27 +104,52 @@ def _write_repro_bundle(
 ) -> None:
     """Write commands.sh, environment.yml, and checksums.sha256 to
     <output_dir>/reproducibility/ via clawbio.common.reproducibility."""
-    cmd = "python skills/equity-scorer/equity_scorer.py --input %s" % input_path
+    args: list = ["--input", _repro_path(input_path, output_dir=output_dir)]
     if pop_map_path is not None:
-        cmd += " --pop-map %s" % pop_map_path
-    cmd += " --output %s --weights %s" % (
+        args += ["--pop-map", _repro_path(pop_map_path, output_dir=output_dir)]
+    args += [
+        "--output", ReproPath(output_dir, anchor="output_dir"),
+        # str() round-trips floats exactly; rounding here would record a
+        # different command than the one that ran.
+        "--weights", ",".join(str(w) for w in weights),
+    ]
+    write_portable_commands_sh(
         output_dir,
-        ",".join("%.2f" % w for w in weights),
+        ReproCommand(
+            script_path=Path("skills/equity-scorer/equity_scorer.py"),
+            args=args,
+            comment="Replay this ClawBio equity-scorer run",
+        ),
+        repo_root=_PROJECT_ROOT,
     )
-    write_commands_sh(output_dir, cmd)
 
+    # Version bounds mirror SKILL.md "Dependencies". pip_deps stays empty:
+    # the script imports clawbio.common via its own sys.path insert, so the
+    # environment needs no installed clawbio package.
     write_environment_yml(
         output_dir,
         env_name="clawbio-equity-scorer",
-        conda_deps=["numpy", "pandas", "matplotlib", "scikit-learn"],
+        conda_deps=[
+            "numpy>=1.24",
+            "pandas>=2.0",
+            "matplotlib>=3.7",
+            "scikit-learn>=1.3",
+        ],
         pip_deps=[],
+        python_version="3.11",
     )
 
-    checksum_targets = [input_path]
-    if pop_map_path is not None:
-        checksum_targets.append(pop_map_path)
-    checksum_targets += [output_dir / "report.md", output_dir / "result.json"]
-    write_checksums(checksum_targets, output_dir)
+    # Checksum the derived numerical/graphical artefacts only, labelled
+    # relative to output_dir so `cd <output_dir> && sha256sum -c` works.
+    # report.md and result.json embed wall-clock timestamps and would never
+    # re-verify; inputs live outside output_dir and are excluded (result.json
+    # already records the input digest as input_checksum).
+    checksum_targets: list = []
+    for sub in ("tables", "figures"):
+        subdir = output_dir / sub
+        if subdir.is_dir():
+            checksum_targets.extend(p for p in sorted(subdir.iterdir()) if p.is_file())
+    write_checksums(checksum_targets, output_dir, anchor=output_dir)
 
 
 # ===================================================================
--- a/skills/equity-scorer/equity_scorer.py
+++ b/skills/equity-scorer/equity_scorer.py
@@ -38,6 +38,11 @@ if str(_PROJECT_ROOT) not in sys.path:
 from clawbio.common.parsers import parse_vcf_matrix
 from clawbio.common.checksums import sha256_file as _sha256_file
 from clawbio.common.report import write_result_json, DISCLAIMER as _DISCLAIMER
+from clawbio.common.reproducibility import (
+    write_checksums,
+    write_commands_sh,
+    write_environment_yml,
+)
 
 # ---------------------------------------------------------------------------
 # Constants
@@ -67,6 +72,41 @@ POP_COLOURS = {
     "MID": "#CC79A7",
     "UNKNOWN": "#999999",
 }
+
+
+# ===================================================================
+# REPRODUCIBILITY BUNDLE (delegates to shared library)
+# ===================================================================
+
+def _write_repro_bundle(
+    output_dir: Path,
+    input_path: Path,
+    pop_map_path: Optional[Path],
+    weights: Tuple[float, ...],
+) -> None:
+    """Write commands.sh, environment.yml, and checksums.sha256 to
+    <output_dir>/reproducibility/ via clawbio.common.reproducibility."""
+    cmd = "python skills/equity-scorer/equity_scorer.py --input %s" % input_path
+    if pop_map_path is not None:
+        cmd += " --pop-map %s" % pop_map_path
+    cmd += " --output %s --weights %s" % (
+        output_dir,
+        ",".join("%.2f" % w for w in weights),
+    )
+    write_commands_sh(output_dir, cmd)
+
+    write_environment_yml(
+        output_dir,
+        env_name="clawbio-equity-scorer",
+        conda_deps=["numpy", "pandas", "matplotlib", "scikit-learn"],
+        pip_deps=[],
+    )
+
+    checksum_targets = [input_path]
+    if pop_map_path is not None:
+        checksum_targets.append(pop_map_path)
+    checksum_targets += [output_dir / "report.md", output_dir / "result.json"]
+    write_checksums(checksum_targets, output_dir)
 
 
 # ===================================================================
@@ -1005,6 +1045,8 @@ def run_vcf_pipeline(
         input_checksum=sha256_file(str(vcf_path)) if vcf_path.exists() else "",
     )
 
+    _write_repro_bundle(output_dir, vcf_path, pop_map_path, weights)
+
     print("\nDone.")
     print("  HEIM Score: %s/100 (%s)" % (heim_result["heim_score"], heim_result["rating"]))
     print("  Report: %s" % report_path)
@@ -1086,6 +1128,8 @@ def run_csv_pipeline(
         data=heim_result,
         input_checksum=sha256_file(str(csv_path)) if csv_path.exists() else "",
     )
+
+    _write_repro_bundle(output_dir, csv_path, None, weights)
 
     print("HEIM Score: %s/100 (%s)" % (heim_result["heim_score"], heim_result["rating"]))
     print("Report: %s" % report_path)
--- a/skills/equity-scorer/equity_scorer.py
+++ b/skills/equity-scorer/equity_scorer.py
@@ -80,8 +80,20 @@ POP_COLOURS = {
 # REPRODUCIBILITY BUNDLE (delegates to shared library)
 # ===================================================================
 
-def _repro_path(value: Path, *, output_dir: Path) -> ReproPath:
-    """Classify a path for portable reproducibility command rendering."""
+def _repro_input_arg(
+    value: Path,
+    *,
+    output_dir: Path,
+    var_name: str,
+    preflight: list,
+):
+    """Classify an input path for portable reproducibility command rendering.
+
+    Inputs outside the repo must not be recorded as absolute paths: a
+    deposited bundle would disclose the local filesystem location of a
+    cohort file. They are rendered as "$<var_name>/<basename>" with a
+    preflight line requiring the variable at replay time.
+    """
     value = value.resolve()
     try:
         value.relative_to(_PROJECT_ROOT)
@@ -93,7 +105,11 @@ def _repro_path(value: Path, *, output_dir: Path) -> ReproPath:
         value.relative_to(output_dir.resolve())
         return ReproPath(value, anchor="output_dir")
     except ValueError:
-        return ReproPath(value, anchor="auto")
+        preflight.append(
+            ': "${%s:?Set %s to the directory containing %s}"'
+            % (var_name, var_name, value.name)
+        )
+        return '"$%s/%s"' % (var_name, value.name)
 
 
 def _write_repro_bundle(
@@ -101,12 +117,26 @@ def _write_repro_bundle(
     input_path: Path,
     pop_map_path: Optional[Path],
     weights: Tuple[float, ...],
+    checksum_paths: List[Path],
 ) -> None:
     """Write commands.sh, environment.yml, and checksums.sha256 to
     <output_dir>/reproducibility/ via clawbio.common.reproducibility."""
-    args: list = ["--input", _repro_path(input_path, output_dir=output_dir)]
+    preflight: list = []
+    args: list = [
+        "--input",
+        _repro_input_arg(
+            input_path, output_dir=output_dir, var_name="INPUT_DIR",
+            preflight=preflight,
+        ),
+    ]
     if pop_map_path is not None:
-        args += ["--pop-map", _repro_path(pop_map_path, output_dir=output_dir)]
+        args += [
+            "--pop-map",
+            _repro_input_arg(
+                pop_map_path, output_dir=output_dir, var_name="POP_MAP_DIR",
+                preflight=preflight,
+            ),
+        ]
     args += [
         "--output", ReproPath(output_dir, anchor="output_dir"),
         # str() round-trips floats exactly; rounding here would record a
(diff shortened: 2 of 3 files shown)


manuelcorpas (changes requested):
Framing first: this is new functionality rather than a refactor, so nothing here is a silent regression. It implements a contract `SKILL.md:127-130` has documented for a while and the code never honoured, which is the right instinct.

Requesting changes because the bundle it produces cannot pass its own documented verification.

**`sha256sum -c` fails on every run, for two independent reasons.** `equity_scorer.py:109`:

```python
checksum_targets = [input_path]
if pop_map_path is not None:
    checksum_targets.append(pop_map_path)
checksum_targets += [output_dir / "report.md", output_dir / "result.json"]
write_checksums(checksum_targets, output_dir)
```

`docs/reproducibility.md:88-95` fixes the contract as `cd <output_dir> && sha256sum -c reproducibility/checksums.sha256`, and says the file "covers selected **output** files".

First, `write_checksums` labels every entry with `p.name` when no `anchor` is passed (`clawbio/common/reproducibility.py:161`), so the input VCF and population map, which live outside `output_dir`, are recorded as bare basenames that resolve to nothing from the verification directory. `sha256sum -c` reports `demo_populations.vcf: No such file or directory` and exits non-zero.

Second, and worse: both recorded outputs embed a wall-clock timestamp. `report.md` gets `datetime.now(timezone.utc).strftime(...)` at `:704`, and `result.json` gets `"completed_at": datetime.now(timezone.utc).isoformat()` via `clawbio/common/report.py:137`, to microsecond precision. So a byte-perfect replay of the same VCF with the same weights produces two mismatching digests, always.

A reproducibility bundle whose checksum file is guaranteed to fail is worse than no checksum file, because the first person to run the documented command concludes the analysis is non-reproducible. Minimal fix: pass `anchor=output_dir` (the prevailing convention here, used by `bioqc-mcp`, both nf-core wrappers, `omics-target-evidence-mapper`, `proteomics-de` and `multiqc-reporter`), drop the two inputs from the manifest, and either exclude the timestamped files or make the timestamps injectable.

**The manifest omits every numerical artefact the report is derived from.** `run_vcf_pipeline` writes `tables/population_summary.csv`, `tables/fst_matrix.csv`, `tables/heterozygosity.csv`, `tables/heim_score.json` and five PNGs (`:980-1020`), none checksummed. A replay producing an identical `report.md` but a different `fst_matrix.csv`, say from a different scikit-learn or BLAS, verifies clean because the manifest never looked. `multiqc-reporter` has a `_output_paths_to_checksum(output_dir)` helper for exactly this shape.

**`commands.sh` does not faithfully record the invocation.** `:89-96`:

```python
cmd += " --output %s --weights %s" % (
    output_dir,
    ",".join("%.2f" % w for w in weights),
)
```

Three issues in four lines. Weights are rounded to two decimals on the way out and re-parsed as floats on the way in (`:1158`), so `--weights 0.001,0.001,0.001,0.001` is recorded as `0.00,0.00,0.00,0.00` and the replay hard-errors on "HEIM weights must not all be zero"; most values survive normalisation, but the recorded command is not the command that ran. The script path is repo-relative while the input path is absolute, so it only works with `cwd` at the repo root, and the repo already ships `write_portable_commands_sh` with `ReproPath(anchor="repo_root")` for this, which methylation-clock uses. And paths are not shell-quoted, so any input path containing a space produces a broken script. `test_commands_sh_reproduces_invocation` asserts substring presence only, so none of the three can turn it red.

Smaller: `conda_deps` at `:101` carries no version bounds while `SKILL.md:164-168` specifies `pandas >= 2.0`, `numpy >= 1.24`, `scikit-learn >= 1.3`, `matplotlib >= 3.7`. An `environment.yml` described as an environment snapshot that pins nothing will not reproduce an environment; #350 gets this right for nutrigx. `python_version` is also left at the helper's 3.10 default while CI tests 3.11 to 3.13, and `pip_deps=[]` means an environment built from the file has no `clawbio` package (survivable, since the script does its own `sys.path` insert, but worth a comment).

On the tests: `test_checksums_match_common_sha256` is the strongest of the five and genuinely catches an ordering bug, since it would fail if the bundle were written before the report. It is nonetheless tautological on the digest itself, as both sides call the same helper, and no test pins a literal expected digest or asserts the complete file list, so a silently truncated manifest would still pass. None of the five would go red on the timestamp defect above.

Two things that are ours rather than yours. The `test` job's hardcoded step runs `skills/equity-scorer/tests/test_equity_scorer.py` by name, so `test_repro_bundle.py` runs on the PR only and stops running the moment it lands on `main`; I am fixing the allowlist. And we have no golden-value test for `sha256_file` anywhere, so its byte semantics are currently unpinned repo-wide.

Nothing in the diff introduces a scientific issue: no disclaimer removed, no remote fetch, no hardcoded value presented as computed. One observation worth a sentence in the PR body rather than a change: `run_csv_pipeline` scores against the hardcoded literature heterozygosity table at `:1075-1082`, correctly labelled `het_source = "literature_estimate"`, and the new bundle now wraps those runs in reproducibility packaging that implies computational provenance. The label carries the caveat, so this is a note, not a finding.

camlloyd:
All findings addressed in 13a7d85:

- **Verification contract**: `checksums.sha256` now lists the derived artefacts under `tables/` and `figures/`, labelled via `anchor=output_dir`, so `cd <output_dir> && sha256sum -c reproducibility/checksums.sha256` passes. Verified end-to-end, including after a replay through the generated `commands.sh`. The out-of-tree inputs and the timestamped `report.md`/`result.json` are excluded (took the exclude option rather than injectable timestamps — `result.json` already carries `input_checksum` for input provenance, matching the methylation-clock convention of checksumming `tables/` + `figures/` only).
- **Manifest coverage**: all four tables (incl. `fst_matrix.csv`, `heim_score.json`) and all five PNGs are now checksummed; `test_checksums_cover_derived_artefacts` pins the complete expected list.
- **commands.sh**: switched to `write_portable_commands_sh` with `ReproCommand`/`ReproPath` (`repo_root` anchoring for the script and repo-resident inputs, `$OUTPUT_DIR` for output, quoted paths) and weights recorded via `str()` at full precision — `0.001,0.001,0.001,0.997` round-trips exactly.
- **environment.yml**: pins `pandas>=2.0`, `numpy>=1.24`, `scikit-learn>=1.3`, `matplotlib>=3.7` per SKILL.md, `python=3.11`; `pip_deps=[]` now carries a code comment explaining the sys.path-based `clawbio` import.
- **Tests**: rewritten red-first; the suite now includes `test_checksums_verify_from_output_dir` (every entry must resolve and match from `output_dir`) and `test_checksums_reproduce_on_replay` (two runs must produce byte-identical manifests) — both go red on the original timestamp and anchoring defects. 43 passed.

PR body updated with the `run_csv_pipeline` literature-estimate provenance note. Left the CI allowlist and the repo-wide `sha256_file` golden test alone per the review, since those are being handled separately.

manuelcorpas (changes requested):
All three review items are genuinely fixed, and I verified by execution rather than by reading: ran both pipelines, ran the documented `sha256sum -c` against the produced bundle, replayed through the generated `commands.sh`, and reverted each fix in a scratch copy to confirm the tests catch it.

**Item 1, the verification contract: fixed, both causes.** `anchor=output_dir` gives relative labels and the timestamped files are excluded, so `cd <output_dir> && sha256sum -c reproducibility/checksums.sha256` now exits 0.

Excluding is the weakest of the three options, but it is sound here for a reason worth stating out loud: `tables/heim_score.json` is byte-identical to `result.json["data"]`, so every number in the excluded `result.json` is still under checksum. Only `report.md`'s prose sits outside. Normalisation would have been cheap given the minute-precision stamp, and injection better still, but this is not a blocker.

**Item 2, coverage: fixed.** All four tables and five figures for the VCF path, both tables and both figures for the CSV path.

**Item 3, `commands.sh`: all three fixed.** Weights round-trip at full precision via `str(w)`; I ran `--weights 0.001,0.001,0.001,0.997` and the recorded command replayed to the same `HEIM Score: 71.5/100`, where the old `%.2f` would have written `0.00,0.00,0.00,1.00`. Paths anchor to `$CLAWBIO_ROOT` and `$OUTPUT_DIR`. Quoting is real rather than asserted: a run with `demo pops.vcf` under an `in dir/` produced a script that replayed cleanly.

**The tests are strong.** I reverted nine fixes one at a time and every single one goes red, including dropping a single table from the manifest. ``test_checksums_cover_derived_artefacts`` pins a concrete complete list, so a silently truncated manifest cannot pass. That is better than most of what lands here.

Requesting changes on one defect the fix itself introduced.

**The manifest is built by globbing rather than by naming, so it certifies artefacts from previous runs.** `equity_scorer.py:147-152`:

```python
checksum_targets: list = []
for sub in ("tables", "figures"):
    subdir = output_dir / sub
    if subdir.is_dir():
        checksum_targets.extend(p for p in sorted(subdir.iterdir()) if p.is_file())
```

`--output` defaults to the fixed relative `equity_report` (line 1195), so two back-to-back invocations land in the same directory. A VCF run followed by a CSV run produces a CSV-run manifest of nine entries, five of which the CSV pipeline never computes:

```
tables/fst_matrix.csv, tables/heterozygosity.csv,
figures/fst_heatmap.png, figures/heterozygosity.png, figures/pca_plot.png
```

`sha256sum -c` exits 0 on all nine. That run's `result.json` records `het_source = "literature_estimate"` and zero pairwise comparisons, while its own bundle certifies computed FST and PCA outputs, so the honest label is contradicted by the manifest sitting next to it. `multiqc-reporter:195` names its fixed outputs and only rglobs the one genuinely variable directory; this took the glob half without the naming half. An `_output_paths_to_checksum(output_dir)` helper per pipeline closes it.

Two smaller ones:

**`SKILL.md:130` now describes the opposite of the file.** It reads `checksums.sha256 # Input file checksums`. That was accurate before this commit, when `checksum_targets` began `[input_path, pop_map_path]`; at head the manifest contains no input checksums at all. Worth also saying in the doc that `report.md` and `result.json` are deliberately outside the envelope, since that justification currently exists only as a source comment.

**The exact-set assertion.** I injected a bogus `tables/STALE_FROM_PREVIOUS_RUN.csv` into the manifest and the suite stayed green. Changing `test_checksums_cover_derived_artefacts` to `set(entries) == {...}` would close it, and would have caught the contamination bug during development.

Two follow-ups that are ours rather than yours, and I will file both:

- Five of the nine entries are PNGs, and every one carries a `tEXt` chunk naming the exact matplotlib build (`Matplotlib version3.10.1`). `environment.yml` pins `matplotlib>=3.7` with no upper bound, so a recreated environment will report five failures with byte-identical numbers underneath. Same "would never re-verify" reasoning you applied to `report.md`.
- `write_portable_commands_sh` emits a bare `python`. On a machine where that is Python 2 the generated replay dies with a `SyntaxError`. Affects `multiqc-reporter` and the nf-core wrappers equally, so it belongs against `clawbio/common/reproducibility.py`. No test executes `commands.sh`, which is why CI cannot see it.

manuelcorpas (commented):
## pr-audit: advisory

Best-engineered of the three bundle PRs. Determinism is handled properly: iteration is over a fixed tuple `("tables", "figures")` and `sorted(subdir.iterdir())`, so no dict or set ordering leaks in; no timestamp enters the manifest, with `report.md` and `result.json` deliberately excluded and a reasoned comment saying why; `anchor=output_dir` yields relative labels. That last point is the thing #350 got wrong, so it is worth calling out.

Two things to confirm before merge, neither blocking on its own.

### Path leakage into a deposited bundle

`equity_scorer.py:96`

```python
return ReproPath(value, anchor="auto")
```

A user-supplied VCF outside the repo falls through to `anchor="auto"`, which the shared helper renders as a bare absolute path in `commands.sh`. A deposited bundle would then disclose the local filesystem location of a cohort file.

`equity_scorer.py:123` has the same shape: `repo_root=_PROJECT_ROOT` bakes the generating machine's absolute repo path into `commands.sh` as the `CLAWBIO_ROOT` default, which leaks a username or hostname and stops the bundle being byte-identical across machines. The portability assertion at `tests/test_repro_bundle.py:74` only excludes the output directory, not the repo root, so it cannot catch this.

### Environment cannot replay a VCF run

`equity_scorer.py:129-140`

```python
conda_deps=["numpy>=1.24", "pandas>=2.0", "matplotlib>=3.7", "scikit-learn>=1.3"]
```

The comment says this mirrors SKILL.md, but SKILL.md:164 lists `biopython>=1.82` as required for VCF parsing and it is missing here.

### Minor

No `provenance.json`, which #350's nutrigx bundle does write. The two PRs land inconsistent bundle contracts in the same batch; worth picking one.

### Confirmed clean

Scoring logic is untouched. The diff has zero deletions anywhere: the only changes are an import block at `:41-47`, one self-contained new section at `:79-152`, and two one-line call sites at `:1091` and `:1175`, both placed after `write_result_json`. Nothing in the HEIM scoring, Fst, heterozygosity or PCA paths is modified.

No participant-level data enters the bundle. It writes only `commands.sh`, `environment.yml` and `checksums.sha256`, copies no data, and the checksum targets are population-level aggregates. `figures/pca_plot.png` is per-sample but pre-existing and only hashed, never copied.

Same request as on #351: none of the digest assertions compare against a known-good constant, so one pinned fixture digest would be worth adding.

camlloyd:
All review-2 items and both review-3 confirmations addressed in 6a5eaaa:

- **Manifest contamination (blocker)**: the glob is gone. Each pipeline passes its own named artefact list to `_write_repro_bundle` (9 entries for VCF, 4 for CSV). Verified by execution: a VCF run followed by a CSV run into the same directory now produces a 4-entry CSV manifest while the stale `fst_matrix.csv`/PCA artefacts still sit on disk uncertified — `test_manifest_ignores_artefacts_from_previous_runs` pins exactly that scenario, red on the old glob.
- **Exact-set assertions**: `test_checksums_cover_derived_artefacts` and the CSV bundle test now assert `set(entries) == {...}`, so both a truncated and a padded manifest go red. The injected `STALE_FROM_PREVIOUS_RUN.csv` case you ran is covered.
- **SKILL.md:130**: the tree entry now reads "Output artefact checksums (tables/ + figures/)", with a paragraph documenting the envelope — `report.md`/`result.json` deliberately excluded (timestamps; `result.json`'s numbers remain covered via the byte-identical `tables/heim_score.json`), inputs covered by `input_checksum`.
- **Path leakage**: out-of-repo inputs are no longer rendered via `anchor="auto"`. They render as `"$INPUT_DIR/<basename>"` / `"$POP_MAP_DIR/<basename>"` with `: "${INPUT_DIR:?...}"` preflights, so no cohort-file location enters a deposited bundle and replay fails loudly rather than guessing. Verified by executing the generated script: unset vars → exit 1 with the message; vars set (paths with spaces) → clean replay. The `CLAWBIO_ROOT` default at the bottom of `commands.sh` still bakes the generating repo path — that comes from `write_portable_commands_sh` and affects multiqc-reporter and the nf-core wrappers equally, so by the same reasoning as your bare-`python` follow-up I've left it for a change against `clawbio/common/reproducibility.py` rather than fixing one caller here. Happy to take that follow-up.
- **environment.yml**: adds `biopython>=1.82` per SKILL.md's required list.
- **Pinned digest**: `test_sha256_file_matches_known_digest` asserts a literal known-good constant for fixed fixture bytes.

On `provenance.json` vs #350: agreed the two bundle contracts should converge; suggest settling that at the `clawbio.common` level with the other follow-ups rather than in this PR.

Full suite: `pytest skills/equity-scorer/tests/ clawbio/common/tests/test_reproducibility.py clawbio/tests/test_portable_commands.py -q` → 96 passed. The documented `cd <output_dir> && sha256sum -c reproducibility/checksums.sha256` verified end-to-end, including after replay through the generated `commands.sh`.

manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

## Domain review: equity-scorer

Thanks for this. The reproducibility bundle is carefully thought through, and several of the decisions here are the right ones for the wrong-looking reasons, so I have checked each against the checkout rather than taking the comments at face value.

### What is correct

- **The manifest matches what the pipelines actually write.** All nine VCF entries listed at `skills/equity-scorer/equity_scorer.py:1122` correspond to real writes at `equity_scorer.py:944`, `:946`, `:953`, `:955` and `:962-978`. The four CSV entries at `:1219` match `:1051`, `:1053`, `:1058` and `:1062`. No phantom artefacts, no missing ones.
- **Excluding `report.md` and `result.json` is justified, not a convenience.** `report.md` stamps wall-clock time at `equity_scorer.py:671`, and `result.json` carries `completed_at` from `clawbio/common/report.py:137`. Both would indeed never re-verify.
- **The coverage claim in `SKILL.md:133-138` holds.** `result.json` receives `data=heim_result` at `equity_scorer.py:1004`, and `tables/heim_score.json` is a dump of that same dict at `:955`; `het_source` and `fst_estimator` are set at `:935-936`, before the dump, so nothing numeric in the envelope falls outside the manifest. The `summary` block is a strict subset.
- **`environment.yml` matches the documented dependencies exactly.** The five bounds at `equity_scorer.py:160` are the same five at `SKILL.md:164-168`.
- **Naming the manifest rather than globbing is the right call**, and `tests/test_repro_bundle.py:122` is a genuine test of it: it runs the VCF pipeline then the CSV pipeline into one directory and asserts the stale `fst_matrix.csv` is on disk but absent from the manifest. That is a real trap, not a restatement of the implementation.
- **Nothing leaves the machine.** The bundle is written entirely under `output_dir`. No network calls, no telemetry, no variant or phenotype content in any outbound path. `write_ro_crate`, which packages every file under `output_dir` via `rglob`, is correctly not used here.

### What needs answering before merge

**1. CI does not run this test file, so the green tick carries no information about this change.** `.github/workflows/ci.yml:120` runs `skills/equity-scorer/tests/test_equity_scorer.py` by name, and no job in the workflow invokes bare `pytest`. `pytest.ini:25` lists `skills/equity-scorer/tests`, but nothing in CI uses `testpaths`. Worse, `test_equity_scorer.py` never calls `run_vcf_pipeline` or `run_csv_pipeline`, so no CI job on 3.11, 3.12 or 3.13 executes `_write_repro_bundle` at all. Please add a step for `tests/test_repro_bundle.py`, or fold these cases into the file CI already runs. This matters most for two of them: `test_repro_bundle.py:137` asserts byte-identical replay, which rests on matplotlib PNG output being byte-stable and is exactly the kind of assumption that breaks silently on a version bump; and `test_repro_bundle.py:195` pins a SHA-256 constant that I cannot verify by inspection and that no runner currently checks.

**2. The privacy claim is stronger than the code delivers.** The docstring at `equity_scorer.py:90-96` says an absolute path would "disclose the local filesystem location of a cohort file". The directory is suppressed, but the basename is still written verbatim, as `test_repro_bundle.py:181` asserts. Cohort filenames are frequently identifying in themselves. Separately, `write_portable_commands_sh` still embeds the absolute repo root as the `CLAWBIO_ROOT` default at `clawbio/common/reproducibility.py:111`, which on most machines contains the operator's home directory. The mitigation is worth keeping; please narrow the docstring so it does not promise more than it does.

**3. Filenames are interpolated into shell text without escaping.** `equity_scorer.py:112` renders `"$INPUT_DIR/<basename>"` and `:109` embeds the basename inside `${VAR:?message}`. Both sit inside double quotes, where bash still expands `$(...)` and backticks at replay time. An awkwardly named input would break the script or execute on replay. `shlex.quote` or single quoting would close this. The same rendering style pre-exists at `clawbio/common/reproducibility.py:93`, so I read this as hardening rather than a regression introduced here.

**4. A skipped figure and a deleted figure are indistinguishable in the bundle.** `write_checksums` silently drops paths that do not exist (`clawbio/common/reproducibility.py:152-154`). If matplotlib is absent, the block at `equity_scorer.py:981` is skipped, the manifest is quietly four or five lines shorter, and `sha256sum -c` still passes on a partial bundle. The comment at `equity_scorer.py:174-182` acknowledges this, but a third party holding only the bundle cannot tell. Recording the skipped names, or a count, would make the omission legible.

### Scanner re-check

The static scan reported no critical and no warning findings, and five INFO items about missing `SKILL.md` sections. I confirmed those are accurate and correctly marked pre-existing: `skills/equity-scorer/SKILL.md` on main has no `## Trigger`, `## Scope`, `## Example Output`, `## Gotchas` or `## Agent Boundary` heading, and this PR neither adds nor removes any of them. Nothing to contest, and none of it is a reason to hold this PR.

No prompt-injection text and no secrets in the diff. Holding on point 1; the rest are notes.

---

manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Thanks, this is a careful bundle. Naming this run's artefacts explicitly instead of globbing `tables/` and `figures/` is the right call, and `test_manifest_ignores_artefacts_from_previous_runs` (tests/test_repro_bundle.py:321) pins exactly that trap. Recording the weights via `str()` rather than a rounded format keeps the replay command faithful, and the `${INPUT_DIR:?...}` preflight fails loudly instead of guessing a path. No scoring logic is touched, so there is no risk of a different HEIM score. This is merging; the points below are for later.

**Non-blocking suggestions**

1. The new tests do not run in CI. `.github/workflows/ci.yml:120` runs only `test_equity_scorer.py`, and that file never calls `run_vcf_pipeline` or `run_csv_pipeline`, so nothing behind the green tick exercises `_write_repro_bundle` (equity_scorer.py:81). A `pytest skills/equity-scorer/tests/ -v` step would cover both.

2. `commands.sh` still embeds the generating machine's checkout path as the `CLAWBIO_ROOT` default (clawbio/common/reproducibility.py:111). Cohort paths are masked, the repo path is not; worth a line in SKILL.md given the deposit use case you describe.

3. `write_checksums` skips missing files silently (clawbio/common/reproducibility.py:153), so a run without matplotlib yields a manifest with the five figures quietly absent that still passes `sha256sum -c`.

The static scan reported no findings here and I agree with it.

---
Approved and merged.

PR #354, opened 2026-08-19
Correct strand guidance and stale expression values in the gi-* skills

boldakov:
Corrections to the six `gi-*` skills, each re-measured against the live Genomic Intelligence API on 2026-08-19 using the skills' own bundled fixtures. No new skills, no infrastructure changes.

## Why

Several claims in these pages had gone stale or were wrong in a way that misleads rather than merely misinforms — one of them makes an integration test fail against production today.

## What changed

**`gi-splice` — a wrong-strand result is undetectable.** The page told you to submit gene-sense but not what happens if you don't. Measured on the bundled HBB fixture at the default 0.5 threshold:

| | Sites | Scores |
|---|---|---|
| Gene-sense | 8 | all ≥ 0.9994 |
| Genomic strand | 7 | 0.5187 – 0.9949 |

Positions and the donor/acceptor split differ, but nothing in the counts or scores tells you which orientation you sent. The page now says so plainly: get the strand right on input, because you will not catch it afterwards.

**`gi-promoter` — the existing near-zero claim holds, and is now quantified.** Gene-sense scores 4 promoter windows topping at 0.92; the genomic strand scores 0 above threshold, maximum 0.32 anywhere in the locus. Because that heuristic is real *here*, the page now also warns explicitly that it does **not** generalise — on `gi-splice` an empty result never indicates a strand error. The demo expectation was also stale: it described a 19 kbp fixture with "near-zero" promoter calls; it is 25.8 kbp and calls 4 of 26.

**`gi-expression` — documented values no longer reproduce.** The page claimed ~2.86 log(TPM+1) ≈ 16 TPM for the bundled fixture, with a ~2.89 / ~0.4 strand pair, under a note asserting those figures had been verified as still current. Measured now: **0.95** gene-sense, **0.06** genomic strand. The strand *ratio* is intact and is what the guidance is really teaching; the absolute values are not, so they are replaced with a measured figure marked indicative plus guidance to re-measure rather than trust a documented absolute.

Consequently `test_demo_mode_predicts_high_hbb_k562_expression` asserted `log_tpm > 2.0` and **fails against production today** — reproducible on `main` before this branch. Its floor is now 0.3, chosen to sit between the two strands so it still catches the strand/plumbing regression it was written for without pinning a prediction that moves with the model checkpoint.

Also newly documented: `description` is free text the model is conditioned on, not an enum, so paraphrases are not equivalent. On one fixture `"K562"` scored 1.38, `"K562 cells"` 0.70, and the canonical assay-format string 0.95. Worth knowing before comparing runs.

## Earlier commits on this branch

The first four commits carry contract corrections staged before this pass: per-task sequence floors with an in-regime warning, closed `options` handling, the 9,198 bp expression window gate with `--tss-index` support, sync/async framing (delivery is a per-request choice via `Prefer: respond-async`, not a property of the task), and `request_id` recovery from the `X-Request-Id` header for responses whose body omits it.

## Testing

59 GI tests pass, including the live integration tests: `clawbio/tests/test_gi_contract.py` plus the `tests/` for all six `gi-*` skills. Catalog regenerates with no diff.


manuelcorpas (changes requested):
The strand work is correct, and I want to record how that was checked rather than just asserting it. There is no reverse-complement code anywhere in this PR or the repo, so double-RC is structurally impossible; orientation is a property of the fixtures. Each was confirmed by searching for known transcript-sense motifs in the sequence and in its reverse complement:

- `expression_hbb_k562.fa`: HBB 5'UTR at offset 4,599, absent from the RC. Gene-sense, and the TSS sits exactly at the centre of the 9,198 bp window, matching `EXPRESSION_TSS_RADIUS`.
- `splice_hbb.fa`: HBB 5'UTR at offset 2,324, which equals 5,229,395 minus 5,227,071, independently confirming both orientation and the HBB TSS.
- `promoter_tp53.fa`: TP53 start codon at offset 10,952, which equals 7,687,546 minus 7,676,594, the known GRCh38 ATG.
- `annotation_tp53.fa`: TP53 ATG found **only** in the reverse complement, so this fixture is genomic plus strand. That is correct and intended, since annotation is a de novo gene finder that predicts strand, and the PR did not wrongly propagate gene-sense guidance to it.

I also independently confirmed the six length floors against the published schema: promoter 300 bp, splice 100 bp, enhancer 50 bp, chromatin 200 bp, annotation 1,000 bp, expression 9,198 bp, all capped at 500,000, matching `TASK_MIN_BP` and `REQUEST_MAX_BP` exactly. The `[4599, n-4599]` bound at `gi_runner.py:155` is right for a half-open `[tss-4599, tss+4599)` window. Donor and acceptor are never labelled locally, so genomic-order mislabelling cannot occur, and `gi-splice/SKILL.md:169`'s warning that a wrong-strand splice result is undetectable from the output is the honest and non-obvious thing to say.

The expression change is the good version of this failure mode: it **deletes** stale absolutes and a note that had asserted those figures "still hold", rather than replacing one hardcoded number with another. No `np.random`, no fabricated data, no credential moved, request path byte-identical to `main`. Rate-limit handling is correct, which I checked specifically given our history with silent 429s: `_check` raises on any non-2xx, the poll loop continues only on an explicit 202, so a 429 cannot collapse into a plausible empty prediction.

Two things before merge.

**1. The scored-window field path disagrees between the docs and the code, and one of them is wrong.** `gi-expression/SKILL.md:177` says to check `meta.task_specific_counts.scored_window`; `gi_runner.py:277-280` reads it from `data.input`:

```python
inp = _as_obj(data.get("input"), "data.input")
out["tss_index"] = inp.get("tss_index")
out["scored_window"] = inp.get("scored_window")
```

If the code's path is the wrong one, `summary["scored_window"]` is always `None`, the `if window:` guard at `:355` never fires, and the "Scored window" line never appears in `report.md`. That line is the only defence the documentation offers against a silently wrong `--tss-index`, and the paragraph immediately above it exists to warn that a bad index "does not error, it lies". No test asserts either path against a well-formed response. Could you confirm the real path against a live response and pin it with a test? This is the same defect class this PR itself fixes at `4077ee4`, where the splice table read field names the API never returned and rendered three of four columns as `-` in every report for months.

**2. `conftest.py:40-61` makes a plain `pytest` call the live API with no opt-in.**

```python
def _default_gi_key(root: Path) -> None:
    if os.environ.get(GI_KEY_VAR):
        return
    key = _demo_key_from_env_example(root)
    if key:
        os.environ[GI_KEY_VAR] = key
```

`pytest.ini:3` sets no `-m "not integration"` and CI runs `uv run pytest "skills/$skill/tests" -v` unfiltered. Each of the six gi skills carries an `@pytest.mark.integration` test that shells out and uploads its bundled FASTA. Before this PR those failed fast on "GI_API_KEY is not set"; now they transmit. The sequences are public reference fixtures so this is not a patient-data leak, but it turns "Remote inference, opt-in required", the phrase every gi SKILL.md still carries, into on-by-default, and it bills a shared community key on every CI run of every PR touching a gi skill. Note the contrast inside that same file: the RNA suites are gated behind `--include-rna` at `:105-115`. Please gate this the same way, or have CI pass `-m "not integration"`.

Lesser, worth a follow-up rather than blocking: the headline feature is the worst-covered path. `--tss-index` reaches the wire only via `gi_client.py:206-207`, and all eight new expression tests exercise `validate_expression_input`, a pure local function, while none assert the POSTed body. Delete those two lines and the suite stays green while `--tss-index` becomes a silent no-op: the user submits a 50 kbp locus, the API scores some default window, and the report claims a prediction for a TSS the model never saw. The same applies to the splice table rename, the pre-header FASTA refusal, the `X-Request-Id` fallback, and the `non_json` to `http_error` code rename, which is a breaking change for any caller branching on `GIError.code`.

Also minor: `gi_runner.py:355-359` indexes `scored_window` positionally with no type guard, inside the one block whose stated purpose is refusing wrong-typed response fields, so `{"start": 0, "end": 9198}` raises an uncaught `KeyError(0)`. The six SKILL.md files now tell users to read `RateLimit-*` headers, but `GIError` captures only `X-Request-Id` and `result.json` stores the parsed body, so that information is discarded. And `expression_hbb_k562.fa`'s header spans 9,199 bases 1-based-inclusive while the file holds 9,198; the other five fixtures are exactly 1-based inclusive, and now that offsets are load-bearing a user deriving one from the header would be one out.

The two new test files are genuinely good and not tautological, driving `predict`, `wait_for_job`, `submit_async`, `_summarize`, `_write_report` and `run_skill` end to end with nine malformed payloads each. The gap is that the coverage lands on response validation rather than request construction.

boldakov:
Both blocking items are fixed, and the first one came out differently than predicted, so the measurement is below rather than just the change.

Scored-window field path. I called the live endpoint rather than reasoning from the code. A POST to /v1/tasks/expression/predict against production at contract revision 5, submitting the bundled expression_hbb_k562.fa with the skill's own default options, returns the pair in both places, populated and identical:

    data.input.tss_index                     = 4599
    data.input.scored_window                 = [0, 9198]
    data.input.submitted_sequence_length     = 9198
    meta.task_specific_counts.tss_index      = 4599
    meta.task_specific_counts.scored_window  = [0, 9198]

So summary["scored_window"] was never None and the "Scored window" line does render. Both paths are real. The runner keeps reading data.input because only that one carries submitted_sequence_length, which the same report line prints. What was actually wrong is narrower than the report: the docs named one path, the code read the other, and nothing pinned either, so a response that later dropped one would have failed in a user's report rather than in CI. gi-expression/SKILL.md now names both and says which one the report uses, and clawbio/tests/test_gi_response_shape.py pins both from the recorded live body, drives _summarize, and asserts _write_report emits "[0, 9198)" and "TSS index 4599".

conftest.py calling the live API. Correct, and it was ours: 15fae3c applied the demo key unconditionally. The key default now applies only under a new --include-gi-live flag, and without it the gi integration tests skip rather than fail, so CI stays green without transmitting. Scoped to the six gi test directories rather than to the integration marker, since just-prs-mcp and phylogenetics-builder also use that marker and do not call our API; phylogenetics-builder still collects its 73 tests unchanged. Deliberately without the explicit-target escape hatch the RNA gate has, because ci.yml names skills/gi-expression/tests explicitly and that is the invocation that was transmitting. Verified: pytest skills/gi-expression/tests skills/gi-promoter/tests is 13 passed, 2 skipped in 1.03s, and the same command with --include-gi-live runs the live tests and passes 12 of 12.

Request-construction coverage. You filed this as a follow-up; it was the most useful point in the review, so it is in this push. I deleted the two lines in _build_body as you suggested. The result is what you predicted: the existing suites report 191 passed, 1 skipped, fully green, while the runner prints "OK, wrote report.md" and states a confident prediction for a window the model was never asked to score. clawbio/tests/test_gi_request_construction.py now asserts the POSTed JSON, 9 tests, 4 of which fail on that deletion. One detail worth flagging: my first draft stubbed gi_runner.Client, which intercepts above _build_body, and it stayed green through the deletion. The seam had to be the session.

scored_window indexed positionally. Fixed. A dict is truthy, so {"start": 0, "end": 9198} raised an uncaught KeyError(0) past run_skill's handler, inside the block whose purpose is refusing wrong-typed fields. There is now an _as_bounds in the style of _as_obj and _as_objs, refusing a non-array, a wrong length and a non-integer element, with a test on the dict case.

Fixture header. Fixed, and settled against the reference rather than by moving whichever end looked wrong: both readings were available, 5222472-5231669 if the start was 1-based and 5222473-5231670 if it was 0-based. I fetched all three candidate spans from Ensembl and compared them to the file. Only chr11:5222473-5231670 reproduces it, as an exact reverse complement, 9,198 of 9,198 bases. So the start was 0-based and the end was already right. That comparison also independently re-confirms the RC(gene-sense) claim in the header. The guard asserts end - start + 1 equals the base count for all six fixtures rather than pinning this one string, and it fails on the pre-fix header.

Two of your lesser points are not in this push, so as not to bundle them with the blocking work. The RateLimit-* headers are documented in the six SKILL.md files and discarded by the client, which is a real gap. The non_json to http_error rename is a breaking change for anything branching on GIError.code, and it has no test. Tell me whether you want either in this PR or as a follow-up.

Head is b2d2dda. Verified across the gi surface at 212 passed, 6 skipped, where the 6 skips are the newly gated live tests. Note that clawbio/tests/ is in pytest.ini but in no CI job, so none of these guards runs upstream; that is separate from this PR and I can open it separately if useful.

boldakov:
Both of the points I left out of the last push are in, as two commits you can drop independently if you would rather have them as a follow-up.

Coverage for the four behaviours you named alongside --tss-index. Each new test was confirmed by deleting the behaviour it guards and watching it fail, then restoring.

The http_error rename needed two tests rather than one, and finding that out is the useful part. My first attempt renamed the default in GIError.__init__ back to non_json and the suite stayed green, because _check builds its own error envelope for an unparseable body with the code written into it, so err.get("code", ...) never reaches that default on the path the rename was about. The two lines spell the same enum value and neither one covers the other. There is now a test for each: one drives an unparseable body, one drives an error body that parsed but carries no error key. A third asserts a code the API did send is carried through unchanged, so http_error stays the default rather than an override.

The pre-header FASTA refusal, the X-Request-Id fallback and the splice table rename each have their own tests. For the splice table the rendered row is the only place the field names are observable, so one test asserts each column carries its value and another pins the pre-fix shape rendering as a row of dashes. Without that second one a revert is invisible: the table still renders, it is just empty, which is what shipped for months.

RateLimit-* headers. I measured rather than assuming, by sending a 4 bp sequence to promoter so the call is a 422 and runs no inference:

    POST /v1/tasks/promoter/predict -> 422
        ratelimit-limit: 200
        ratelimit-policy: 200;w=60
        ratelimit-remaining: 199
        ratelimit-reset: 1
    GET /health -> 200
        no rate-limit headers at all

Two things in that changed the prose as well as the code. The wire form is lowercase, so the extraction normalises rather than indexing, since a requests response is case-insensitive and a plain mapping is not. And "on any response" was wrong for /health, so the six SKILL.md sentences now say "on any /v1/tasks/ response". RateLimit-Policy states the window the limit applies over, which is what turns "per-minute burst allowance" in gi-promoter/SKILL.md into a measured claim instead of an asserted one.

GIError now carries the headers on every error and prints them only on a 429. On a 422 the numbers bury the message that says what was actually wrong with the request; on a 429 they are the message. On the success path they go into result.json under rate_limit, and wait_for_job's poll-failure branch records them too, since it builds its error without going through _check. Retry-After is captured but was not observed: seeing it means provoking a 429 on the shared demo key, which would spend other people's allowance to confirm a header name.

Head is 4c99830, 18 tests added across the two commits, gi surface at 230 passed and 6 skipped where the 6 are the live tests gated behind --include-gi-live.

One thing I did not touch. outward-text.sh, our own prose checker, flags the emoji in the gi SKILL.md frontmatter and headings. 96 of the skills in this repo carry an emoji field, so that is your house style and not something to strip on a PR. Our checker is the thing that is wrong for this repo, and I have recorded it on our side.

manuelcorpas (approved):
Re-audited at head `4c99830e`. Both blocking items are genuinely resolved, and I verified by execution rather than by reading your commit messages: a 21-mutation battery reverting each behaviour to see whether the suite goes red. 13 of 16 attempted reverts produced named failures. That is unusually mutation-resistant for this repo.

**The `scored_window` field path is fixed properly.** SKILL.md now names both real response paths, states which one the runner reads, and the code reads that one. The `_as_bounds` guard also closes the `KeyError(0)` path the old positional index had. Reverting the extraction goes red on 4 tests; removing the report line goes red on 2.

**The conftest auto-key is fixed.** Gating on `--include-gi-live` before `_default_gi_key` runs, plus the collection-time skip, and a bare `pytest` across the six gi skills now gives 21 passed / 6 skipped with `GI_API_KEY` unset. Two design choices are right and worth recording: scoping to `GI_TEST_DIRS` rather than to the `integration` marker, so you do not hijack `just-prs-mcp` or `phylogenetics-builder`; and deliberately omitting the `_is_explicit_target` escape hatch the RNA gate has, since CI names `skills/$skill/tests` explicitly and an exemption would gate nothing.

`b2d2dda8` genuinely asserts the wire. Deleting the two `_build_body` lines goes red on 4 tests including the end-to-end one, and putting the seam at the session rather than at `gi_runner.Client` is the placement that makes that true. The falsy-zero case and the closed-schema case are both pinned.

The rate-limit work is clean and does not weaken the property I checked last time: `_check` still raises on any non-2xx, `wait_for_job` still returns only on 200 and continues only on explicit 202, so a 429 still cannot collapse into a plausible empty prediction. No retry or backoff was added. Missing headers are omitted rather than nulled, values are never parsed to int so a malformed `Retry-After` cannot raise, and the 429 suffix is correctly scoped so a 422's own message is not buried.

Merging. Three follow-ups, none blocking:

**1. One test cannot fail.** `clawbio/tests/test_gi_request_construction.py:187`:

```python
code, recorder = self._run(tmp_path, monkeypatch, [], "promoter", 4000)
assert "tss_index" not in recorder.body
```

The `[]` means `--tss-index` is never passed, so the assertion is satisfied by argparse's `None` default rather than by the `tss_index = args.tss_index if task == "expression" else None` guard it is named for. Deleting that guard leaves the suite green. Change `[]` to `["--tss-index", "1200"]`. If the guard ever goes, `--tss-index` on any of the other five tasks lands in the body, every per-task model is `additionalProperties: false`, and every such request becomes a hard 422.

**2.** The paired `--tss-index applies only to gi-expression` warning is also uncovered; the sibling `--description` case has a test, so this reads as an omission.

**3.** Whitespace-within-line FASTA acceptance (`"".join(line.split())`) is uncovered. Reverting it is a loud refusal rather than a wrong answer, so it is minor, but it was on the list.

One thing for us rather than you: gating the live tests means `test_demo_mode_predicts_high_hbb_k562_expression` now runs nowhere automatically, and its own docstring says a regression there usually means the strand wiring broke. That is the regression this PR exists to protect, now without a standing guard. The gate is still the right call against billing a shared key on every PR; I will sort a scheduled run with a repo secret.

Issue #356, opened 2026-08-21
gwas-prs: five of six bundled PGS accessions carry the wrong trait, and three cited PMIDs are unrelated papers


manuelcorpas:
`skills/gwas-prs` ships six PGS Catalog scoring files and a curated trait map. **Five of the six accessions are bound to the wrong trait, the wrong publication and the wrong variant count**, and at least three of the cited PMIDs point at entirely unrelated papers.

Verified 2026-08-21 against the PGS Catalog REST API (`https://www.pgscatalog.org/rest/score/<id>`) and PubMed E-utilities.

## Accession to trait

| File header in `skills/gwas-prs/data/` | PGS Catalog (authoritative) |
|---|---|
| PGS000001 Breast cancer, Mavaddat 2015, 77 | Breast cancer, Mavaddat N 2015, 77. **Correct** |
| PGS000004 Coronary artery disease, Abraham 2016, 46 | **Breast cancer**, Mavaddat N 2018, **313** |
| PGS000011 Atrial fibrillation, Tada 2014, 12 | **Coronary artery disease**, Tada H 2015, **50** |
| PGS000013 Type 2 diabetes, Vassy 2014, 8 | **Coronary artery disease**, Khera AV 2018, **6,630,150** |
| PGS000039 Body mass index, Khera 2019, 97 | **Ischemic stroke**, Abraham G 2019, **3,225,583** |
| PGS000057 Prostate cancer, Schumacher 2018, 147 | **Coronary heart disease**, Natarajan P 2017, **57** |

## Citations

`skills/gwas-prs/curated_scores.json` cites PMIDs that are not the papers named:

| Cited as | PMID | Actually |
|---|---|---|
| Tada H et al. (2014), atrial fibrillation | 25087147 | "Critical care nurses' knowledge of measures to prevent ventilator-associated pneumonia", Am J Infect Control 2014 |
| Abraham G et al. (2016), coronary artery disease | 27535533 | Lek et al., "Analysis of protein-coding genetic variation in 60,706 humans", Nature 2016 (the ExAC paper) |
| Khera AV et al. (2019), body mass index | 30239722 | Pulit et al., "Meta-analysis of GWAS for body fat distribution in 694,649 individuals", Hum Mol Genet 2019 |

## The files are not PGS Catalog downloads

[...]

PR #357, opened 2026-08-21
fix(gwas-prs): correct panel citations and stop demo panels shadowing real PGS Catalog scores

manuelcorpas:
Fixes the two defects in #356 that can be fixed without a cross-skill rename, and one that #356 did not know about and which is more serious than anything in it.

## The bug #356 missed: silent score substitution

`gwas_prs.py` consulted the local cache path before contacting the API, and never checked what it found:

```python
local_path = DATA_DIR / f"{pgs_id}_hmPOS_{args.build}.txt"
if local_path.exists():
    filepath = local_path
    print(f"  Using pre-downloaded scoring file: {filepath}")
```

The bundled demo panels sit at exactly that path. So `--pgs-id PGS000013`, on an ordinary non-demo run, returned the bundled **8-variant type 2 diabetes panel** in place of the real PGS000013 (Khera 2018, **coronary artery disease, 6,630,150 variants**) and never touched the network. Same for the other five accessions.

That is not a mislabelled demo, it is a wrong score returned in answer to a specific request, silently, with a reassuring "Using pre-downloaded scoring file" line.

`is_curated_demo_panel()` now detects this, and the fetch path emits an explicit warning naming #356.

**It warns rather than refuses, and that was not the original plan.** I first made it refuse and fetch the genuine score, which is the correct behaviour. CI caught what that costs: `clawbio-bench` drives this exact path offline as `--pgs-id PGS000013` for eight scoring-arithmetic cases, so refusing took the `gwas-prs` harness from 62.5% to **0.0%**. That destroys real coverage rather than fixing anything, so refusal has to land in the same step as an upstream bench change that lets it ask for the panel by its honest id. The detection and its tests ship now; the refusal is a follow-up. A comment in the code says so, to stop someone flipping it and zeroing the harness again.

Worth noting: **the bench already knew.** Its own ground truth reads "NOT a PGS Catalog entry, PGS000013 is GPS_CAD (Khera 2018, coronary artery disease, 6.6M variants)", and it names the correct Vassy T2D accessions as PGS000031/32/33. So this defect had independent external confirmation sitting in a pinned dependency.

## Citations

`CURATED_SCORES` (live) and `curated_scores.json` (read by nothing) carried different publications for the same panel, and between them named three papers with no connection to their trait:

| Cited as | PMID | Actually |
|---|---|---|
| Tada 2014, atrial fibrillation | 25087147 | ventilator-associated pneumonia, nursing knowledge, Am J Infect Control |
| Abraham 2016, coronary artery disease | 27535533 | Lek et al., the ExAC paper, Nature |
| Khera 2019, BMI | 30239722 | Pulit et al., body fat distribution, Hum Mol Genet |

All six re-resolved against PubMed E-utilities on 2026-08-21:

| Panel | Loci | Publication | PMID |
|---|---|---|---|
| Type 2 diabetes | 8 | Vassy JL et al. (2014) *Diabetes* | 24520119 |
| Atrial fibrillation | 12 | Tada H et al. (2014) *Stroke* | 25123217 |
| Coronary artery disease | 46 | Abraham G et al. (2016) *Eur Heart J* | 27655226 |
| Breast cancer | 77 | Mavaddat N et al. (2015) *J Natl Cancer Inst* | 25855707 |
| Prostate cancer | 147 | Schumacher FR et al. (2018) *Nat Genet* | 29892016 |
| BMI | 97 | Locke AE et al. (2015) *Nature* | 25673413 |

Two things worth noting, because they change how bad this looked. First, **the science was never wrong**: the panels contain genuinely correct loci for their declared traits (PITX2 4q25 for AF, FGFR2/TOX3 for breast cancer, 8q24 for prostate, TCF7L2/PPARG for T2D). Only the identity and attribution were wrong. Second, the three bad PMIDs are transcription errors rather than invention: the author named the right papers in prose every time, so this restores what was intended rather than substituting my guess. The 12-variant AF panel matching Tada's "Twelve-single nucleotide polymorphism genetic risk score" and the 97-locus BMI panel matching Locke's 97 loci are both good confirmation.

## Drift

`curated_scores.json` is read by no code, which is why it drifted from `CURATED_SCORES` unnoticed. It disagreed on BMI in a way that mattered: the JSON said Khera 2019 while the Python said Locke 2015, and Locke is the correct one. It is now generated from `CURATED_SCORES` and pinned against it by a test.

Each data file now declares what it is (`#clawbio_panel=curated_demo`, panel id, source publication, and a note) and no longer carries a `#pgs_id` header claiming an accession it has no right to.

## Tests

53 new tests in `tests/test_score_provenance.py`, red before and green after. They pin the citations, the cross-source agreement, the file headers, the declared-versus-actual variant counts, and the substitution guard. Included is a banned-string check that fails if any retired PMID reappears anywhere in the skill, which is the only guard that fires for a stray copy nobody edited.

Full suite: **78 passed**.

## Behaviour is unchanged

Demo percentiles are identical before and after (T2D 14.3%, AF 98.3%, CAD 8.2%, and the rest), so this is a provenance and safety fix with no numerical effect. I also ran the four consuming skills' suites; the 8 failures in `just-prs-mcp` are `PackageNotFoundError: No package metadata was found for clawbio`, reproduce identically on unmodified `main`, and are environmental rather than caused by this change.

## Deliberately not in scope

Refusal on shadowing, per the bench interaction above. And this does **not** renumber the accession keys. `PGS000013` and friends remain the dict keys, because ~69 references across 17 files and six skills (`profile-report`, `just-prs-mcp`, `wgs-prs`, `ancestry-risk-profiler`, plus a workshop deck and CLAUDE.md) consume them, and bundling that rename here would make the change unreviewable. `curated_panel_id` (`CLAWBIO-T2D-8` and so on) is now the honest identifier and `pgs_catalog_id` is explicitly `None`, so the rename is a mechanical follow-up. I will open a separate issue.

Closes the citation half of #356. Leaves open the accession rename and the refusal-on-shadowing, both of which need coordination beyond this skill.

---

> [!NOTE]
> **Medium Risk**
> Corrects scientific provenance for genetic risk reports and still serves bundled demo panels for real PGS accessions (warns instead of refusing, to keep clawbio-bench passing). Mislabeling of published PRS remains possible if consumers ignore the new flags.
> 
> **Overview**
> Stops treating the six bundled GWAS-PRS files as PGS Catalog scores. They are now ClawBio-curated demo panels (`CLAWBIO-T2D-8` and similar), with honest headers, corrected PubMed citations, and `pgs_catalog_id` set to `None` except for the lossy PGS000001 derivative.
> 
> `--pgs-id` and `--trait` still hit those files in the download-cache path, so a request for e.g. PGS000013 still scores the 8-locus T2D demo instead of the real catalog score. The run now warns on stdout, marks the report and `result.json` (`curated_demo_panel`), and drops the PGS Catalog provenance line. It does **not** refuse, because that zeros the offline `clawbio-bench` harness.
> 
> `curated_scores.json` is generated from `CURATED_SCORES` and pinned by new provenance tests. Scoring numbers are unchanged. Accession keys are not renamed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 84ef734db252d6945420eb72d0e8549014f5fb42. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


manuelcorpas:
## Round 2, after an independent adversarial review

That review found four defects in my own first pass, all of them classes this repo spent the week flagging in contributor PRs. Recording them here because the symmetry is the point.

1. **SKILL.md documented the opposite of the code.** `b2aed2e` made the refusal real, `fe598fe` reverted it, and the sentence promising it stayed. An agent reading that file to decide whether it needed `--demo` got a false assurance. This is the same defect I flagged in #346.
2. **The disclosure reached stdout only.** `prs_report.md` still printed `**Scoring files**: PGS Catalog` and `result.json` still carried a bare `pgs_id` that `profile-report` reads and propagates. Two `print()` calls are a note, not a mitigation. This is disclosure-not-abstention, which I flagged in #348.
3. **I guarded one of two call sites.** The `--trait` path had the identical shortcut and no warning, and it is the likelier path.
4. **The only behavioural change I shipped had no test.** Deleting the entire warning block left all 78 green. Same class as the vacuous guard tests I flagged in #344, #346 and #354 the same day.

All four are fixed. The panel marker now rides on the result itself and reaches the Score Details table, the provenance footer and `result.json` as `curated_demo_panel` / `curated_panel_id` / `pgs_catalog_id`. Both non-demo call sites warn. Three mutations (drop the `--trait` guard, drop the `--pgs-id` warning, restore the false provenance line) each go red with a named failure.

Two overclaims of my own also corrected:

- **`#source_publication=` implied the weights came from the cited paper.** They do not. The real PGS000001 shares 60 of 77 rsIDs with the bundled panel and 31 of those 60 weights differ by more than 0.02; Vassy 2014 is a 62-locus score against 8 loci here, Abraham 2016 is 49,310 SNPs against 46. Renamed to `#loci_reference=` with an explicit approximate-weights note. Correcting the citations and then attaching them to a header that reads as attribution was replacing one provenance error with a subtler one.
- **"That accession belongs to a different published score"** is false for PGS000001, which genuinely is Mavaddat 2015 PRS77_BC. Now recorded per panel; `pgs_catalog_id` is no longer a blanket `None`.

Plus the drift gaps: `trait_id` (EFO) was silently dropped in the first pass and is restored; the pin covers five more fields; and the "generated" claim is now backed by `generate_curated_scores.py --check` rather than being a claim about a file nothing writes.

Demo percentiles unchanged, data rows byte-identical to main in all six files, 105 tests passing, CI green including `scientific-audit`.

Still open in #356 and deliberately out of scope: the accession rename across six consuming skills, and refusal-on-shadowing, which needs clawbio-bench to be able to ask for the panel by `curated_panel_id`.

PR #358, opened 2026-08-22
fix(claw-semantic-sim): drop the false "works out of the box" demo claim

### LINKED ISSUE #353: [Bug]  semantic_sim.py not available
## Skill affected

claw-semantic-sim

## Describe the bug

Topline README.md and claw-semantic-sim SKILL.md both describe:
`python semantic_sim.py --demo --output demo_report`
But this file is not available.

## Steps to reproduce

Search for file `semantic_sim.py`

## Expected behaviour

Location of `semantic_sim.py` is documented in SKILL.md and located where documented.

## Environment

Github repo

AmirF194:
## Summary

- `skills/claw-semantic-sim/SKILL.md` and `README.md` both instruct readers to run
  `python semantic_sim.py --demo`, but that script and the `05-00..05-05` pipeline
  behind it have never existed in this repository.
- Replaces the runnable demo command in both docs with a plain "not yet implemented"
  status note.

## Skill affected

claw-semantic-sim (docs only, no code change)

## Root cause

`git log --all --diff-filter=A` over `semantic_sim.py` and `heim-sem` returns zero
hits on any branch, so the script was never committed alongside the docs that
describe it, not recently deleted. `skills/catalog.json` already records this
correctly (`maturity_tier: spec-only`, `has_script: false`, `has_demo: false`,
`demo_command: null`); only the prose in `SKILL.md` and `README.md` claimed the
demo runs "out of the box".

## Checklist

- [x] SKILL.md is present and complete (YAML frontmatter + methodology) - unchanged,
      only the Status section and demo block were touched
- [ ] Tests included and passing (`pytest -v`) - no code changed, nothing to test
- [ ] Demo data included for reviewers to test - not applicable, no demo exists yet

## Test output

Docs-only change. Verified in a clean `python:3.11-slim` container with
`uv sync --locked --all-extras`:
- `uv run python clawbio.py list` - exits clean, lists claw-semantic-sim unchanged
- `uv run pytest tests/ -v` - 171 passed, including `test_checked_in_catalog_is_current`
  (confirms `skills/catalog.json` needed no regeneration, since only prose changed)
- Competing-PR sweep on both touched files: 0/8 open PRs touch them

Not verified: I did not implement the missing pipeline itself, only corrected the
docs. Whether and when `semantic_sim.py` gets built is a separate, larger decision.

Fixes #353

### DIFF
--- a/README.md
+++ b/README.md
@@ -381,9 +381,8 @@ Computes a Semantic Isolation Index for diseases using 13.1M PubMed abstracts an
 - **Temporal Drift**: how research focus evolves over time
 - Publication-quality **4-panel figure**
 
-```bash
-python semantic_sim.py --demo --output sem_report
-```
+**Status: not yet implemented.** `semantic_sim.py` and the pipeline scripts behind it are not
+in this repository yet (see `skills/claw-semantic-sim/SKILL.md`); there is no runnable demo.
 
 **Key finding**: Neglected tropical diseases are **+38% more semantically isolated** (P < 0.0001, Cohen's d = 0.84). 14 of the 25 most isolated diseases are Global South priority conditions. Knowledge silos kill innovation — a malaria immunology breakthrough could help leishmaniasis, but the literatures don't talk to each other.
 
--- a/skills/claw-semantic-sim/SKILL.md
+++ b/skills/claw-semantic-sim/SKILL.md
@@ -98,13 +98,10 @@ Neglected tropical diseases (NTDs) are significantly more semantically isolated 
 05-05-heim-sem-integrate.py # Merge with biobank + clinical trial dimensions
 ```
 
-### Demo (works out of the box)
+## Status
 
-```bash
-python semantic_sim.py --demo --output demo_report
-```
-
-The demo uses pre-computed embeddings and metrics for 175 GBD diseases and generates the full 4-panel figure instantly.
+**Not yet implemented.** `semantic_sim.py` and the `05-00..05-05` pipeline scripts described
+above are not present in this repository yet; there is no runnable demo.
 
 ## Example Output
 


manuelcorpas (approved):
Verified the claim before approving, and it holds more strongly than the PR title suggests.

`skills/claw-semantic-sim/` contains exactly one file, `SKILL.md`. `git ls-files | grep semantic_sim` returns nothing across the whole repository. So the old wording was not optimistic about setup, it invited people to run a file that has never existed, which is why #353 got filed.

The repo already knew. `skills/catalog.json` records `has_script: false`, `has_demo: false`, `demo_command: null`, `maturity_tier: spec-only`, and `tests/test_generate_catalog.py:56-57` asserts exactly that. The machine-derived fields have been honest the whole time while the human-facing docs contradicted them. Thank you for closing that gap, and for finding it from the outside.

Two things you got right that are worth naming, because they are the parts people usually miss:

- The replacement is concrete rather than hedged. It names the missing artefacts and says plainly there is no runnable demo, so an agent reading it can tell there is no action to take. Replacing an overclaim with vague wording would have been almost as unhelpful.
- You scoped it correctly. `claw-metagenomics` and `claw-ancestry-pca` both carry the same `### Demo (works out of the box)` heading, and both actually ship their `.py`, so leaving them alone was right.

CI is green on all seven jobs.

Merging. Two follow-ups, neither blocking and neither really yours:

1. `SKILL.md` still carries an `## Example Output` block two lines below the new "there is no runnable demo" note, and its transcript ends with `Figures saved to: demo_report/`, which is verbatim the `--output demo_report` argument from the line you deleted. Same file, one surface over. Happy to take it in this PR if you want to push one more commit, otherwise I will pick it up.
2. `skills/catalog.json` still reports `"status": "mvp"` and `docs/CLAWBIO-BRIEF.md:94` still shows **MVP**, both sourced from a hardcoded `MVP_FOLDERS` set in `scripts/generate_catalog.py` whose own comment defines membership as "have working Python". That is maintainer-side and I will open an issue for it.

AmirF194:
Thanks for the careful check and for merging. Good catch tying it back to catalog.json's own has_demo/maturity_tier fields, that is a cleaner way to state the same fact than I found. Sounds like you've got both follow-ups covered, happy to leave them with you.

Issue #360, opened 2026-08-22
CI never runs on R and shell skills: no workflow run has ever existed for #329


manuelcorpas:
Querying by head SHA and by branch name, **zero workflow runs have ever existed** for `feat/claw-amplicon-qc` (#329). Not `action_required`, not queued: none, across five remediation rounds. That PR's contributor has been iterating for weeks with the repo's own pipeline never once verifying their work.

Two separate defects behind it.

**1. The `test` job cannot run a non-pytest suite.** `.github/workflows/ci.yml` runs `uv run pytest "skills/$skill/tests"` whenever a `tests/` directory exists. `claw-amplicon-qc` ships `tests/run_test.sh` with 17 assertions, so pytest collects 0 items and exits 5, and the job goes red for a reason that has nothing to do with the skill. Flagged on 2026-08-11 and still open. Consequence: `run_test.sh` has never executed in CI, so those 17 assertions are unverified by anything but the contributor's own machine, and the red tick trains everyone to ignore the signal.

**2. Fork-PR approval is a silent gate.** Runs for first-time contributors park in `action_required` and surface in the API as "no checks reported", which is indistinguishable from a pass in `gh pr checks` output. Five open PRs sat in that state this week. There is no notification, so approval only happens if a maintainer happens to look.

Suggested fixes:

- Detect the test runner per skill: pytest when `tests/test_*.py` exists, `bash tests/run_test.sh` when that exists, and fail explicitly (rather than exiting 5) when a `tests/` directory contains neither.
- Provision the R toolchain for R skills, or the job will fail on `Rscript` not found even once it invokes the right harness.
- Surface pending fork approvals somewhere visible, or relax the policy for PRs that touch only `skills/`.

Related: #359 covers the same blind spot in `generate_catalog.py`, which also cannot see non-Python skills. The two share a root cause worth stating plainly: **this repo's tooling assumes every skill is Python, and neither the catalog nor CI reports honestly when that assumption fails.**

manuelcorpas:
**Correction: defect 1 in this issue was already fixed, and I should have checked before filing.**

`33f98be` (2026-08-11, "fix(ci): pick the test runner from what tests/ contains, not pytest always") already added the `pycount` guard and the NOT VERIFIED path. So the claim above that pytest runs unconditionally and "is still open" was wrong. Apologies for the noise.

[...]

PR #361, opened 2026-08-22
fix(catalog,ci): stop assuming every skill is Python

### LINKED ISSUE #359: generate_catalog.py cannot see non-Python skills, and silently reverts hand-corrected entries
`scripts/generate_catalog.py` derives every maturity signal from Python file patterns, so a skill written in any other language can never register above `spec-only` no matter what it ships.

The evidence path, using `claw-amplicon-qc` (#329) as the live case. That skill ships a 1,274-line R script, a 17-assertion shell harness and a working `--demo`. The contributor hand-corrected its catalog entry to `version 0.2.2`, `status mvp`, `maturity_tier tested`, script/tests/demo all true, with a real `demo_command`. Running `build_catalog()` against the same tree produces `planned` / `spec-only` / all-false / `demo_command: null`, i.e. exactly the values that were asked to be corrected.

Causes, all in `generate_catalog.py`:

- `has_script` scans `rglob("*.py")` (line 441)
- `has_tests` requires `tests/test_*.py` (line 448), so `tests/run_test.sh` is invisible
- `demo_command` falls back to a `.py` script (line 455)
- `status` comes from a hardcoded `MVP_FOLDERS` set (line 463)

Two consequences beyond the obvious:

1. **The correction is not durable.** The catalog is generated and a bot writes it on a schedule, so the next run silently reverts a contributor's fix and the PR review that asked for it.
2. **`MVP_FOLDERS` is unfalsifiable in the other direction too.** Its own comment defines membership as "have working Python", yet `claw-semantic-sim` is in the set and has no Python at all (see #353). So the hand-maintained `status` field says `mvp` while the evidence-derived `maturity_tier` says `spec-only`, in the same object.

Suggested fix: derive `has_script` from any executable entry point (`*.py`, `*.R`, `*.sh`), `has_tests` from `tests/test_*.py` **or** `tests/run_test.sh`, and `demo_command` from a documented field in SKILL.md frontmatter rather than by globbing. Then delete `MVP_FOLDERS` and compute `status` from the evidence, so the two fields cannot disagree.

Also worth a non-zero-size check on test files while in there: the glob currently grants the `tested` tier to a zero-byte `test_*.py`, which is live for all skills.

manuelcorpas:
Closes #359. Addresses the remaining half of #360.

Both issues have one root cause: the catalog generator and CI each derived every signal from Python file patterns, and neither reported honestly when that assumption failed.

## #359, the catalog generator

`has_script` scanned `rglob("*.py")`, `has_tests` required `tests/test_*.py`, and `demo_command` fell back to a `.py`. So a skill written in any other language could never register above `spec-only`. `claw-amplicon-qc` (#329) ships a 1,274-line R script, a 17-assertion shell harness and a working `--demo`, and generated as planned / spec-only / all-false / `demo_command: null`. A scheduled job regenerates this file, so it also silently reverted that contributor's hand-corrected entry.

Replaced with `detect_entry_point` / `detect_tests` / `demo_invocation`: recognise `.py`, `.R` and `.sh` with deterministic precedence, and render the correct interpreter. A `.R` entry point recorded as `python foo.R` is a documented command that cannot run.

`detect_tests` also now requires a non-empty file. The old bare glob granted the `tested` tier to `my-awesome-skill`'s zero-byte `test_my_skill.py` (#345).

**MVP_FOLDERS contradicted itself.** Its own comment defines membership as "have working Python", yet it contained `claw-semantic-sim` (SKILL.md only, the #353 case) and `drug-photo` (no script). Both removed, and two tests now make the contradiction impossible.

**Deliberately not done:** the issue proposed deriving `status` from evidence and deleting `MVP_FOLDERS`. Measuring first: 62 skills ship code and are not in that set, so `status = "mvp" if has_script` would reclassify 62 public entries. "mvp" evidently means curated and supported, not merely present. The set stays hand-maintained, with a test that it cannot disagree with the evidence. Happy to revisit as its own decision.

## #360, CI

**Correcting my own issue: half of it was already fixed.** `33f98be` (2026-08-11) already picks the runner from what `tests/` contains, so the pytest-exits-5 problem I described was not live. I have corrected the issue.

What remained: a shell harness was reported `NOT VERIFIED` and never actually ran, so #329 went five review rounds with its 17 assertions never executing anywhere but the contributor's laptop. Honest, but honest-and-unverified is still unverified.

New `skill-harness` job provisions the skill's **own** `environment.yml` via micromamba, so it works for any future non-Python skill without further edits. Gated on a PR actually touching such a skill, because a Bioconductor solve costs minutes and most PRs need none of it.

## Validation, per the gate rules

Fires on exactly **1 of 97** skills, and it is the right one. No false positives.

And on its first real run against #329 it found something five review rounds missed: **that environment.yml has never been buildable.** It pins `r-base>=4.4`, while the bioconda builds of `dada2`, `ShortRead` and `Biostrings` resolve only against `r-base 4.1.3`. Dropping the pin solves cleanly; keeping it cannot. Reported to #329 separately.

A missing `environment.yml` is deliberately **not** an error, since a harness can legitimately be pure bash. Accusing a contributor of a missing file their skill may not need is the expensive mistake.

## Blast radius

Catalog diff against main is 4 fields:

| Skill | Field | Before | After |
|---|---|---|---|
| claw-semantic-sim | status | mvp | planned |
| drug-photo | status | mvp | planned |
| xena-tcga-gene-query | has_demo | false | **true** |
| xena-tcga-gene-query | demo_command | null | `python skills/.../query_tcga_api.py --demo` |

The xena change is a fix, not a regression: its script lives in `scripts/`, which the old root-only `select_demo_script` glob missed while `rglob` set `has_script: true`. I confirmed that script genuinely accepts `--demo` before advertising it.

That regression was caught by diffing the regenerated catalog against main, not by my fixtures, which is the argument for doing it.

## Tests

13 new tests covering entry-point detection across three languages, subdirectory scripts, empty and whitespace-only test files, interpreter selection, and the MVP invariant. 30 catalog tests and 191 repo-wide, all green.

---

> [!NOTE]
> **Medium Risk**
> Touches CI (new micromamba job that executes contributor harnesses and treats unsolvable environment.yml as failure) and public catalog status/demo fields. Not auth or data-handling, but a failed env solve or harness can now block PRs.
> 
> **Overview**
> Makes catalog maturity and CI verification language-agnostic so R/shell skills can be marked tested and actually run, instead of silently staying spec-only / NOT VERIFIED.
> 
> Catalog generation now finds `.py` / `.R` / `.sh` entry points (including nested `scripts/`), counts non-empty pytest, R, and shell tests, and emits the matching interpreter in `demo_command`. Empty test files no longer grant the `tested` tier. `claw-semantic-sim` and `drug-photo` are dropped from `MVP_FOLDERS` so `status: mvp` cannot sit next to `spec-only`; `xena-tcga-gene-query` now advertises its nested `--demo`.
> 
> CI defers shell harnesses from the pytest job to a new PR-gated `skill-harness` job that solves each skill’s `environment.yml` with micromamba (or runs bare if none) and fails if the env cannot be built.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b0df11ce7158063ae4f96cdd60f8d7b17eff96f0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


manuelcorpas:
Pushed a follow-up fix to a bug in this PR's own new job, found before merging rather than by a contributor.

The detect step ran `find "skills/$skill/tests"` under `set -euo pipefail`. When that directory does not exist, find exits non-zero, pipefail propagates it through the command substitution, and `set -e` terminates the step. Traced locally: the loop exits 1 and never reaches its end. Any skill without a `tests/` directory would therefore have failed `skill-harness` on most PRs, which would have been a maintainer-side red tick on contributors' work, the exact failure class this PR exists to reduce.

Guarded with `[ -d "skills/$skill/tests" ] || continue` and re-verified across all three cases: no tests directory survives the loop, a Python-only skill is correctly ignored, and a shell harness is correctly selected.

One honest limit on the validation. CI proves the **skip** path (the job runs, detects nothing, exits clean in 7s). The **run** path is proven only locally, where it correctly failed to solve #329's environment. Nothing in CI has exercised a real harness yet, because no open PR touching an R skill has run against this workflow. #329's next push will be the first genuine test, and I would rather say that than imply more coverage than exists.

PR #364, opened 2026-08-26
feat(skills): completeness-aware-caller — three-state gene calls for incomplete genomes

Qihao-Duan:
## What this adds

`completeness-aware-caller` — a skill that decides whether an assembly's completeness supports a conclusion, instead of letting "not found" silently become "absent".

- Given a completeness estimate (or a `busco-assessor` `result.json`) and gene-search results, it returns three-state calls: `present` / `absent` / `cannot_conclude`. Absence may only be asserted at ≥95% completeness; below that it refuses, with the miss probability spelled out.
- Given two ANI values, it gates the ranking: the margin must exceed 2× the drift that incompleteness alone induces. The drift coefficient is calibrated — not validated — on the benchmark below, and the SKILL.md says so explicitly.
- **Explicitly out of scope** (per the Scope section): measuring completeness, computing ANI, gene search, assembly, and contamination assessment. It chains after `busco-assessor` and consumes FastANI values; run CheckM upstream when contamination is in doubt.

## Evidence

Truth-known degradation benchmark: *P. phymatum* STM815 (complete genome, all symbiosis genes annotated) cut into 175 × 50 kb windows (seed 42) and retained at 100/90/70/50%, with real BUSCO 6.0.0 and FastANI 1.34 runs at every level. At 70% retention every copy of `nodB/nodC/nodS/nodU` falls into discarded windows and a naive caller reports a textbook symbiont unable to nodulate; this skill returns CANNOT CONCLUDE instead. Raw evidence, machine-readable ground truth, and a scoring script live at https://github.com/Qihao-Duan/completeness-aware-caller.

## Also in this PR

- **One-line fix to `busco-assessor`**: it builds its BUSCO command with `--out-path`, but BUSCO 6 accepts only `--out_path`, so every real (non-demo) invocation fails with `unrecognized arguments`. The bundled demo never invokes the binary, which is why it went unnoticed; found by running the skill for real during the benchmark. Happy to split this into its own PR if preferred.
- `skills/catalog.json` regenerated on top of current main (`scripts/generate_catalog.py`, 97 → 98).

## Checks

- 12 tests (red/green TDD per CONTRIBUTING); stdlib only; offline `--demo` exercises the refusal path (`python3 skills/completeness-aware-caller/completeness_aware_caller.py --demo --output /tmp/cac_demo`).
- SKILL.md follows `templates/SKILL-TEMPLATE.md`: trigger fire/do-not-fire lists, one-skill-one-task scope, numbered workflow, rendered example output, 5 gotchas, safety disclaimer, agent boundary, chaining partners, maintenance, citations.

Context: built at the Berlin hackathon (Open Challenge track). Follow-up to Manuel's review note on naming and scope — the name is settled as `completeness-aware-caller`, and the unimplemented genome-quality-assessor proposal has been removed from the fork.

### DIFF
--- a/skills/busco-assessor/busco_assessor.py
+++ b/skills/busco-assessor/busco_assessor.py
@@ -292,7 +292,7 @@ def build_busco_command(
         "-i", str(args.input),
         "-m", args.mode,
         "-c", str(args.cpu),
-        "--out-path", str(run_dir.parent),
+        "--out_path", str(run_dir.parent),
         "--out", run_dir.name,
         "-f",
     ]
--- a/skills/catalog.json
+++ b/skills/catalog.json
@@ -1,7 +1,7 @@
 {
   "version": "1.0.0",
   "generated_by": "scripts/generate_catalog.py",
-  "skill_count": 97,
+  "skill_count": 98,
   "galaxy_tool_count": 8182,
   "skills": [
     {
@@ -1056,6 +1056,52 @@
         "structural variant interpretation",
         "ClinGen dosage sensitivity",
         "deletion duplication pathogenic"
+      ],
+      "chaining_partners": [],
+      "license": "MIT",
+      "model_license": "",
+      "data_license": ""
+    },
+    {
+      "name": "completeness-aware-caller",
+      "cli_alias": null,
+      "description": "Three-state gene-presence calls (present / absent / cannot-conclude) and ANI-ranking gates for incomplete genomes and MAGs. Refuses to call a gene absent, or to rank references by ANI, when assembly completeness cannot support the claim.",
+      "version": "0.1.0",
+      "status": "planned",
+      "maturity_tier": "tested",
+      "maturity_evidence": {
+        "has_skill_md": true,
+        "has_script": true,
+        "has_tests": true,
+        "has_demo": true,
+        "cli_registered": false,
+        "ci_tested": false,
+        "benchmark_validated": false
+      },
+      "has_script": true,
+      "has_tests": true,
+      "has_demo": true,
+      "demo_command": "python skills/completeness-aware-caller/completeness_aware_caller.py --demo",
+      "dependencies": [
+        "python>=3.10"
+      ],
+      "tags": [
+        "comparative-genomics",
+        "completeness",
+        "metagenomics",
+        "mag",
+        "ani",
+        "abstention"
+      ],
+      "trigger_keywords": [
+        "is this gene really absent",
+        "completeness-aware",
+        "can I trust this MAG",
+        "gene absence incomplete genome",
+        "cannot conclude",
+        "abstention genomics",
+        "MAG interpretation",
+        "ANI confidence"
       ],
       "chaining_partners": [],
       "license": "MIT",
--- a/skills/completeness-aware-caller/SKILL.md
+++ b/skills/completeness-aware-caller/SKILL.md
@@ -1,0 +1,262 @@
+---
+name: completeness-aware-caller
+description: >-
+  Three-state gene-presence calls (present / absent / cannot-conclude) and
+  ANI-ranking gates for incomplete genomes and MAGs. Refuses to call a gene
+  absent, or to rank references by ANI, when assembly completeness cannot
+  support the claim.
+license: MIT
+metadata:
+  version: "0.1.0"
+  author: Qihao Duann
+  domain: genomics
+  tags:
+    - comparative-genomics
+    - completeness
+    - metagenomics
+    - mag
+    - ani
+    - abstention
+  inputs:
+    - name: completeness
+      type: value
+      format:
+        - float
+      description: >-
+        Assembly completeness as fraction (0-1] or percent (1-100], or via
+        --busco-json pointing at a busco-assessor result.json (required
+        unless --demo)
+      required: true
+    - name: genes
+      type: file
+      format:
+        - json
+      description: 'Gene search results: [{"gene": "nifH", "found": false}, ...]'
+      required: false
+  outputs:
+    - name: report
+      type: file
+      format: md
+      description: Three-state call table with rationale and refusal messages
+    - name: result
+      type: file
+      format: json
+      description: Machine-readable calls, gate decision, and thresholds used
+  dependencies:
+    python: ">=3.10"
+    packages: []
+  demo_data:
+    - path: examples/demo_genes.json
+      description: STM815 frag70 gene set (nod cluster undetectable; nifH survives on a 2nd copy)
+  endpoints:
+    cli: python skills/completeness-aware-caller/completeness_aware_caller.py --busco-json {busco_result} --genes {genes_json} --output {output_dir}
+  openclaw:
+    requires:
+      bins:
+        - python3
+    always: false
+    emoji: "🚦"
+    homepage: https://github.com/ClawBio/ClawBio
+    os:
+      - darwin
+      - linux
+    trigger_keywords:
+      - "is this gene really absent"
+      - "completeness-aware"
+      - "can I trust this MAG"
+      - "gene absence incomplete genome"
+      - "cannot conclude"
+      - "abstention genomics"
+      - "MAG interpretation"
+      - "ANI confidence"
+---
+
+# 🚦 Completeness-Aware Caller
+
+You are the **completeness-aware-caller**, a ClawBio skill that separates
+taxonomic confidence from functional confidence in incomplete genomes. A
+fragmented genome makes missing genes look absent and close relatives look
+functionally identical; this skill's deliverable is the refusal — an explicit
+`CANNOT CONCLUDE` printed as a first-class result, not a footnote.
+
+## Trigger
+
+**Fire when the user says any of:**
+- "is this gene really absent", "gene absent or just missing"
+- "can I trust this MAG", "MAG completeness interpretation"
+- "completeness-aware call", "abstention", "cannot conclude"
+- "is the ANI difference meaningful", "ANI confidence"
+- "does incompleteness affect my conclusion"
+
+**Do NOT fire when:**
+- User wants to *measure* completeness → route to `busco-assessor`
+- User wants to *compute* ANI or relatedness → run FastANI (or
+  `galaxy-bridge`), then come back here with the values
+- User wants variant-level interpretation → `clinical-variant-reporter`
+- User wants microbiome community profiling → `claw-metagenomics`
+
+## Why This Exists
+
+- **Without it**: pipelines print "gene absent" from 70%-complete MAGs and
+  rank references by ANI margins smaller than the noise incompleteness
+  induces. Absence of evidence is silently converted into evidence of
+  absence.
+- **With it**: every absence claim and every ANI ranking carries either a
+  confidence grounded in measured completeness, or an explicit refusal with
+  an actionable message.
+- **Why ClawBio**: chains directly after `busco-assessor` (reads its
+  `result.json` verbatim) and complements `galaxy-bridge` ANI workflows.
+
+## Scope
+
+One skill, one task: **decide whether completeness supports a claim**. This
+skill does not measure completeness, compute ANI, search for genes, or
+assemble genomes — and it does not assess contamination: a `present` call
+assumes the contig truly belongs to the genome (Gotcha 5), so run CheckM
+upstream when that is in doubt. It consumes those results and gates the
+conclusions.
+
+## Domain Decisions
+
+1. **Absence needs ≥95% completeness** (`MIN_COMPLETENESS_FOR_ABSENT =
+   0.95`). Under a random-loss model, P(gene missed | truly present) ≈
+   1 − completeness; 0.95 caps that at 5%. Anchored to MIMAG quality tiers
+   (Bowers et al. 2017) where "high-quality draft" starts at >90%.
+2. **ANI drift model**: drift = 0.82 × (1 − completeness) ANI points,
+   calibrated on the *P. phymatum* STM815 degradation benchmark (0.41 points
+   of observed drift at 50% retention against *P. xenovorans* LB400, 0.44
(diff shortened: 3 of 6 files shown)


manuelcorpas (changes requested):
## pr-audit: changes requested

This is a thoughtful skill and the safety architecture is genuinely good. The SKILL.md volunteering that the drift coefficient is calibrated rather than validated is the kind of honesty this repo should want more of, and the calibration numbers check out: I traced `DEMO_ANI = (81.45, 80.75)`, the 81.28/80.58 test pair, the 0.41/0.44 drift figures, `DRIFT_COEFF = 0.82` and the `C:68.1%` figure back to the deposited frag70/frag50/frag100 runs, and every one matches. No fabricated data anywhere.

Three paths, though, emit a confident call where only "cannot conclude" is supportable, which is precisely what this skill exists to prevent.

### 1. CI is red, and the fix is one line

`SKILL.md:47`

```yaml
    packages: []
```

`skill-lint` fails with `Invalid YAML in frontmatter ... Found ugly disallowed JSONesque flow mapping` at line 47, column 15. PyYAML parses this fine, which is why it was easy to miss; the pinned `strictyaml` validator rejects inline flow style. Use an empty block or quote it as `'[]'`. This is the third time this class has bitten the repo and it is partly our fault, see the note at the end.

### 2. CRITICAL, `completeness_aware_caller.py:62-70`

BUSCO completeness is a statistic over chromosomal single-copy core orthologues. It carries almost no information about plasmid or accessory content, yet it is used directly as the per-gene miss probability for *any* gene:

```python
if completeness >= min_completeness:
    return {"status": "absent", "confidence": completeness, ...}
```

A MAG bin that drops an entire plasmid still returns C of roughly 98%, and the skill will then confidently call every plasmid-borne gene absent. This PR's own demo genes (`nodB`, `nodC`, `nifH`) are all plasmid-borne on NC_010627.1, so the demo set is exactly the class the model cannot cover. Gotcha 4 covers within-assembly clustering and Scope defers contamination to CheckM, but nothing guards replicon-level dropout. Suggest a scope caveat at minimum, and ideally a refusal when the gene's replicon is unknown.

### 3. CRITICAL, `completeness_aware_caller.py:44` and `:197`

`parse_completeness` guesses units, so a genuine BUSCO percentage at or below 1.0 is silently read as a fraction:

```python
if 1.0 < c <= 100.0:
    c /= 100.0
```

`busco-assessor` writes `C` as a percentage (`busco_assessor.py:366`, and its own fixture at line 149 carries `"C": 2.1`), so sub-1% C is an expected value in this codebase. A 0.96%-complete assembly becomes 96% complete and yields a confident `absent` on every gene. On the `--busco-json` path the units are known, so divide by 100 unconditionally rather than guessing.

### 4. CRITICAL, `completeness_aware_caller.py:89-95`

The ANI gate fails open at high completeness:

```python
drift = drift_coeff * (1.0 - completeness)
...
if margin >= safety * drift:
    decision = "rank"
```

At C=1.0, which is reachable from a real `result.json`, drift is 0.0 and `margin >= 0.0` is true for a margin of exactly zero, so two identical ANI values are reported as "ranking is supported". At C=0.98 the gate passes a 0.04-point margin, well inside FastANI's own estimator variance, which the model never accounts for. Suggest flooring the gate at FastANI's precision rather than letting it approach zero.

### Warnings

- `tests/test_completeness_aware_caller.py:111` is tautological: `assert data["ani_gate"]["decision"] in ("rank", "cannot_conclude")` accepts both possible values. The demo actually returns `rank` (margin 0.70 against a gate of 0.492 at C=0.70), so the headline refusal demo quietly ranks two references and the test declines to pin it.
- No test exercises `--busco-json` at all, although SKILL.md:107-108 calls it the primary integration. No ANI-gate test uses completeness above 0.90 either, which is why finding 4 is invisible.
- The `busco-assessor` one-liner (`--out_path`, `busco_assessor.py:295`) is a genuine and valuable fix, since `--out-path` would be rejected by BUSCO's argparse and every real invocation was broken. It has no test: `TestBuildBuscoCommand` asserts `-m`, `-l`, `--auto-lineage-prok` and `-c` only. Happy to take it as its own PR with a test if you would rather split it out.
- `catalog.json:1084` omits `--output` from the demo command, which is `required=True` at `completeness_aware_caller.py:233`, so the demo as catalogued exits 2. SKILL.md's own examples are correct.
- `present` is hardcoded to confidence 1.0 with no identity or coverage gating, so a paralogue or partial hit scores as certainty in machine-readable output. Documented in Gotcha 5, but a literal 1.0 will be consumed downstream at face value.
- Gotcha 4 says to report the residual as a modelled bound and never as a measured rate, but the emitted report line states it flatly: `residual chance of a missed gene is ~{miss_probability:.1%}`. The caveat never reaches the rendered `report.md`.
- The calibration benchmark is real and I verified it, but it is not in the Citations list and the URL appears only in the PR body, unpinned to a commit or DOI. Worth pinning.

### Not your fault

The `packages: []` failure is the third occurrence of the strictyaml flow-mapping class in this repo, and `docs/tutorial-build-a-skill.ipynb` still teaches the old flat frontmatter schema that triggers it. We are fixing the tooling side.

manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  
Static scan: 0 critical, 2 warning. Domain review: complete.

## Domain review: NEEDS REVIEW

Thanks for this. The core idea is sound and the write-up is unusually honest: `SKILL.md:125` states outright that the 0.82 drift coefficient is calibrated on the STM815 benchmark rather than validated against it, and that the reference ranking never actually failed in that benchmark. Gotcha 4 at `SKILL.md:204` correctly refuses to sell `1 - C` as a measured miss rate and explains the clustered-loss counter-example. `call_gene` at `completeness_aware_caller.py:53` is the right shape: detection is positive evidence, non-detection is only absence above threshold, everything else abstains. The 95% floor at `completeness_aware_caller.py:28` sits above the MIMAG high-quality draft bar and the three citations at `SKILL.md:253` all resolve to the papers claimed (Bowers 2017 MIMAG, Jain 2018 ANI species boundaries, Manni 2021 BUSCO). Nothing in the module reaches the network: the imports at `completeness_aware_caller.py:22` are argparse, json, sys, datetime and pathlib only, and all output is written locally by `write_outputs` at `completeness_aware_caller.py:119`.

### Must fix before merge: the busco-assessor handoff can invert the safety guarantee

`completeness_aware_caller.py:197` passes the BUSCO `C` field straight into `parse_completeness`, which auto-detects units at `completeness_aware_caller.py:44`. In a busco-assessor `result.json` the `C` field is always a percentage; `busco_assessor.py:366` writes it as `float(c)` parsed from the short summary. So a genuinely near-empty assembly is read as a near-complete one:

- `C: 1.0` (1% complete) fails the `1.0 < c` test, is not divided, and returns 1.0. `call_gene` then reports every unfound gene as **absent** with confidence 1.00 and a residual miss chance of "~0.0%".
- `C: 0.9` returns 0.90, and the ANI drift at `completeness_aware_caller.py:89` collapses to 0.07 points, so the ranking gate passes almost anything.

This is not a hypothetical band. busco-assessor ships its own low-completeness example at `busco_assessor.py:149` with `C: 2.1` for a mitochondrial assembly scored against a nuclear lineage; a wrong lineage dataset or a small MAG lands under 1% routinely. The failure is silent and it points the wrong way: the one skill whose deliverable is the refusal would emit a maximally confident false absence. Please divide by 100 unconditionally on the `--busco-json` path, where the unit is known, and consider rejecting the overlapping (0, 1] band on `--completeness` or requiring an explicit unit flag. The current behaviour also contradicts the workflow text at `SKILL.md:142`, which says values outside the two ranges are rejected but says nothing about which reading wins inside the overlap.

### CI is red

`skill-lint` failed on 63411598. The repo runs `agentskills validate` per skill directory at `.github/workflows/ci.yml:73`, and that validator is not vendored in this checkout, so I cannot confirm the exact rule from here. The likely cause is the inline flow sequence `packages: []` at `SKILL.md:47`: no other SKILL.md in the tree uses that form, the convention is a bare `packages:` (see `skills/cnv-acmg-classifier/SKILL.md:53`) or `packages: null` (see `skills/seq-wrangler/SKILL.md:36`). Worth trying that first.

### Calibration questions only you can answer

1. **Regime transfer.** `completeness_aware_caller.py:186` records the calibration pair at 81.45 and 80.75 ANI. FastANI does not consider values below roughly 80% reliable, so the coefficient was fitted at the very edge of the tool's usable range, and it is then applied as a flat number of ANI points at every ANI level, including the 94-96% boundary zone handled at `completeness_aware_caller.py:91` and the 99%-plus comparisons where real ranking decisions are made and margins are hundredths of a point. Is there any evidence drift in ANI points is level-independent? If not, the gate is probably too loose where it matters most.
2. **Which observation set the coefficient.** `SKILL.md:125` reports 0.41 points of drift against LB400 and 0.44 against J2315 at 50% retention, and 0.82 reproduces the 0.41 figure exactly. Taking the smaller of the two makes the noise model marginally anti-conservative before the safety factor is applied. Please say so explicitly, or fit to the larger value.
3. **Benchmark provenance.** Neither the degradation benchmark nor the script that produced the frag50/frag70/frag90 numbers is in this PR, so the numbers at `completeness_aware_caller.py:186` and the marker-loss observations at `SKILL.md:127` cannot be reproduced or checked by anyone reading the repository. A fixture or a pointer to the data would fix that.
4. **Unverifiable organism claim.** The demo comment at `completeness_aware_caller.py:175` attributes the surviving nifH to two copies on pBPHY02. I could not verify the replicon assignment from anything in this repository, and published descriptions of P. phymatum STM815 place the symbiosis genes on the larger plasmid. Please add a citation or correct it. This affects narrative only, not the call logic.

### Test gaps

The unit tests are real, not written to match the implementation: `test_completeness_aware_caller.py:64` exercises the refusal path with the benchmark margin and `test_completeness_aware_caller.py:115` drives the CLI end to end. Three gaps:

- `--busco-json` has no test at all, despite being the documented primary entry point at `SKILL.md:137`. A fixture in busco-assessor's `result.json` shape, including a low `C` such as 0.9, would have caught the unit bug above.
- `test_completeness_aware_caller.py:112` asserts `data["ani_gate"]["decision"] in ("rank", "cannot_conclude")`. Those are the only two values the function can return, so the assertion cannot fail. The demo deterministically yields `rank`; assert that.
- `test_completeness_aware_caller.py:51` covers 72.5 and 0.725 but nothing in the ambiguous (0, 1] band, which is exactly where the parser misbehaves.

### busco-assessor flag change

`busco_assessor.py:295` moving from `--out-path` to `--out_path` is correct. BUSCO 5 spells this option with an underscore, consistent with `--download_path` already used two lines further down. `skills/busco-assessor/SKILL.md:156` still documents the hyphenated form; please update it in the same PR.

### Static scanner recheck

- CI failure warning: accurate, `skill-lint` is FAILURE on this head commit.
- `packages: []` at `SKILL.md:47`: the line citation is exact and the concern is consistent with repo convention. The specific claim about strictyaml rejecting the flow sequence is plausible but unproven from this checkout.
- Three `subprocess` INFO items in the test file: accurate and benign. Each invokes the skill's own script through `sys.executable` with a list argv and no shell.

No prompt-injection text, no secrets, no telemetry and no patient or sequence data leaving the machine. The second-person framing in the SKILL.md body matches the convention used across the other skills and is not addressed at a reviewer.

Holding for the unit bug on the busco-assessor handoff, the red lint, and the calibration questions. The refusal logic itself is the right design.

---

manuelcorpas:
Re-read at head 63411598. No commits since the 26 August review, the branch now conflicts with main on `skills/catalog.json` (both sides edit `skill_count`; the right value after regeneration is 97), and skill-lint is red on `packages: []` at SKILL.md:47, which strictyaml rejects (write it as a block list or drop the key).

The three points from 26 August stand, and I want to be precise about why the first is a safety problem rather than a style one.

1. **Unit guess inverts the guarantee.** `completeness_aware_caller.py:43-45` does `if 1.0 < c <= 100.0: c /= 100.0`. `run()` at 197 passes `busco["C"]` straight in, and busco-assessor writes `C` as a percentage (`busco_assessor.py:366`, and its own fixture has `"C": 2.1`). So a 1.0% complete assembly is read as fully complete, and every gene not found is called `absent` with confidence 1.00. Take the input in one declared unit and refuse anything else.
2. **ANI gate fails open at C = 1.0.** Line 89 `drift = 0.82 * (1 - completeness)` is zero, so line 95 `margin >= safety * drift` is `0 >= 0` and identical ANIs return `rank`. Add an estimator-variance floor.
3. **Plasmid genes under chromosomal BUSCO.** Line 61 uses `1 - C` as the per-gene miss probability, and the demo genes are plasmid-borne, which BUSCO cannot see. A replicon guard or a scope statement in the refusal text.

Smaller, from this read: the 0.82 coefficient is derived from your unpublished benchmark and cited nowhere in the Citations block; `SAFETY_FACTOR`, the 94 to 96 boundary window and the 0.90 gate have no source; `tests:111` `assert decision in ("rank", "cannot_conclude")` cannot fail; the SKILL.md example output (68.1%, nifH CANNOT CONCLUDE) is not what the demo produces (0.70, nifH PRESENT); the `--out_path` fix in busco-assessor is correct but untested and `skills/busco-assessor/SKILL.md:156` still documents the old flag.

Happy to look again as soon as a rebase lands with 1 and 2 fixed.

manuelcorpas:
Closing for now: no push since 26 August, the branch conflicts with main on skills/catalog.json, and skill-lint is red on the flow-style `packages: []`. The two safety items from the 26 August and 8 September reviews still stand: the unit guess at completeness_aware_caller.py:43-45 reads a 1% complete assembly as fully complete, and the ANI gate fails open at C = 1.0. Three-state calls for incomplete genomes are worth having, so if you return to this, reopen or open a fresh PR with those two fixed and a rebase, and it will be looked at straight away. Thank you for the work.

PR #366, opened 2026-08-26
docs: disclose Nebius report upload behavior

### LINKED ISSUE #338: examples/nebius_agent.py: disclose that skill report text is sent to a third-party inference provider
From the audit of #336 (merged).

`examples/nebius_agent.py:230` posts each skill report verbatim into the conversation sent to Nebius, up to 6000 characters per report (`:108`).

The local-first guarantee is real and I verified it: every whitelist entry pins `"args": ["--demo"]` (lines 40, 50, 59, 68) and `run_skill` builds `cmd = [sys.executable, str(script), *spec["args"], "--output", td]`, so the model cannot supply a path and only repo-bundled demo data can ever run. No `--input` reaches the subprocess. That is why this was not a blocker.

Two gaps remain:

1. **No disclosure in-repo.** Nothing in the file docstring or anywhere else states that skill output is transmitted to a third-party inference provider. `nebius_agent.py` is the only file in the tree mentioning Nebius. What goes out is not innocuous-looking either: running `clinical_variant_reporter --demo` produces a report containing an absolute local filesystem path, uploaded alongside the variant table.

2. **It is an example.** A hackathon attendee copying this and adding a real VCF path to `args` breaks the guarantee, and the file carries no warning against exactly that edit. One comment beside the whitelist would close it.

Suggested: a short docstring paragraph saying report text is sent to Nebius and that the `--demo` pinning is why no patient data can be, plus an inline warning at the whitelist.

Worth landing before the docs pages publish for the 18 August Berlin hackathon.

krudo-taco:
## Summary

- Disclose that live runs send ClawBio skill report text to Nebius as tool output.
- Explain that the current whitelist pins every skill to `--demo`, so patient input paths cannot reach subprocesses.
- Warn contributors not to add real patient input paths to the whitelist.

Fixes #338.

## Skill affected

Example script only: `examples/nebius_agent.py`

## Checklist

- [x] No `SKILL.md` change is required for this example-only update
- [x] Validation included and passing (`py_compile`; no runtime behavior changed)
- [x] Existing demo data remains available

## Test output

```text
python3 -m py_compile examples/nebius_agent.py
# passed

git diff --check
# passed
```


manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Thanks for this. Documenting an egress path is the right instinct, and it is the kind of disclosure that is easy to leave out.

**What I checked.** The diff is seven added lines, all of them prose: a docstring paragraph at `examples/nebius_agent.py:18-21` and two comment lines above the whitelist at `examples/nebius_agent.py:43-44`. No executable statement changes. There is no clinical logic here, no source vocabulary matched against ClinVar, VEP, gnomAD, CPIC or PharmGKB, no citations, no new network call, and no test to assess. The added text speaks to future contributors, not to any reviewer or agent, and it makes no claim of approval, so I did not treat it as injection.

**Why not SAFE.** The new text asserts two security properties that I could not confirm from the material available to me. The docstring at `examples/nebius_agent.py:18-21` states that report text is sent to Nebius as tool output and that this only happens when the script is run without `--dry-run`. The comment at `examples/nebius_agent.py:43-44` states that the whitelist pins every skill to `--demo`, and the docstring concludes from that premise that patient input paths cannot reach the subprocesses. The diff context stops at the `SKILLS` opening brace on `examples/nebius_agent.py:45` and the first key on `:46`, so none of the per-skill `args` are visible, and neither is the `--dry-run` branch or the call that forwards the report. `examples/nebius_agent.py` is also absent from the checkout of `main` I was given, so I could not read the surrounding file to verify either claim. A disclosure that understates egress is worse than none, so this needs the author or a maintainer to confirm against the real file that every entry in `SKILLS` is demo-pinned and that no path outside `--dry-run` sends more than the demo report.

**One design point.** As written, the demo-only rule at `examples/nebius_agent.py:43-44` is a comment, so nothing enforces it. A later PR that adds one input path argument to an entry in `SKILLS` makes the docstring at `:20-21` silently false while the disclosure still reads as a guarantee. Consider a check at load time over `SKILLS` that requires `--demo` in every entry's args and refuses to start otherwise. That turns the promise into an invariant and keeps the docstring honest without further review effort.

**On the static scan.** Its single warning is that CI is awaiting maintainer approval and has not run. That matches the record for this PR: the status check rollup is empty and the merge state is `UNSTABLE`, so the warning is correct and not a false positive. The report also flags the change as out of bioscience scope on a zero keyword match. That heuristic is misfiring on a documentation edit; an example agent that runs ClawBio genomics skills is plainly in scope for this repository, and scope is not a reason to hold this.

---

krudo-taco:
Confirming the two claims against the file on this branch:

- Every SKILLS entry is args: ["--demo"] (rare_high_impact_variants, vcf_annotator, clinical_variant_reporter, gwas_prs). None include --input.
- `--dry-run` never sends a report to Nebius. The live path is the chat loop after that flag is absent; `run_skill()` returns report text that becomes tool output.

I also turned the demo-only comment into an import-time invariant: `_require_demo_only_whitelist` refuses to start if any entry drops `--demo` or adds `--input`. Covered by `tests/test_nebius_agent.py`.

manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Thanks, this is a useful disclosure and the shape of the guard is right. Saying plainly in the module docstring that skill report text is sent to Nebius as tool output is exactly what an example that calls a hosted model should state. Enforcing the `--demo` pin at import time and again inside `run_skill` fails closed: a bad whitelist edit stops the example starting rather than quietly uploading patient content. No clinical or scientific logic is touched, so there is no path here to a wrong result.

This one is merging. The points below are for later and none of them block.

**Non-blocking suggestions**

1. The `--input` test in `_require_demo_only_whitelist` (examples/nebius_agent.py:91) matches an exact token, so `--input=patient.vcf` would slip through, as would other input flags such as `--vcf` or `--file`. An allowlist of permitted flags, or a prefix check, would close that gap.
2. tests/test_nebius_agent.py will not actually run in CI. The test job lists individual pytest paths (.github/workflows/ci.yml:116-147) and `tests/` is not among them, so the invariant is only enforced by the runtime guard. Adding a step, as #370 does for vcf-annotator, would make the test count.
3. skills/catalog.json:3043 carries the same one line nutrigx bump to 0.2.0 as #377, so whichever lands second will want a rebase. Worth confirming skills/nutrigx/SKILL.md already declares 0.2.0 at merge time, so the catalog does not advertise a version the skill does not yet have.

The static scan reported no findings on this PR, and I found nothing in the diff to contradict that.

---
Approved and merged.

PR #368, opened 2026-08-26
fix: fail closed on VEP annotation batch failures

### LINKED ISSUE #341: [Bug] clinical-variant-reporter: a failed VEP batch is silently reported as unannotated variants, which classify as ordinary VUS
### Summary

In live mode, `annotate_variants_vep` catches **every** exception per batch, prints a warning to
stderr, and appends empty `VariantEvidence` for all records in that batch. Those records then carry
no gene, no consequence, no ClinVar significance and no allele frequency, so every one of them
classifies as **Uncertain Significance** — and the run exits 0 with a `result.json` that looks
completely normal.

A transient network failure is therefore indistinguishable, in the output, from a batch of genuine
uncertain findings. With `VEP_BATCH_SIZE` at 200, one blip silently converts up to 200 variants.

### Where

`skills/clinical-variant-reporter/clinical_variant_reporter.py`, in `annotate_variants_vep`
(current `main`):

```python
except Exception as exc:
    print(f"WARNING: VEP batch failed ({exc}). Marking {len(batch)} variants as unannotated.", file=sys.stderr)
    for rec in batch:
        evidence_list.append(VariantEvidence(
            chrom=rec.chrom, pos=rec.pos, ref=rec.ref, alt=rec.alt,
            gene=rec.info.get("GENE", ""),
        ))
    continue
```

### Reproduction

Point `VEP_BASE_URL` at a closed port and classify any two variants:

```python
import os
os.environ["VEP_BASE_URL"] = "http://127.0.0.1:9/vep/homo_sapiens/region"   # discard port, closed

recs = [VcfRecord(chrom="11", pos=5227002, id="rs334", ref="T", alt="A", qual=50.0, filt="PASS", info={}),
        VcfRecord(chrom="7", pos=117606754, id=".", ref="G", alt="A", qual=50.0, filt="PASS", info={})]

for ev in annotate_variants_vep(recs):
    print(ev.chrom, ev.pos, classify_variant(ev).classification, classify_variant(ev).triggered_codes)
```

Observed:

```
WARNING: VEP batch failed (HTTPConnectionPool(... Connection refused)).
11 5227002 'Uncertain Significance' []
7  117606754 'Uncertain Significance' []
```

Both are well-known pathogenic variants — rs334 is the sickle-cell HBB allele and 7:117606754 G>A is
a CFTR canonical splice-donor variant — and both come back as ordinary VUS with **no triggered
criteria at all**. Exit code 0.

### Why this is worth raising

A count of zero triggered criteria is *suggestive* but not a reliable tell: a genuinely
unannotatable variant can produce the same shape, so a downstream consumer cannot use it to detect
the failure. There is no other signal in `result.json` — no per-variant annotation status, no
summary counter, no non-zero exit — so a caller has nothing to check.

Any pipeline that records a classification alongside its tool and database versions for later
reproduction is, in this case, recording a result that was never derived from the databases it names.

### Suggested fix

Fail the run rather than emitting evidence that was never annotated:

```python
except Exception as exc:
    raise RuntimeError(
        f"VEP annotation failed for a batch of {len(batch)} variants "
        f"(batch starting at index {batch_start})."
    ) from exc
```

`raise ... from exc` keeps the underlying cause diagnosable without interpolating a URL or a payload
fragment into a message that a caller may log.

**Dropping the failed batch and continuing is not an alternative**: the caller is handed a shorter
list with no way to tell which variants are missing, which trades a wrong answer for a silently
incomplete one.

If a partial result is wanted for large runs, that seems better expressed as an explicit opt-in
(`--allow-partial-annotation`) plus a machine-readable per-variant annotation status in
`result.json`, so the choice is the caller's and is visible in the output. Happy to send a PR for
either shape — say which you'd prefer.

### Notes

Found while running the skill against a self-hosted VEP endpoint. Reproduced 4/4.

krudo-taco:
Closes #341.

## Summary

- Raise `RuntimeError` when any VEP annotation batch fails instead of returning placeholder evidence that can be misclassified.
- Reject non-list or non-object VEP batch payloads before classification.
- Add regression coverage for complete success, total and partial batch failure, malformed response shape, and CLI propagation.
- Keep partial-result mode out of scope, as requested by the maintainer.

## Skill affected

`clinical-variant-reporter`

## Checklist

- [x] Existing demo data remains available

## Test output

- Red phase: new tests on the old implementation — `4 failed, 1 passed`
- `uv run --no-project --with pytest --with requests pytest skills/clinical-variant-reporter/tests/ -q` — 77 passed
- `python -m py_compile skills/clinical-variant-reporter/clinical_variant_reporter.py skills/clinical-variant-reporter/acmg_engine.py` — passed
- `git diff --check` — passed

The live Ensembl VEP API and the full repository test suite were not run. The focused suite uses mocked VEP responses and covers the failure boundary without sending patient data or depending on network availability.

### DIFF
--- a/skills/clinical-variant-reporter/clinical_variant_reporter.py
+++ b/skills/clinical-variant-reporter/clinical_variant_reporter.py
@@ -321,13 +321,14 @@ def annotate_variants_vep(
 ) -> tuple[list[VariantEvidence], dict[str, str] | None]:
     """Annotate variants via Ensembl VEP REST API.
 
-    Returns (evidence_list, source_versions). source_versions distinguishes
-    two failure states rather than collapsing both to None:
-      - None: no VEP batch succeeded this run (nothing was annotated), so a
-        report must not make any data-source claim at all.
-      - {} (empty dict): at least one VEP batch succeeded, but the separate
-        Ensembl data-source version lookup itself failed. Annotation is real;
-        the version check is what's missing.
+    Returns (evidence_list, source_versions). A VEP batch failure aborts the
+    run rather than returning placeholder evidence that can be misclassified.
+    source_versions still distinguishes successful annotation states:
+      - None: no annotation was attempted, so a report must not make any
+        data-source claim at all.
+      - {} (empty dict): annotation succeeded, but the separate Ensembl
+        data-source version lookup itself failed. Annotation is real; the
+        version check is what's missing.
       - a non-empty dict: the lookup succeeded (individual sources may still
         be absent from Ensembl's response).
 
@@ -359,13 +360,16 @@ def annotate_variants_vep(
             resp.raise_for_status()
             vep_results = resp.json()
         except Exception as exc:
-            print(f"WARNING: VEP batch failed ({exc}). Marking {len(batch)} variants as unannotated.", file=sys.stderr)
-            for rec in batch:
-                evidence_list.append(VariantEvidence(
-                    chrom=rec.chrom, pos=rec.pos, ref=rec.ref, alt=rec.alt,
-                    gene=rec.info.get("GENE", ""),
-                ))
-            continue
+            raise RuntimeError(
+                f"VEP annotation failed for a batch of {len(batch)} variants "
+                f"(batch starting at index {batch_start})."
+            ) from exc
+
+        if not isinstance(vep_results, list) or not all(isinstance(vr, dict) for vr in vep_results):
+            raise RuntimeError(
+                f"Malformed VEP response for a batch of {len(batch)} variants "
+                f"(batch starting at index {batch_start})."
+            )
 
         any_batch_succeeded = True
         result_map: dict[str, dict] = {}
--- a/skills/clinical-variant-reporter/tests/test_clinical_variant_reporter.py
+++ b/skills/clinical-variant-reporter/tests/test_clinical_variant_reporter.py
@@ -739,7 +739,7 @@ class TestAnnotateCapturesSourceVersions:
         assert called_url == VEP_REST_HOST + ENSEMBL_INFO_VARIATION_PATH
         assert captured_kwargs[0].get("timeout") == 10
 
-    def test_source_versions_none_when_every_batch_fails(self, monkeypatch):
+    def test_batch_failure_raises_before_version_fetch(self, monkeypatch):
         import requests
         from clinical_variant_reporter import annotate_variants_vep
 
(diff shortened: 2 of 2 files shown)


manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

## Domain review: clinical-variant-reporter

Thanks for this. It closes a real silent-failure path, and the tests prove the behaviour rather than restating it.

**What is right**

- The old handler at the same location caught every exception, printed a warning and appended a bare `VariantEvidence` for each record in the failed batch. Those placeholders carry no consequence, no ClinVar significance and no gnomAD AF, so no ACMG criterion fires, `_aggregate` falls through to its final `return "Uncertain Significance"` (`acmg_engine.py:460`), and the variant lands in the VUS table of `report.md` and in `result.json` with `abstained: false`. A network timeout was therefore indistinguishable from a genuine VUS call. Raising at `clinical_variant_reporter.py:363` is the correct direction for a clinical reporter.
- The self-audit added earlier does not cover this case, so the fix is not redundant: `audit_classified` only checks identity mismatch, mutually exclusive criteria and provenance of *triggered* criteria (`self_audit.py:99`, `self_audit.py:121`, `self_audit.py:128`). A placeholder triggers nothing, so it passed the audit and was reported as a confident VUS.
- The shape guard at `clinical_variant_reporter.py:368` is worth having. Ensembl VEP REST returns a JSON array for `POST /vep/human/region`; a 200 carrying an object is not a result set, and previously it would have raised inside the `for vr in vep_results` loop or silently produced no matches.
- The docstring rewrite at `clinical_variant_reporter.py:324` now matches the code. With batch failure raising, `None` is reachable only when no batch ran at all, and `any_batch_succeeded = True` at `clinical_variant_reporter.py:374` is set after both guards, so a failed batch can never mark the run as annotated.
- The `from exc` chaining preserves the underlying `requests` error for the operator. Nothing is swallowed.

**Tests**

- `tests/test_clinical_variant_reporter.py:819` is the one that matters: batch size forced to 1, first batch succeeds, second fails, and the version fetch is asserted unreachable. That proves no partial annotation escapes.
- `tests/test_clinical_variant_reporter.py:863` exercises `main()` end to end and asserts at `:866` that no `result.json` was written. This is a real fail-closed assertion, not a vacuous one.
- The VEP vocabulary in the success fixture at `tests/test_clinical_variant_reporter.py:792` is genuine: `most_severe_consequence`, `transcript_consequences[].gene_symbol` and `consequence_terms` are real VEP REST fields, `missense_variant` and `synonymous_variant` are real Sequence Ontology terms, and the `input` echo `"1 100 100 A/G 1"` is the region format VEP returns. Nothing here was invented to fit the implementation.
- No clinical logic changed. No ACMG criterion, threshold or classification string was touched, and no ClinVar, gnomAD or CPIC literal was added or altered.

**Follow-ups, not blocking**

1. Pre-existing and outside this PR's stated scope: the per-variant miss path a few lines below `clinical_variant_reporter.py:375`, where `result_map.get(region, {})` returns empty, still appends the same unannotated placeholder and still classifies as VUS. An empty list from VEP passes the new guard at `:368` cleanly, and VEP does omit inputs it cannot map. The batch-level hole is closed; this per-variant one is not. Worth a separate issue.
2. `RuntimeError` propagates out of `main()` as an uncaught traceback. Other failures in this module report and exit deliberately, for example `clinical_variant_reporter.py:713`. A message plus `sys.exit(1)` would be more consistent, though the safety outcome is identical.
3. A single transient failure now discards every preceding batch. VEP REST returns 429 routinely at `VEP_BATCH_SIZE = 200`, so a retry with backoff on 429 and 5xx before aborting would make the strict behaviour practical on large panels. Aborting remains the right default.

**Static scanner**

Re-checked all four findings. They are INFO only, all four are `import requests` inside the new test bodies. Every one is monkeypatched at `tests/test_clinical_variant_reporter.py:795`, `:822`, `:838` and in the success test, and no test issues a live call. No false positives to record, and no critical or warning findings to contest.

No secrets, no telemetry, no patient data leaving the machine, no instruction-like text in the diff. CI is green on 3.11, 3.12 and 3.13.

---

Thanks for this. Replacing the silent placeholder path with a hard failure is the right direction for a clinical reporter, and the direction of travel is exactly what this skill needed. I confirmed the mechanics: `batch_start` is in scope at the raise site because it is the loop variable (`clinical_variant_reporter.py:363`, loop at `clinical_variant_reporter.py:284` on main), `from exc` preserves the cause, and the success-path fixture is genuine rather than written to fit the code. The `"input": "1 100 100 A/G 1"` echo in `test_complete_annotation_success_still_returns_evidence_and_versions` matches what `_vcf_to_vep_region` actually emits for an SNV (`clinical_variant_reporter.py:160` on main), so the gene assertion really does exercise the `result_map` join rather than passing on placeholders. The static scan reported no findings on this PR, so there is nothing for me to refute.

Two things to resolve before this merges.

**1. The new type guard leaves the exact hole the docstring says it closes.**

The guard at `clinical_variant_reporter.py:368` is `not isinstance(vep_results, list) or not all(isinstance(vr, dict) for vr in vep_results)`. An empty list passes both halves: `isinstance([], list)` is true, and `all()` over an empty iterable is true. A VEP 200 response of `[]`, which is what Ensembl returns when none of the submitted regions map, therefore flows straight through to the unchanged `result_map` lookup below the guard (main: `clinical_variant_reporter.py:307-320`), and every variant in the batch becomes an unannotated `VariantEvidence` placeholder. The same applies to a well-formed list that simply omits some inputs.

That matters clinically, because those placeholders are not inert. They trigger no ACMG criteria, so they classify as Uncertain Significance and appear in the VUS table indistinguishable from a real VUS with real gnomAD and ClinVar evidence. The fail-closed self-audit does not catch them either: `MISSING_PROVENANCE` iterates only over criteria that are already triggered (main: `self_audit.py:127-133`), so a variant with zero triggered criteria passes the audit and is never abstained.

The docstring added at `clinical_variant_reporter.py:325-326` states that a batch failure "aborts the run rather than returning placeholder evidence that can be misclassified", and the test name `test_malformed_vep_json_raises_instead_of_classifying_unannotated` repeats the claim. As written the guarantee holds only for transport failures and for a non-list payload. Please either extend the guard to fail when the response does not cover every submitted region, which would make the docstring true, or narrow the docstring to the cases actually covered and say plainly that a short or empty response still yields placeholders.

**2. CI does not run this test suite, so the green checks are not evidence for the 112 new lines.**

The test job runs a fixed allowlist of suites (`.github/workflows/ci.yml:116-147`) and `skills/clinical-variant-reporter/tests/` is not on it. The three passing `test (3.11/3.12/3.13)` checks on this PR therefore never executed any of the new tests. Relatedly, `test_partial_batch_failure_raises_instead_of_returning_partial_results` calls `self._mock_vep_success()`, which is not defined anywhere in the diff. It presumably exists on your branch base, since this branch is ahead of main (main still has `annotate_variants_vep` returning a bare list, `clinical_variant_reporter.py:271-274`), but I cannot verify it from the checkout and CI would not have caught it if it were missing. Please confirm the suite passes locally, and consider adding the file to the allowlist in this PR so the guarantee is enforced from here on.

**Non-blocking observations.**

- `main()` has no handler for the new `RuntimeError`, so the CLI now exits with a traceback. The exit status is non-zero and nothing is written, so fail-closed holds, and `test_cli_propagates_annotation_failure_without_writing_report` is right to assert that. It is inconsistent with the file's own convention of a printed `ERROR:` line plus `sys.exit(1)` (main: `clinical_variant_reporter.py:279-280` and `713-715`). A catch in `main()` would give the same safety with a readable message.
- There is no retry or backoff, so one transient 429 from Ensembl now discards every batch already annotated. Failing closed is still the correct default; a bounded retry would make large VCFs usable. Follow-up, not this PR.
- The version fixture `{"name": "ClinVar", "version": "09/2025"}` is not a shape Ensembl's `/info/variation/homo_sapiens` emits; it reports the ClinVar source version as a `YYYYMM` string such as `202509`. The value is passed through unchanged, so nothing is wrong with the code, but a fixture in the real format is better evidence of the contract.
- Now that any batch failure raises, `any_batch_succeeded` (`clinical_variant_reporter.py:374`) is true whenever `records` is non-empty. The updated docstring is consistent with that, but the flag no longer carries information.

---

manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Thanks, this is the right fix at the right layer. Replacing the placeholder path in `clinical_variant_reporter.py:360` with a raise removes a genuine silent-failure mode: before this, a transient VEP outage produced evidence objects with no ClinVar and no gnomAD, and those flowed into ACMG classification and into a report still labelled "Live (VEP REST API)". A network blip could therefore read as a clean VUS run. Failing closed is the safer default for a clinical report.

The `isinstance` guard on the response is a good addition, and the tests are checked against real VEP output shape rather than against the implementation: `input`, `most_severe_consequence`, `transcript_consequences`, `gene_symbol` and `consequence_terms` are fields VEP genuinely emits, and the `1 100 100 A/G 1` input string matches what `_vcf_to_vep_region` submits and what VEP echoes back. `test_partial_batch_failure_raises_instead_of_returning_partial_results` and the CLI test that asserts no `result.json` is written are the two that matter most, and both are here.

Merging. The points below are for later, not for this PR.

**Non-blocking suggestions**

1. The fail-closed guarantee is still partial. Immediately after the new guard, a variant missing from `result_map` in an otherwise successful batch still gets an unannotated placeholder. Worth flagging those variants explicitly in the report.
2. `main()` has no handler, so the new `RuntimeError` reaches the user as a traceback. A caught error with a clear message and a non-zero exit would read better.

---
Approved and merged.

Issue #365, opened 2026-08-26
`rnaseq-de`: four defects in the declared `nfcore-rnaseq-wrapper` → `rnaseq-de` handoff, two silent


javrodriguez:
The catalogue declares `nfcore-rnaseq-wrapper` and `rnaseq-de` as chaining partners, but their
formats do not meet. Running the handoff on real data surfaced four separate breaks. The second
and fourth are the serious ones: each produces a complete, plausible, **wrong** result with no
error and no warning.

**Environment:** `clawbio==0.6.1`, Python 3.12, Linux (Slurm HPC), nf-core/rnaseq 3.26.0,
`--backend pydeseq2`.

### 1. Non-numeric column rejected

nf-core/rnaseq emits `gene_id`, `gene_name`, then one column per sample. `rnaseq-de` documents
"first column is gene identifier" and coerces every later column to numeric, so `gene_name`
raises:

```
Count matrix contains non-numeric entries
```

A consumer of the declared partner's native output cannot use it unmodified.

### 2. Silent loss of gene identifiers — **the important one**

`rnaseq_de.py:288`:

```python
results_df.reset_index().rename(columns={"index": "gene"})
```

This assumes the index is *unnamed*. When the count matrix was read with a **named** index, the
rename is a no-op, the `gene` column never appears, and the identifier is dropped from the output
column selection.

**There is no error and no warning.** The first run produced a complete, well-formed
`de_results.csv` — correct row count, correct statistics — in which every gene was anonymous. It
surfaced only because a later line happened to crash on `row.gene`. Had that line not existed, the
result would have looked publishable.

Minimal reproduction of the no-op:

```python
import pandas as pd
df = pd.DataFrame({"x": [1]}, index=pd.Index(["ENSG1"], name="gene_id"))
out = df.reset_index().rename(columns={"index": "gene"})
print(list(out.columns))   # ['gene_id', 'x'] -- no 'gene' column, no error
```

Naming the identifier column `gene`, as the skill's own demo data does, avoids it. Suggested fix:
select the identifier column explicitly rather than relying on the index being unnamed, and fail
loudly if it is absent after the rename.

### 3. `FileExistsError` on rerun

The skill refuses a populated `--output` directory, so a rerun requires moving the previous
directory aside. That is defensible as a guard, but combined with (2) it means the natural
recovery loop — fix input, rerun — needs manual filesystem work.

### 4. Published `log2FoldChange` does not reflect the data — **also silent**

[...]

PR #371, opened 2026-08-27
fix(rnaseq-de): keep gene IDs on nf-core count handoff

krudo-taco:
Fixes the first two defects in #365.

## Skill affected

rnaseq-de

## Summary

- Accept nf-core `gene_id` + `gene_name` count matrices by dropping the annotation column before numeric coercion.
- After `reset_index()`, rename the identifier column to `gene` even when the index was named. Fail closed if that column is missing or empty.
- Leave output-directory overwrite behavior and LFC shrinkage for follow-up. Those are defects 3 and 4 in the issue.

## Test output

```text
load_counts nfcore ok
named index gene column ok
missing gene fail closed ok
pytest -k 'not pydeseq2'  # new tests passed; result.json checks need the real clawbio.common.report helper
```

### DIFF
--- a/skills/rnaseq-de/rnaseq_de.py
+++ b/skills/rnaseq-de/rnaseq_de.py
@@ -54,13 +54,28 @@ def parse_contrast(contrast: str) -> tuple[str, str, str]:
     return parts[0], parts[1], parts[2]
 
 
+_GENE_ID_COLUMNS = ("gene_id", "Geneid", "gene", "GENEID", "GeneID")
+_GENE_NAME_COLUMNS = ("gene_name", "gene_symbol", "GeneName", "symbol")
+
+
+def _select_gene_id_column(columns: pd.Index) -> str:
+    for name in _GENE_ID_COLUMNS:
+        if name in columns:
+            return name
+    if len(columns) == 0:
+        raise ValueError("Count matrix is missing a gene identifier column")
+    return str(columns[0])
+
+
 def load_counts(path: Path) -> pd.DataFrame:
     sep = _sep_for(path)
     df = pd.read_csv(path, sep=sep)
     if df.shape[1] < 3:
         raise ValueError("Count matrix must have one gene column and at least two samples")
-    gene_col = df.columns[0]
-    counts = df.set_index(gene_col)
+    gene_col = _select_gene_id_column(df.columns)
+    drop = [col for col in _GENE_NAME_COLUMNS if col in df.columns and col != gene_col]
+    counts = df.drop(columns=drop).set_index(gene_col)
+    counts.index.name = gene_col
     counts = counts.apply(pd.to_numeric, errors="coerce")
     if counts.isna().any().any():
         raise ValueError("Count matrix contains non-numeric entries")
@@ -285,13 +300,31 @@ def _try_de_pydeseq2(
             shrinkage["lfc_shrinkage_note"] = f"Attempted LFC shrinkage with '{coeff}' but failed: {exc}"
     else:
         shrinkage["lfc_shrinkage_note"] = "Could not identify a PyDESeq2 coefficient for LFC shrinkage."
-    res = stats.results_df.reset_index().rename(columns={"index": "gene"})
+    res = _require_gene_column(stats.results_df.reset_index())
     if "baseMean" not in res.columns:
         base_mean = (counts.div(counts.sum(axis=0), axis=1) * 1_000_000.0).mean(axis=1)
         res = res.merge(base_mean.rename("baseMean"), left_on="gene", right_index=True, how="left")
     cols = ["gene", "baseMean", "log2FoldChange", "pvalue", "padj"]
     keep = [col for col in cols if col in res.columns]
     return res[keep].sort_values("padj", ascending=True), shrinkage
+
+
+def _require_gene_column(results: pd.DataFrame) -> pd.DataFrame:
+    """Return DE results with a non-empty `gene` column, or fail closed."""
+    res = results.copy()
+    if "gene" not in res.columns:
+        for name in (*_GENE_ID_COLUMNS, "index"):
+            if name in res.columns:
+                res = res.rename(columns={name: "gene"})
+                break
+    if "gene" not in res.columns:
+        raise ValueError("Differential expression results are missing gene identifiers")
+    if res["gene"].isna().any():
+        raise ValueError("Differential expression results are missing gene identifiers")
+    genes = res["gene"].astype(str).str.strip()
+    if (genes == "").any() or (genes.str.lower() == "nan").any():
+        raise ValueError("Differential expression results are missing gene identifiers")
+    return res
 
 
 def _resolve_shrinkage_coeff(dds: object, factor: str, numerator: str) -> str:
--- a/skills/rnaseq-de/tests/test_rnaseq_de.py
+++ b/skills/rnaseq-de/tests/test_rnaseq_de.py
@@ -8,6 +8,7 @@ import pytest
 sys.path.insert(0, str(Path(__file__).parent.parent))
 
 from rnaseq_de import (
+    _require_gene_column,
     align_and_validate,
     compute_qc,
     de_simple,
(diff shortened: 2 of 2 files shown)


manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Thanks for this. I read the full diff and the surrounding skill at `skills/rnaseq-de/rnaseq_de.py`, and the change is correct for its domain.

**What is right**

- The header vocabulary is real, not invented to match the code. `salmon.merged.gene_counts.tsv` from nf-core/rnaseq genuinely emits `gene_id` and `gene_name` as its first two columns, and the fixture at `tests/test_rnaseq_de.py:155` reproduces that header exactly. `Geneid` at `rnaseq_de.py:57` is likewise the identifier featureCounts writes. This is the check that matters most on this repo, and it passes.
- Behaviour on existing inputs is unchanged. Both bundled fixtures use `gene` as the first column, and `gene` is in `_GENE_ID_COLUMNS` at `rnaseq_de.py:57`, so `_select_gene_id_column` at `rnaseq_de.py:61` selects the same column as the old positional `df.columns[0]`. The positional fallback at `rnaseq_de.py:67` preserves the old behaviour for unrecognised headers.
- No gene-to-statistic mis-assignment is introduced. `_require_gene_column` at `rnaseq_de.py:312` only renames a column; it never reorders or reindexes, so log2FoldChange and padj stay bound to the row they were computed on. The rename list covers both cases the PyDESeq2 path can produce: a named index carried through from `counts.index.name` at `rnaseq_de.py:78`, and an unnamed index that `reset_index()` turns into `index`.
- It fails closed rather than emitting a plausible-looking table. The three raises at `rnaseq_de.py:320` to `rnaseq_de.py:326` reject missing, NaN, empty and `"nan"` identifiers, so a results frame without gene IDs aborts instead of being written to `de_results.csv`. `tests/test_rnaseq_de.py:175` asserts that path, so the guard is not vacuous.
- The end-to-end test at `tests/test_rnaseq_de.py:181` asserts the actual Ensembl IDs survive to `de_results.csv`, not merely that a `gene` column exists.
- No network calls, no file uploads, no telemetry, no secrets, no clinical or pharmacogenomic logic touched. Nothing in the diff addresses the reviewer or attempts to direct the review. The static scan reported zero findings for this PR and I found nothing in the diff to contradict that, so there is no false positive to record.

**Non-blocking notes for a follow-up**

1. The drop at `rnaseq_de.py:76` is silent. A count matrix whose sample column happened to be named `symbol` or `gene_name` would have that sample removed from the analysis with no warning, where the old code raised "Count matrix contains non-numeric entries". The names involved are implausible for a sample, and the reported sample count at `rnaseq_de.py:559` is taken after the drop so the report would not overstate n, but a warning on any dropped column would keep the fail-loud property.
2. The arity guard at `rnaseq_de.py:73` runs before the drop. A file with `gene_id`, `gene_name` and a single sample passes it, then fails later with the unrelated message about contrast group sizes. Moving the check after the drop would give a clearer error.
3. Recognising `Geneid` at `rnaseq_de.py:57` implies featureCounts support that is not delivered: a raw featureCounts matrix also carries `Chr`, `Start`, `End`, `Strand` and `Length`, which are not in `_GENE_NAME_COLUMNS`, so it still fails on the non-numeric check. Worth either dropping those columns too or removing `Geneid` from the list.
4. `SKILL.md:63` still documents "first column is gene identifier", which is no longer what the loader does when a named identifier column appears later in the header. A one-line doc update would keep the contract accurate.

The test at `tests/test_rnaseq_de.py:202` pins `backend="simple"`, so the changed line at `rnaseq_de.py:303` is exercised end to end only by `test_pydeseq2_reports_lfc_shrinkage`, which is behind an `importorskip`. That is adequate given the unit tests on `_require_gene_column`, but it is worth knowing that the PyDESeq2 rename is untested wherever PyDESeq2 is absent.

---

Thanks for this. The column names you match are real, which is the first thing I checked: `gene_id` and `gene_name` are exactly what nf-core/rnaseq writes into `salmon.merged.gene_counts.tsv`, and `Geneid` is the featureCounts header. No invented vocabulary here. The underlying bug is real too. With an nf-core header the counts index is named `gene_id`, so `stats.results_df.reset_index()` yielded a `gene_id` column, the old `rename(columns={"index": "gene"})` never fired, and the `keep` filter then dropped the identifiers out of `de_results.csv` altogether. Failing closed on a missing identifier is the right instinct. Four things to resolve first.

**1. The loader and the checker disagree on what counts as a gene identifier.** `_select_gene_id_column` falls back to the first column whatever it is called (`rnaseq_de.py:67`), but `_require_gene_column` only recognises `_GENE_ID_COLUMNS` plus `index` (`rnaseq_de.py:316`). A matrix headed `Name` (Salmon), `target_id` (kallisto) or `ID` therefore loads without complaint and then dies in the PyDESeq2 backend at `rnaseq_de.py:321` with "Differential expression results are missing gene identifiers", when the identifiers are present under a name the allowlist does not know. The message is wrong and the hard failure is new. `_try_de_pydeseq2` already holds `counts`, so passing `counts.index.name` into `_require_gene_column` and renaming that column would cover every identifier name rather than five of them.

**2. The new end to end test does not exercise the code it is meant to cover.** `test_run_analysis_nfcore_counts_keep_gene_ids` runs with `backend="simple"` (`tests/test_rnaseq_de.py:203`), and `de_simple` builds the `gene` column straight from `counts.index` at `rnaseq_de.py:234`, so `_require_gene_column` is never called. That test passes identically with or without the change at `rnaseq_de.py:303`; it covers `load_counts` only. The one test that touches the renaming path builds the frame by hand with `name="gene_id"` (`tests/test_rnaseq_de.py:170`), which asserts the allowlist against itself rather than against PyDESeq2. Whether `DeseqStats.results_df` carries an index named `gene_id` after the AnnData round trip, or an unnamed one, is the whole question, and the diff does not answer it. `test_pydeseq2_reports_lfc_shrinkage` (`tests/test_rnaseq_de.py:135`) shows the pattern to follow: `pytest.importorskip("pydeseq2")`, a real run over a `gene_id` header, then assert the ENSG IDs survive into `de_results.csv`.

**3. The arity guard now counts the wrong columns.** `df.shape[1] < 3` at `rnaseq_de.py:73` runs before the gene name columns are dropped at `rnaseq_de.py:76`. A file headed `gene_id, gene_name, sample_1` passes a guard whose message promises "at least two samples" and produces a single sample matrix. It does fail later in `align_and_validate`, so nothing wrong is reported to the user, but the check belongs below the drop.

**4. Columns are dropped silently, and SKILL.md still documents the old contract.** `rnaseq_de.py:76` removes any column named `gene_name`, `gene_symbol`, `GeneName` or `symbol` with no message. If a sample is labelled `symbol`, its counts disappear from the library sizes and the QC table without a word. Same class of problem at `rnaseq_de.py:62`: the identifier is now searched across all columns rather than at position 0, so a sample named `gene` would be taken as the index. Please warn when a column is dropped or when the chosen identifier is not the first column, and update `SKILL.md:63`, which still states "first column is gene identifier".

One observation that does not block. Adding `Geneid` implies featureCounts support that is not actually there: that format also carries `Chr`, `Start`, `End`, `Strand` and `Length`, and `Chr` and `Strand` still trip the non-numeric check at `rnaseq_de.py:81`. Worth a separate PR if a featureCounts handoff is wanted.

I also checked the diff for network calls, credentials, telemetry and any movement of sample data off the machine, and found none. The static scan reported nothing on this PR, so there is no tool finding to dispute.

---

manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Thanks, this is a well targeted fix and the diagnosis is right.

The real defect was in `rnaseq_de.py:288`: `stats.results_df.reset_index()` yields a column named after the count matrix index (`gene_id`), not `index`, so the old rename was a no-op and `gene` then fell out of the `keep` list, writing a `de_results.csv` with no gene identifiers. Making `_require_gene_column` raise rather than pass through is the correct call; a DE table with anonymous rows is worse than an error. The `!= gene_col` guard in the drop list at `rnaseq_de.py:30` correctly leaves symbol-indexed matrices alone, and `gene_id`/`gene_name` are the real header names in `salmon.merged.gene_counts.tsv`, so the vocabulary checks out. featureCounts matrices still fail loudly on the `Chr`/`Strand` columns, which is the right behaviour.

This is merging. The points below are for later, not blockers.

**Non-blocking suggestions**

1. `rnaseq_de.py:30` drops name columns silently. A sample column named `symbol` or `gene_name` would vanish from the analysis with no warning, and `align_and_validate` only checks counts against metadata, not the reverse, so nothing would flag it. Dropping only columns that fail `pd.to_numeric` would be tighter.
2. `test_rnaseq_de.py:117` runs end to end with `backend="simple"`, which never reaches `_require_gene_column`. The pydeseq2 handoff, the path you actually fixed, is covered only by the unit test at `test_rnaseq_de.py:102`.
3. Minor: the empty-columns branch at `rnaseq_de.py:17` is unreachable given the `shape[1] < 3` check, and `counts.index.name = gene_col` at `rnaseq_de.py:32` is already set by `set_index`.

---
Approved and merged.

Issue #372, opened 2026-08-29
Reproducibility bundles are optional by construction: 34 skills have none, 18 created after the shared layer landed


camlloyd:
## Summary

34 of 101 skills produce no reproducibility artefacts at all, and only 12 use the shared `clawbio.common.reproducibility` layer. The remaining 46 hand-roll their own bundle writers.

This is not legacy debt. The shared layer landed 2026-04-12, and **18 of the 34 bundle-less skills were created after it**. The two most recent — `ancestry-risk-profiler` (2026-07-07) and `clinical-variant-prioritizer` (2026-08-09) — have neither a `reproducibility/` entry in their SKILL.md output tree nor a `TestOutputContract`.

## Why it keeps happening

Nothing in the repo requires a bundle. Every layer that looks like enforcement is descriptive or opt-out:

- **The conformance checklist has no reproducibility check.** `CLAUDE.md`'s 17 mandatory checks cover YAML fields, Trigger, Scope, Workflow, Example Output, Gotchas, Safety, Agent Boundary, demo data, tests and line count. Reproducibility is absent; the only occurrence of the word in the file is the routing row for `repro-enforcer`.
- **`docs/reproducibility.md` documents rather than mandates** — "Many skills write these files", "Do not assume every skill emits the exact same extras".
- **CI only validates a bundle that already exists.** `.github/workflows/ci.yml` solves `environment.yml` and fails if it will not build, but the preceding branch is `if [ ! -f "$env_file" ]` → run bare and `continue`. A skill with no bundle skips silently; a skill with a broken one fails. The incentive runs backwards.
- **`scaffold_skill.py` generates the prose but not the code.** It mentions `reproducibility/` twice, both inside the SKILL.md text it emits (output-tree diagram, "Audit trail" safety bullet), and generates zero bundle code. A freshly scaffolded skill documents a directory it will never produce.
- **`/pr-audit` is not in the repo.** `CLAUDE.md` says it "enforces this on every PR", but there is no definition in `commands/` or `.claude/commands/`. Whatever it checks is not version-controlled alongside the skills it audits.

### The loophole

`TestOutputContract` (`scaffold_skill.py:517`, added 2026-06-04) is real enforcement: it parses the SKILL.md `## Output Structure` tree and asserts every promised file exists after a `--demo` run. The template's default tree does list `reproducibility/commands.sh` and `environment.yml`.

[...]

PR #373, opened 2026-08-29
docs: require the reproducibility bundle in CLAUDE.md and AGENTS.md

camlloyd:
## Summary

- Adds an 18th check to the `CLAUDE.md` SKILL.md conformance checklist requiring a `reproducibility/` bundle written via `clawbio.common.reproducibility`, and updates the "(17 checks)" reference in step 9
- Rewrites the `AGENTS.md` Output bullet, which told agents skills write `reproducibility/` "as needed", into its own required bullet naming the three helpers and the `anchor=output_dir` convention, plus a bundle step in How to Add a New Skill
- Applies the same 18th check to the **second copy of the checklist** in `commands/new-skill.md`, which the `/new-skill` command reads, and adds the bundle to its Step 7 implementation requirements
- All three files now say not to prune `reproducibility/` from the SKILL.md `## Output Structure` tree to make `TestOutputContract` pass

Docs only. No skill code changes, no existing skill touched.

## Why

Part 1 of #372. 34 of 101 skills produce no reproducibility artefacts at all, and only 12 use the shared layer. This is not legacy debt: the shared layer landed 2026-04-12 and **18 of those 34 were created after it**, the most recent three weeks ago.

The root cause is that nothing required a bundle. The conformance checklist is the authoritative definition of a conforming skill and reproducibility appeared nowhere in its 17 checks — the only occurrence of the word in `CLAUDE.md` was the routing row for `repro-enforcer`. `AGENTS.md`, which is what non-Claude agents read, actively described the directory as optional.

The one real enforcement mechanism, `TestOutputContract`, parses the Output Structure tree and asserts every promised file exists after a `--demo` run. But its guidance — *"Keep the list honest: do not document outputs the skill does not always write"* — offers removal as the remedy, so for an author who never built a bundle, deleting the two `reproducibility/` lines is both the easiest fix and the documented-correct one. The test then passes vacuously. That instruction is right on its own terms; the problem is it was the only pressure in the system, so with no counterweight it pointed at deletion.

This PR supplies the counterweight. It does not touch the 34 skills or the 46 hand-rolled ones — those are tracked separately in #372.

### Note on checklist duplication

The conformance checklist is duplicated across `CLAUDE.md`, `commands/new-skill.md` and (as code) `scaffold_skill.py`. I initially updated only the first and caught the second on review; the third is updated in #374. `CONTRIBUTING.md` correctly defers to `CLAUDE.md` rather than keeping a fourth copy. That duplication is itself a drift risk and is tracked separately.

## Skill affected

None — repository-level agent instructions only.

## Checklist

- [x] Tests included and passing (`pytest -v`) — n/a, docs-only; suite run below to confirm no regression

## Test output

Docs-only change, so the verification that matters is that the checklist now parses to 18 rows and the AGENTS.md step list renumbers cleanly:

```
$ n=$(sed -n '/^### SKILL.md Conformance Checklist/,$p' CLAUDE.md | grep '^|' | grep -v '^|---\|^| Check |' | wc -l)
$ echo "checklist checks: $n"
checklist checks: 18

$ sed -n '/^## How to Add a New Skill/,/^## How to Modify/p' AGENTS.md | grep -oE '^[0-9]+\.' | tr -d '.' | tr '\n' ' '
1 2 3 4 5 6 7 8 9 10 11
```

Follow-up: the scaffolder change (`scaffold_skill.py` emitting working bundle code and `TestOutputContract` by default) is deliberately a separate PR, since this one is docs and that one is code with its own tests.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -53,7 +53,8 @@ Tests must pass on Python 3.10, 3.11, and 3.12. CI runs all three via GitHub Act
 - **Naming**: Skill folders use lowercase-hyphens (`gwas-lookup`). Python files use lowercase_underscores (`gwas_lookup.py`).
 - **Imports**: Sibling modules loaded via `importlib.util.spec_from_file_location` — no package structure.
 - **CLI**: Every skill script accepts `--input`, `--output`, and `--demo`. Use `argparse`.
-- **Output**: Skills write to `<output_dir>/report.md` (primary), plus `figures/`, `tables/`, `reproducibility/` subdirectories as needed. Return a `result.json` with structured findings.
+- **Output**: Skills write to `<output_dir>/report.md` (primary), plus `figures/` and `tables/` subdirectories as needed. Return a `result.json` with structured findings.
+- **Reproducibility**: Every skill with a Python implementation writes `<output_dir>/reproducibility/` (`commands.sh`, `environment.yml`, `checksums.sha256`) using `clawbio.common.reproducibility` — `write_commands_sh`, `write_environment_yml`, `write_checksums`. Do not hand-roll these writers. Label checksums relative to the output directory (`anchor=output_dir`) so `cd <output_dir> && sha256sum -c reproducibility/checksums.sha256` resolves.
 - **Dependencies**: Add to `requirements.txt` for core deps. Skill-specific heavy deps go in the skill's SKILL.md YAML `install` section.
 
 ## Project Structure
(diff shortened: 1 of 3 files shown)


manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Docs only, and the claims check out against the code.

I verified the three helpers named in `AGENTS.md:11` really exist with those signatures in `clawbio/common/reproducibility.py:130`, `:169` and `:223`, and that `write_checksums(..., anchor=output_dir)` labels each line relative to `output_dir` (`clawbio/common/reproducibility.py:155`). The documented `cd <output_dir> && sha256sum -c reproducibility/checksums.sha256` invocation therefore does resolve. The 17 to 18 renumbering is right as well: the existing table carries exactly 17 rows, the new row is added in both `CLAUDE.md:648` and `commands/new-skill.md:149`, and the AGENTS.md step list is renumbered consistently.

No skill logic, no variant or clinical handling, no change to data flow. This is merging.

**Non-blocking suggestions**

1. `AGENTS.md:11` states as present fact that "Every skill with a Python implementation writes `<output_dir>/reproducibility/`". That is not true of the tree yet; `skills/phylogenetics-builder/phylogenetics_builder.py:738`, `skills/bioconductor-bridge/bioconductor_bridge.py:390` and `skills/clinical-trial-finder/writers.py:682` each still hand-roll their own writer. Phrasing it as a requirement for new and updated skills would stop the line reading as a description of what is already there.
2. The new 18th check is manual only. Nothing under `scripts/` greps for `reproducibility`, so `skill-lint` will not enforce it. A lint rule in a follow-up would make the checklist self-checking.

The static scan reported nothing on this PR, and I found nothing to add to it.

---

---

PR #376, opened 2026-08-29
fix(nutrigx): correct rs4988235 lactose polarity and dominance

### LINKED ISSUE #237: fix(nutrigx): rs4988235 lactose polarity appears reversed
## Summary

NutriGx appears to have the lactase persistence / lactose intolerance polarity reversed for `rs4988235` in the SNP panel.

The current panel entry marks `A` as the `risk_allele` for `lactase_non_persistence`:

```json
{"rsid": "rs4988235", "gene": "MCM6", "ref_allele": "G", "risk_allele": "A", "nutrient_domain": "lactose", "weight": 0.95, "effect_direction": "lactase_non_persistence", "pmid": "12369657"}
```

On GRCh38 / forward strand, `rs4988235` is at `2:135851076` with alleles `G/A/C/T` in Ensembl, and the commonly discussed European lactase-persistence allele `T-13910` corresponds to `A` on the minus-oriented dbSNP/SNPedia representation. Therefore `AA` should indicate likely lactase persistence, not lactase non-persistence.

## Reproduction

Minimal 23andMe-style input:

```text
# rsid	chromosome	position	genotype
rs4988235	2	135851076	AA
```

Run:

```bash
python skills/nutrigx-advisor/nutrigx_advisor.py \
  --input /tmp/lct-mini.txt \
  --format 23andme \
  --output /tmp/nutrigx-lct-test \
  --no-figures
```

Observed `result.json` excerpt:

```json
{
  "lactose": {
    "score": 10.0,
    "category": "Elevated",
    "contributing_snps": [
      {
        "rsid": "rs4988235",
        "gene": "MCM6",
        "genotype": "AA",
        "risk_count": 2,
        "effect_direction": "lactase_non_persistence"
      }
    ]
  },
  "call": {
    "rsid": "rs4988235",
    "genotype": "AA",
    "normalised": "AA",
    "risk_allele": "A",
    "risk_count": 2
  }
}
```

This produces `Elevated` lactose intolerance / lactase non-persistence for `AA`.

## Expected behavior

For European lactase persistence interpretation:

- `rs4988235` persistence allele should be treated as protective / persistence, not as `lactase_non_persistence` risk.
- `AA` on GRCh38 forward-strand input should not score as high-risk lactose intolerance.
- If the skill wants to report non-persistence risk directly, the panel/scoring should encode the non-persistence allele for the input strand/orientation, or explicitly model persistence vs non-persistence instead of relying on a generic `risk_allele` field.

A related marker, `rs182549`, is already present in the ancestry panel and could be used as a consistency check. In common DTC/SNP-array files, `rs4988235 AA` together with `rs182549 TT` should point toward lactase persistence, not intolerance.

## Suggested fix

Audit the lactose domain panel and strand handling. At minimum, fix the `rs4988235` polarity in `skills/nutrigx-advisor/data/snp_panel.json` and add regression tests for:

- `rs4988235 AA` → likely lactase persistent / low non-persistence risk
- `rs4988235 AG` → intermediate or partial, depending on chosen model
- `rs4988235 GG` → lactase non-persistence risk

It would also be useful to include `rs182549` in NutriGx lactose scoring or as a validation warning when available.

krudo-taco:
Fixes #237

## Summary
- Risk allele on GRCh38 plus is G (non-persistence), not A.
- PMID 11788828 replaces the unrelated 12369657.
- Lactase persistence is dominant, so heterozygotes score Low.

## Testing
- pytest skills/nutrigx/tests/test_nutrigx.py -k rs4988235  (4 passed)

### DIFF
--- a/skills/catalog.json
+++ b/skills/catalog.json
@@ -3040,7 +3040,7 @@
       "name": "nutrigx",
       "cli_alias": "nutrigx",
       "description": "Personalised nutrition report from consumer genetic data (23andMe, AncestryDNA, VCF) — interrogates nutritionally-relevant SNPs and generates actionable dietary guidance, all computed locally.",
-      "version": "0.1.0",
+      "version": "0.2.0",
       "status": "mvp",
       "maturity_tier": "ci-validated",
       "maturity_evidence": {
--- a/skills/nutrigx/SKILL.md
+++ b/skills/nutrigx/SKILL.md
@@ -132,7 +132,7 @@ The Bio Orchestrator should route to this skill when the user says anything like
 
 | Gene    | SNP        | Sensitivity          | Effect                          |
 |---------|------------|----------------------|---------------------------------|
-| MCM6    | rs4988235  | Lactose intolerance  | Non-persistence of lactase      |
+| MCM6    | rs4988235  | Lactose intolerance  | Risk allele G (GRCh38 plus / -13910C) is non-persistence; persistence allele A (-13910T) is dominant |
 | HLA-DQ2 | Proxy SNPs | Coeliac / gluten     | HLA-DQA1/DQB1 risk haplotypes   |
 
 ### Antioxidant & Detoxification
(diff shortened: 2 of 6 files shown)


manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

**nutrigx: rs4988235 polarity**

Good fix, and the genetics is right. On the plus strand rs4988235 G is -13910C (non-persistence) and A is -13910T (persistence), so making G the risk allele corrects a genuinely inverted call: on main a GG homozygote scored 0.0 and was told "Likely lactase-persistent" at `generate_report.py:74`. PMID 11788828 checks out as Enattah et al., *Nat Genet* 2002;30:233-237, the right primary source. The dominant model is the standard one here, and the new tests earn their space because they exercise both the A/G and C/T encodings rather than only the form the code happens to emit. Raising on an unknown `inheritance` at `score_variants.py:27` rather than falling back to additive is the right call too.

**Non-blocking suggestions**

This is merging; these are for later.

- `snp_panel.json:35` now sets `ref_allele: "A"`, which is the non-risk allele, not the reference base (that is G). It matches how the field is actually used at `extract_genotypes.py:98-113`, so behaviour is correct, but the name is a trap. Consider `non_risk_allele`, or a comment, so nobody later matches it against a VCF REF column.
- With one SNP in the lactose domain the score is now only ever 0.0 or 10.0, so the "Moderate" text at `generate_report.py:75` is unreachable. Worth dropping or gating.

The static scan reported nothing here, and I found nothing to dispute in it.

---
Approved and merged.

PR #378, opened 2026-08-30
fix(rnaseq-de): verify LFC shrinkage matches the requested contrast

### LINKED ISSUE #365: `rnaseq-de`: four defects in the declared `nfcore-rnaseq-wrapper` → `rnaseq-de` handoff, two silent

krudo-taco:
Fixes defect 4 (the remaining silent one) of #365. Defects 1–2 were fixed in #371; defect 3 (rerun overwrite guard) stays as-is per that issue's own assessment that it is defensible.

## Skill affected

rnaseq-de

## Summary

**The mechanism, verified against pydeseq2 0.5.1–0.5.4 source:** `DeseqStats.lfc_shrink(coeff)` *replaces* `results_df["log2FoldChange"]` with that single coefficient's column. The published effect sizes are therefore only the requested contrast when the coefficient **is** that contrast — i.e. when the design's reference level is the contrast denominator and the column is exactly `factor[T.numerator]`. The skill relies on its re-leveling to guarantee this but never verified it, and the resolver's fallback loops matched by suffix/substring, which can resolve a **different level's** coefficient (e.g. levels `A`/`AB`) and silently publish a sign-flipped or near-null effect in place of the requested contrast.

Changes to `rnaseq_de.py`:

- `_resolve_shrinkage_coeff()` now returns a coefficient by **exact match only** (`factor[T.numerator]` present in `dds.varm["LFC"].columns`); the suffix and substring fallbacks are gone.
- New `_coefficient_matches_contrast()` guard: before shrinking, the candidate coefficient's MLE column must reproduce the requested contrast's MLE `log2FoldChange` (compared against the fitted model itself, so the check does not depend on pydeseq2's column-naming or reference-level conventions). If it doesn't — or no exact coefficient resolves — shrinkage is refused, the **unshrunk Wald MLE for the requested contrast** is published, and the reason is stated loudly: `warnings.warn` plus the existing `lfc_shrinkage_note` fields that already flow into `report.md` and `result.json` (`lfc_shrinkage_applied: false`).
- A crash inside pydeseq2's own machinery now surfaces as a clear `RuntimeError` naming the pydeseq2 version and the `--backend simple` escape hatch, instead of a bare internal traceback. Concretely: pydeseq2 0.5.0/0.5.1 with recent anndata raise a raw `IndexError` inside `fit_genewise_dispersions` on valid input; the same installed base behaves differently with older anndata, which is consistent with the version-dependent corruption reported in #365.
- `_require_gene_column()` accepts the input's identifier column name as a hint, so a count matrix whose first column is not one of the hardcoded gene-id spellings no longer fails only on the pydeseq2 path.

P-values, `baseMean`, output formats, and the simple backend are unchanged. On the standard path (re-leveling respected) shrinkage still applies exactly as before — `test_pydeseq2_reports_lfc_shrinkage` is untouched and still passes.

## Checklist

- [x] Demo data included for reviewers to test (not needed for this bug fix)
- [x] Follows [CONTRIBUTING.md](../blob/main/CONTRIBUTING.md) conventions

## Test output

New tests (17 total in the file, all passing on the CI-locked stack — pydeseq2 0.5.4, anndata 0.12.6, pandas 3.0.2 — and on pydeseq2 0.5.1/0.5.2/0.5.3):

- `test_pydeseq2_lfc_reflects_requested_contrast` — the acceptance test proposed in #365: feeds an nf-core-shaped `salmon.merged.gene_counts.tsv` (`gene_id` + `gene_name`) through `run_analysis(backend="pydeseq2")`, asserts gene identifiers survive, the published `log2FoldChange` correlates > 0.9 with log2 ratios of the normalized group means, and top-by-padj genes carry real effect sizes (not |LFC| < 0.05).
- `test_pydeseq2_refuses_shrinkage_on_mismatched_coefficient` — three-level factor, resolver monkeypatched to hand back the wrong level's coefficient: shrinkage refused with the loud note, and the published LFC still matches the requested contrast's direction.
- `test_resolve_shrinkage_coeff_exact_match_only` — regression for the substring trap: levels `A`/`AB` with `A` as reference must resolve to `""`, not `condition[T.AB]`.
- `test_pydeseq2_internal_failure_is_reported_clearly` — monkeypatched internal crash surfaces as `RuntimeError` naming the backend and `--backend simple`.

```text
$ pytest skills/rnaseq-de/tests/test_rnaseq_de.py -q
17 passed, 4 warnings in 5.74s
```

One honest caveat, in the spirit of the #360/#356 threads: I could not reproduce the reporter's exact corr ≈ 0.33 numbers on any pydeseq2 version reachable today (0.5.1–0.5.4 all publish correct LFCs on my synthetic and realistic-data runs when the re-leveling is respected; 0.5.0/0.5.1 with recent anndata crash loudly instead). What I could verify from source is that the published LFC after `lfc_shrink` is *unconditionally* the resolved coefficient's column, that nothing enforced the coefficient ≡ contrast invariant, and that the substring fallback is capable of resolving a wrong-level coefficient. This PR makes the invariant explicit and verified rather than accidental, which is the fix #365 asks for.

### DIFF
--- a/skills/rnaseq-de/rnaseq_de.py
+++ b/skills/rnaseq-de/rnaseq_de.py
@@ -6,6 +6,7 @@ import argparse
 import hashlib
 import math
 import sys
+import warnings
 from datetime import datetime, timezone
 from pathlib import Path
 
@@ -241,6 +242,9 @@ def de_simple(
     return result
 
 
+_LN2 = math.log(2)
+
+
 def _try_de_pydeseq2(
     counts: pd.DataFrame,
     metadata: pd.DataFrame,
@@ -250,6 +254,7 @@ def _try_de_pydeseq2(
     denominator: str,
 ) -> tuple[pd.DataFrame, dict[str, str | bool]] | None:
     try:
+        import pydeseq2
         from pydeseq2.dds import DeseqDataSet
         from pydeseq2.ds import DeseqStats
     except Exception:
@@ -282,25 +287,60 @@ def _try_de_pydeseq2(
             refit_cooks=True,
         )
 
-    dds.deseq2()
-    stats = DeseqStats(dds, contrast=[factor, numerator, denominator])
-    stats.summary()
+    try:
+        dds.deseq2()
+        stats = DeseqStats(dds, contrast=[factor, numerator, denominator])
+        stats.summary()
+    except Exception as exc:
+        # Some pydeseq2 releases crash inside their own machinery on valid
+        # input (e.g. 0.5.0/0.5.1 with recent anndata raise a bare IndexError
+        # in fit_genewise_dispersions). Surface why the backend died instead
+        # of letting an opaque internal traceback escape (issue #365).
+        raise RuntimeError(
+            f"The pydeseq2 backend failed (pydeseq2 "
+            f"{getattr(pydeseq2, '__version__', 'unknown')}): "
+            f"{type(exc).__name__}: {exc}. "
+            "If this persists, rerun with --backend simple for Welch-t "
+            "statistics instead of DESeq2."
+        ) from exc
+
     shrinkage = {
         "lfc_shrinkage_applied": False,
         "lfc_shrinkage_coeff": "",
         "lfc_shrinkage_note": "",
     }
+    # DeseqStats.lfc_shrink(coeff) REPLACES results_df["log2FoldChange"]
+    # with that single coefficient's column. That is only the requested
+    # contrast when the design's reference level is the contrast denominator
+    # and the coefficient is exactly factor[T.numerator]. Shrink only when
+    # both hold; otherwise publish the unshrunk Wald MLE for the requested
+    # contrast and say so loudly (issue #365, defect 4).
     coeff = _resolve_shrinkage_coeff(dds, factor, numerator)
-    if coeff:
+    if not coeff:
+        shrinkage["lfc_shrinkage_note"] = (
+            "No pydeseq2 coefficient exactly matches the requested contrast "
+            f"(expected '{factor}[T.{numerator}]' with '{denominator}' as the "
+            "reference level); publishing the unshrunk Wald MLE for the "
+            "requested contrast."
+        )
+        warnings.warn(shrinkage["lfc_shrinkage_note"], stacklevel=2)
+    elif not _coefficient_matches_contrast(stats, dds, coeff):
+        shrinkage["lfc_shrinkage_note"] = (
+            f"Resolved coefficient '{coeff}' does not reproduce the requested "
+            "contrast's MLE log2FoldChange; refusing LFC shrinkage and "
+            "publishing the unshrunk Wald MLE for the requested contrast."
+        )
+        warnings.warn(shrinkage["lfc_shrinkage_note"], stacklevel=2)
+    else:
         try:
             stats.lfc_shrink(coeff=coeff)
             shrinkage["lfc_shrinkage_applied"] = True
             shrinkage["lfc_shrinkage_coeff"] = coeff
         except Exception as exc:
             shrinkage["lfc_shrinkage_note"] = f"Attempted LFC shrinkage with '{coeff}' but failed: {exc}"
-    else:
-        shrinkage["lfc_shrinkage_note"] = "Could not identify a PyDESeq2 coefficient for LFC shrinkage."
-    res = _require_gene_column(stats.results_df.reset_index())
+    res = _require_gene_column(
+        stats.results_df.reset_index(), identifier_name=counts.index.name
+    )
     if "baseMean" not in res.columns:
         base_mean = (counts.div(counts.sum(axis=0), axis=1) * 1_000_000.0).mean(axis=1)
         res = res.merge(base_mean.rename("baseMean"), left_on="gene", right_index=True, how="left")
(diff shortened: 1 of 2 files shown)


krudo-taco:
The only failing check across the 3.11/3.12/3.13 matrix is `tests/test_generate_catalog.py::test_checked_in_catalog_is_current`, which also fails on `origin/main` at the base commit of this PR (verified locally: `1 failed, 29 passed` on a clean main checkout) and is addressed by #377. The RNA-seq DE test step passes on all three versions: `17 passed, 4 warnings`. No files related to the catalog are touched here.

manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Good fix, and the right shape of one. Verifying the coefficient against the fitted model rather than against pydeseq2's column naming is correct: `results_df["log2FoldChange"]` is `LFC @ contrast_vector / ln 2`, so dividing `varm["LFC"][coeff]` by `_LN2` and comparing the vectors (rnaseq_de.py:398) really does test whether the contrast vector is the unit vector on that coefficient. Dropping the suffix and substring fallbacks in `_resolve_shrinkage_coeff` (rnaseq_de.py:371) closes the 'A' versus 'AB' trap, and when nothing matches you publish the unshrunk Wald MLE for the requested contrast and record it in the report, which is the safe direction. Wrapping the `dds.deseq2()` crash in a named RuntimeError is fail-loud, not fail-quiet. This is merging; the points below are for later.

**Non-blocking suggestions**

1. `test_resolve_shrinkage_coeff_exact_match_only` is gated on `pytest.importorskip("pydeseq2")` (test_rnaseq_de.py:336) but only touches a fake dds and pandas. Drop the skip so the resolver regression still runs where pydeseq2 is absent; at present all four new tests vanish together.
2. `np.allclose(..., equal_nan=True)` (rnaseq_de.py:403) returns True when both vectors are entirely NaN, so a fully degenerate fit would pass the guard. A finite-entry count would close that.
3. The naive comparator at test_rnaseq_de.py:282 takes log2 of log1p CPM, a double log. The correlation assertion still holds, but one scale would read more clearly.

---
Approved and merged.

PR #379, opened 2026-08-30
feat(gwas-prs): add reproducibility bundle

krudo-taco:
## Summary

- Add a reproducibility bundle to `gwas-prs` using the shared `clawbio.common.reproducibility` layer.
- Record input and scoring-file SHA-256 digests without storing genotype paths, contents, or raw trait queries.
- Generate portable replay commands, a rebuildable environment, provenance, and output checksums.
- Align the documented output contract and skill/catalog version with the current implementation.
- Expand targeted CI to execute the complete `skills/gwas-prs/tests/` directory.

Related to #372.

## Skill affected

`gwas-prs`

## Checklist

- [x] Demo data included for reviewers

## Test output

- `uv run --with pytest python -m pytest skills/gwas-prs/tests -q`
  - 115 passed
- `uv run python clawbio.py run prs --demo --output <temp>`
  - exit 0
  - generated all 8 documented files
- `agentskills validate skills/gwas-prs`
  - valid
- Clean Python 3.11 replay environment:
  - installed only requests, opentelemetry-sdk, numpy, and pandas
  - generated `commands.sh` completed successfully
- Ruff checks and formatting for new files passed
- `python -m compileall` passed
- `git diff --check` passed

The repo-level catalog suite currently has one unrelated baseline failure for
the NutriGx 0.2.0 catalog sync. That fix is isolated in #377 and is not
duplicated here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,7 +135,7 @@ jobs:
         run: uv run pytest skills/clinpgx/tests/test_clinpgx.py -v
 
       - name: Run GWAS PRS tests
-        run: uv run pytest skills/gwas-prs/tests/test_gwas_prs.py -v
+        run: uv run pytest skills/gwas-prs/tests/ -v
 
       - name: Run GWAS Lookup tests
         run: uv run pytest skills/gwas-lookup/tests/test_gwas_lookup.py -v
--- a/skills/catalog.json
+++ b/skills/catalog.json
@@ -2265,7 +2265,7 @@
       "name": "gwas-prs",
       "cli_alias": "prs",
       "description": "Calculate polygenic risk scores from DTC genetic data using the PGS Catalog",
-      "version": "0.1.0",
+      "version": "0.2.0",
       "status": "mvp",
       "maturity_tier": "ci-validated",
       "maturity_evidence": {
--- a/skills/gwas-prs/SKILL.md
+++ b/skills/gwas-prs/SKILL.md
@@ -3,7 +3,7 @@ name: gwas-prs
 description: Calculate polygenic risk scores from DTC genetic data using the PGS Catalog
 license: MIT
 metadata:
-  version: 0.1.0
+  version: 0.2.0
   openclaw:
     requires:
       bins:
@@ -111,14 +111,18 @@ When the user asks for a polygenic risk score calculation:
 
 ```
 output_directory/
-├── report.md              # Full narrative report with risk categories
-├── tables/
-│   └── scores.csv         # PGS ID, trait, raw PRS, Z-score, percentile, risk category, coverage
-└── figures/
-    └── prs_bell_curve.png # Bell curve with individual score marked (optional)
+├── prs_report.md          # Full narrative report with risk categories
+├── prs_results.json       # Compact per-score result records
+├── prs_variants.csv       # Per-variant dosage and contribution details
+├── result.json            # Standard ClawBio result envelope
+└── reproducibility/
+    ├── commands.sh        # Portable replay command
+    ├── environment.yml    # Rebuildable Python environment
+    ├── provenance.json    # Input and scoring-file hashes plus safe parameters
+    └── checksums.sha256   # Integrity digests for outputs and bundle metadata
 ```
 
-### report.md Format
+### prs_report.md Format
 
 The report includes:
 - Patient summary (file name, total SNPs, date)
@@ -127,7 +131,7 @@ The report includes:
 - Methodology notes and references
 - Safety disclaimer
 
-### `prs_results.json` identity fields
+### `prs_results.json` Fields
 
 | Column | Description |
 |---|---|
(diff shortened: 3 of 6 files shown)


manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Thanks, this is a careful change. No PRS maths is touched: dosage, raw score, z-score and percentile logic are all untouched, so no result can change. The SKILL.md rewrite is a genuine fix, not a rename: `gwas_prs.py:1348`, `:1372` and `:1389` already wrote `prs_report.md`, `prs_results.json` and `prs_variants.csv`, so the old `report.md` and `tables/scores.csv` docs were stale. Provenance keeps the genotype path out of the bundle and stores only digests, and `test_provenance_contains_hashes_without_private_paths_or_raw_input` asserts that rather than assuming it. The demo CLI test asserting the exact output file set is the right kind of contract test. Nothing leaves the machine.

The scanner raised no critical or warning findings. Its INFO note on `subprocess.run` in the test is expected; the test runs the demo CLI deliberately.

Merging. The points below are for later.

**Non-blocking suggestions**

1. `repro_bundle.py:132` checks `args.trait` before `args.pgs_id`, but `main()` resolves the other way round (`gwas_prs.py:1054` for `--pgs-id`, then `:1144` for `--trait`). If someone passes both flags, the replay command and the `selection` field describe a run that did not happen. Matching the precedence would fix it.
2. `repro_bundle.py:168` hashes the trait query unsalted. Trait names are a tiny search space, so the digest is a fingerprint, not anonymisation; worth wording the SKILL.md line that way.
3. `_read_skill_version` raises at import time if the SKILL.md indentation shifts, and the test pins "0.2.0", so each bump needs a test edit.

---

Thanks, this is a careful bundle. It reuses `clawbio.common.reproducibility` rather than reimplementing it; `provenance.json` keeps the genotype path and the free-text trait out of the record while still hashing the input and every scoring file actually used; and `test_demo_cli_writes_the_documented_output_contract` runs the real CLI and pins the output set, so the stale `tables/scores.csv` docs cannot drift back. The SKILL.md rename matches what the script already wrote.

One blocking point.

`repro_bundle.py:132` picks the replay mode with `if args.trait: ... elif args.pgs_id:`, but `main()` in `gwas_prs.py` resolves scores with the `elif args.pgs_id:` branch ahead of `elif args.trait:`. The flags are not mutually exclusive; the CLI only requires one of the three. So a run with both `--pgs-id` and `--trait` scores the given PGS ID, while `commands.sh` replays a trait search and `_safe_parameters` (`repro_bundle.py:168`) records `mode: trait` with a query hash. The bundle then describes a different analysis from the run that produced it, which is the one thing it exists to prevent. Mirroring the order used in `main()` in both functions should fix it. Could you confirm and adjust?

**Non-blocking suggestions**

- `repro_bundle.py:73` declares `numpy>=1.24` and `pandas>=2.0`, but I could not find either on this skill's import path; only `requests` and the opentelemetry stack look eager. Trimming would make the declared environment match reality.

The static scan reported no findings on this PR, and I found nothing it wrongly accused.

---

krudo-taco:
Rebased onto current `main` (`1c224f1`) and addressed the audit. Still one commit, `dd20b09`.

**Blocking point, selector precedence.** Confirmed and fixed. `repro_bundle.py` now resolves the selector through one helper, `_selection`, in the order `main()` uses: `--demo`, then `--panel-id`, then `--pgs-id`, then `--trait`. Both the replay command and `provenance.json`'s `selection` come from it, so they cannot disagree with each other or with the run. `main()` on current `main` rejects more than one selector, so the both-flags case is no longer reachable from the CLI, but the bundle no longer relies on that. Two tests cover it: a parametrised one that hands the bundle every combination of selectors and checks which flag lands in `commands.sh` and which `mode` lands in provenance, and one that reads the `if/elif` chain out of `gwas_prs.py`'s source and asserts `_selection` tests the attributes in the same order, so a future reorder in `main()` goes red here.

**`--panel-id`.** #380 landed on `main` after this PR was opened and added `--panel-id` as the normal way to score a bundled panel. The bundle did not know the flag: after the rebase a `--panel-id` run replayed with no selector and provenance recorded `mode: pgs_id, pgs_id: "None"`. It is now replayed as `--panel-id <id>` and recorded as `mode: panel_id`. Same rebase: `prs_results.json` gained `score_id`, `legacy_pgs_id` and `legacy_pgs_compatibility`, so `provenance.json` scoring entries now carry those too, with a genuinely null `pgs_id` for panels rather than the string `"None"`. The demo CLI test now cross-checks the provenance `score_id` set against `prs_results.json` and asserts every result field has a row in the SKILL.md table.

**numpy and pandas in `environment.yml`.** Kept, and I would push back on trimming them. This skill never calls either, but `clawbio/common/__init__.py` imports `scrna_io`, which imports both at module level, so the moment `gwas_prs.py` does `from clawbio.common.checksums import sha256_hex` they load. Checked empirically rather than by reading imports:

```
uv run --no-project --python 3.11 --with 'requests>=2.31' --with 'opentelemetry-sdk>=1.20,<2' \
  python skills/gwas-prs/gwas_prs.py --demo --output /tmp/prs
# ModuleNotFoundError: No module named 'numpy'

# same command plus --with 'numpy>=1.24' --with 'pandas>=2.0'
# Reproducibility bundle written to /tmp/prs/reproducibility
```

The dependency list is now a named constant with that reasoning in a comment, and `test_numpy_and_pandas_really_are_eager_imports` asserts both are in `sys.modules` after importing the bundle module, so if `clawbio.common` ever stops loading them the test says so and the two lines can go.

**Trait digest.** Reworded in SKILL.md and in the code comment: an unsalted SHA-256 over a small search space is a fingerprint that lets a replay confirm it used the same query, not anonymisation.

**`_read_skill_version`.** Now matches `version:` at any indentation inside the frontmatter block, and the sync test compares `TOOL_VERSION` against SKILL.md and the catalog entry instead of pinning `"0.2.0"`, so a bump needs no test edit. A test covers a re-indented frontmatter.

Verification on `dd20b09`:

- `uv run --with pytest python -m pytest skills/gwas-prs/tests -q`: 156 passed (all three files, including #380's `test_curated_panel_ids.py`)
- `tests/test_generate_catalog.py` and `tests/test_approve_skills_only.py`: 64 passed; `scripts/generate_catalog.py` produces no diff
- `clawbio-bench --smoke --harness gwas_prs` at the pinned revision: 5/8 (62.5%), same as `bench_baseline.json`
- `python clawbio.py run prs --demo`: exit 0, all eight documented files present
- `agentskills validate skills/gwas-prs`: valid
- ruff on the three changed Python files reports nothing `main` does not already report

manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Read the full diff against `clawbio/common/reproducibility.py`. This is careful work, and it changes no scoring maths: no PRS, percentile or risk-category logic is touched, and nothing leaves the machine.

The bundle keeps the genotype path, filename and contents out of `provenance.json` and `commands.sh`, asserted at `tests/test_repro_bundle.py:129` and `:221`, and the trait digest is documented honestly as a fingerprint rather than anonymisation, which is the right call for a search space that small. The SKILL.md field table fix is a real doc correction, not cosmetics: the old table promised `coverage_pct`, whereas the value is a fraction, now correctly named `overlap_fraction`. `tests/test_repro_bundle.py:391` pins the documented contract against a real `--demo` run, so those names cannot drift back.

**Non-blocking suggestions** (this is merging; these are for later)

- `repro_bundle.py:183` asks the replayer for `INPUT_FILE`, but nothing checks it against `provenance.json`'s `input.sha256`. A replay against a different genotype file yields different scores and still looks successful. A digest check in the preflight would close that.
- `repro_bundle.py:156` reads `args.panel_id`, so it depends on the `--panel-id` selector being present in `main()`. `tests/test_repro_bundle.py:323` pins the order against the source, so a rebase that loses it fails loudly rather than recording the wrong selector; worth a note on merge order.
- `_safe_parameters` (`repro_bundle.py:218`) records the raw `--pgs-id` string, while `main()` upper-cases it. `scoring_files[].score_id` already carries the resolved identity, so this is cosmetic.

---
Approved and merged.

PR #380, opened 2026-08-31
fix(gwas-prs): give curated panels honest identifiers

krudo-taco:
## Summary

- make the six bundled PRS panels use canonical `CLAWBIO-*` identifiers and filenames instead of PGS Catalog accessions
- add `--panel-id` to the skill and top-level CLI, while preserving one explicitly marked PGS000013 compatibility path for the pinned clawbio_bench revision
- carry panel provenance through JSON reports, profile-report, ancestry guidance, documentation, and workshop examples

Refs #356.

## Skill affected

gwas-prs, with downstream identity rendering updates in profile-report and ancestry-risk-profiler.

## Checklist

- [x] SKILL.md is present and updated

## Demo output

`python clawbio.py run prs --input skills/gwas-prs/demo_patient_prs.txt --panel-id CLAWBIO-T2D-8 --output /tmp/prs-panel`

Produces:

- `score_id: CLAWBIO-T2D-8`
- `pgs_id: null`
- `curated_demo_panel: true`
- `legacy_pgs_id: PGS000013`
- `legacy_pgs_compatibility: false`

The report states that the panel is illustrative, uses approximate weights, and is not a PGS Catalog score or published PRS result.

## Test output

- related unit and integration tests: 217 passed
- top-level CLI flag tests: 9 passed
- `generate_curated_scores.py --check`: passed
- Python compile check: passed
- `git diff --check`: passed
- pinned clawbio_bench gwas-prs harness: 5/8, 62.5%, unchanged from the recorded baseline

The remaining three benchmark failures are the existing low-coverage `missing_output` cases; this change does not introduce a benchmark regression.


manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

This is a real fix, and it closes the right hole. Renaming the bundled files to `CLAWBIO-*_GRCh37.txt` means the panels no longer sit in the PGS Catalog download-cache path, so `--trait` and five of the six accessions now reach the genuine API instead of being answered by a lookalike. `gwas_prs.py:1407` turning the `--trait` cache hit from a warning into a refusal is the right direction. Keying `CURATED_SCORES` on the panel id also removes the older hazard where a freshly downloaded real PGS000001 would have inherited the curated panel's reference distribution in `estimate_percentile`. The PMIDs and loci counts are unchanged and still pinned, and `test_curated_panel_ids.py:113` asserts the absence of Catalog cache paths rather than restating the implementation, which is the kind of test that actually holds.

Merging.

### Non-blocking suggestions

1. `gwas_prs.py:191` keeps one alias, and a `--pgs-id PGS000013` run still writes `"pgs_id": "PGS000013"` into `prs_results.json` next to `curated_demo_panel: true`. A consumer keying on `pgs_id` alone would still read a Catalog accession. `profile_report.py:322` is safe because it prefers `score_id`, but please drop the alias once the pinned bench can ask for `CLAWBIO-T2D-8`.

2. `skills/wgs-prs/wgs_prs.py:357` only forwards `--pgs-id`, so wgs-prs users can no longer reach the curated panels except through that alias. Worth forwarding `--panel-id` in a follow-up.

3. `test_score_provenance.py` still calls it `test_there_are_exactly_three_known_lookups` while asserting 2. Rename when convenient.

---
Approved and merged.

PR #381, opened 2026-09-01
fix(pathway-enricher): submit gene lists as multipart form data

krudo-taco:
## Summary

- Submit Enrichr `/addList` gene lists as multipart form fields, matching the upstream API contract.
- Add regression coverage for the endpoint, multipart payload, description, and timeout.
- Restore the live `pathway-enricher --demo` workflow, which previously failed with HTTP 400.

## Skill affected

`pathway-enricher`

## Checklist

## Test output

- Python 3.11: `16 passed`
- Python 3.12: `16 passed`
- Python 3.13: `16 passed`
- Live six-database demo: exit 0
- Generated report, result JSON, figures, CSV tables, checksums, replay commands, and environment metadata
- `git diff --check`: passed
- `compileall`: passed

### DIFF
--- a/skills/pathway-enricher/pathway_enricher.py
+++ b/skills/pathway-enricher/pathway_enricher.py
@@ -115,12 +115,12 @@ RATE_LIMIT_DELAY = 0.5      # seconds between library queries
 def _post_gene_list(genes: list[str], description: str = "ClawBio gene set") -> str:
     """Upload a gene list to Enrichr and return the userListId."""
     payload = {
-        "list": "\n".join(genes),
-        "description": description,
+        "list": (None, "\n".join(genes)),
+        "description": (None, description),
     }
     resp = requests.post(
         f"{ENRICHR_BASE}/addList",
-        data=payload,
+        files=payload,
         timeout=30,
     )
     resp.raise_for_status()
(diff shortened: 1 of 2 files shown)


manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Thanks, this is the right fix. Enrichr's `/addList` endpoint genuinely requires `multipart/form-data`; a urlencoded body is rejected, so `pathway_enricher.py:117` posting `data=payload` would have failed the `raise_for_status()` on every real run. The `(None, value)` tuple form in `pathway_enricher.py:111` is exactly the shape Enrichr's own API example uses, and it keeps the field names `list` and `description` that the endpoint expects. Nothing here touches scoring, p-value handling or the result row indices, so enrichment output is unaffected. Failures still propagate rather than being swallowed. Merging.

**Non-blocking suggestions**

1. `tests/test_pathway_enricher.py:40` asserts the exact call shape against a `MagicMock`, so it pins the request format but cannot catch Enrichr changing its contract. A note in the test docstring pointing at the Enrichr API docs would make it clear to a future reader where that shape comes from, rather than it looking like it was copied from the implementation.

2. Unrelated to your change, but adjacent to it: `SKILL.md:30` claims enrichment runs "all locally, with no data leaving the machine", which `SKILL.md:143` then correctly qualifies as gene symbols being sent to the public Enrichr API. Since your PR is the code path that does that upload, you may want to tidy line 30 while you are here. Entirely optional and not a condition of merge.

---
Approved and merged.

PR #382, opened 2026-09-02
fix(clinical-variant-reporter): parse ClinVar significance into terms instead of crashing on VEP's list

### LINKED ISSUE #328: [Bug] clinical-variant-reporter: live ACMG classification crashes on list-valued clin_sig from VEP

krudo-taco:
Fixes #328.

## Summary

- Parse ClinVar significance into structured terms before PS1, PP5 and BP6 read it, so the engine accepts VEP REST's list-valued `clin_sig` without crashing and never substring-matches a joined string.
- Define explicit conflict semantics: any spelling of ClinVar's "submitters disagree" token (`conflicting_interpretations_of_pathogenicity`, `conflicting_classifications_of_pathogenicity`, `conflicting_data_from_submitters`, and anything else on the `conflicting` stem), or pathogenic-family and benign-family terms co-occurring, withholds all three ClinVar-backed rules and says so in the criterion detail line.
- Keep every scalar input that worked before working: `Pathogenic/Likely pathogenic`, `Benign/Likely benign`, `Pathogenic|risk_factor`, `Pathogenic, low penetrance` and the review-star gate are unchanged. Two scalar behaviours change on purpose, both pinned by their own test: a mixed string such as `Benign; Pathogenic` is withheld instead of read as pathogenic support, and `Benign|conflicting_data_from_submitters` no longer fires BP6. The old BP6 had no conflicting check at all; the other two conflicting tokens only blocked it because "pathogenicity" contains "pathogenic".
- Render the ClinVar column in `acmg_classifications.tsv` and the ClinVar line in `report.md` through a display helper, so a list never lands in a clinical artefact as a Python repr.

This takes the direction agreed on #328 and in the #344 reviews (structured terms, abstain on conflict) and addresses the points that PR was blocked on: real vocabulary instead of an invented token, splitting on `/`, `|` and `;` as well as `,`, a stem match so the 2024 ClinVar rename and the pre-rename spelling are both caught, the TSV writer, and tests whose vocabulary comes from the data source.

## Skill affected

`clinical-variant-reporter`

## What changed

`acmg_engine.py`

- `normalize_clinvar_terms(value)` reduces a string, a list/tuple or `None` to an ordered, de-duplicated tuple of snake_case terms. It splits on `[,;/|]`, lower-cases, collapses whitespace to `_` and strips stray underscores (VEP writes `pathogenic,_low_penetrance`). Any other shape raises `TypeError` rather than silently reading as "no ClinVar evidence".
- `ClinVarAssertion` (frozen dataclass) exposes `pathogenic_terms`, `benign_terms`, `conflicting_terms` and the derived `is_conflicting`, `supports_pathogenic`, `supports_benign`, `conflict_reason`.
- `VariantEvidence.clinvar_significance` is widened to `str | Sequence[str] | None`; rules go through `ev.clinvar_assertion`, writers through `ev.clinvar_significance_text`.
- PS1, PP5 and BP6 share `_clinvar_source` and `_withheld_detail`, so a withheld rule's detail reads e.g. `PS1 withheld — ClinVar flags this record as conflicting (conflicting_classifications_of_pathogenicity)`.

`clinical_variant_reporter.py`

- `_extract_evidence_from_vep` keeps the list VEP returns (comment explains why); `report.md` and the TSV use the display text.

`SKILL.md`

- New "ClinVar Assertion Handling" subsection under Algorithm / Methodology documenting the table below. Frontmatter untouched, so `skills/catalog.json` is unchanged.

| ClinVar value | PS1 / PP5 | BP6 |
|---|---|---|
| `Pathogenic`, `Likely pathogenic`, `Pathogenic/Likely pathogenic`, `Pathogenic\|risk_factor`, `["uncertain_significance", "pathogenic"]` | eligible | no |
| `Benign`, `Likely benign`, `Benign/Likely benign` | no | eligible |
| any `conflicting_*` token, alone or alongside other terms | withheld | withheld |
| pathogenic-family and benign-family terms together, e.g. `["benign", "pathogenic"]` | withheld | withheld |
| `Uncertain significance`, `drug_response`, `risk_factor`, `not_provided`, unknown terms | no | no |

The co-occurrence clause matters for VEP specifically: at release 116 the `clin_sig` list is aggregated over every ClinVar record at the site, across alleles, and does not carry the conflicting token itself. rs1042522 comes back as `["uncertain_significance", "benign", "likely_benign", "pathogenic"]` because the C allele is pathogenic and the T allele benign. Withholding there is the conservative direction for an ACMG engine.

## Reproduction, before and after

`issue328.vcf` with two lines, `13 32339267 rs886040553 A T` and `17 43093464 . A G`, run live against rest.ensembl.org (release 116):

`main` (`1fe0064`):

```
  File ".../acmg_engine.py", line 234, in _eval_ps1
    clinvar_lower = ev.clinvar_significance.lower()
AttributeError: 'list' object has no attribute 'lower'
```

No `report.md`, `result.json` or tables are written.

This branch: exit 0, all artefacts written. The TSV row for rs886040553 reads `uncertain_significance; pathogenic` in the `clinvar_significance` column, and `report.md` shows `ClinVar=uncertain_significance; pathogenic, stars=0` on the PS1/PP5/BP6 rows.

## Test output

`uv run --with pytest python -m pytest skills/clinical-variant-reporter/tests -q`

```
177 passed in 2.93s
```

77 on `main`, 100 new. Red first: the new tests were written before the engine change and fail against `main` (the parser API does not exist there, and the list-valued evidence the end-to-end tests build raises the issue's `AttributeError: 'list' object has no attribute 'lower'` from `main`'s `classify_variant`).

New tests, all parametrised on vocabulary ClinVar or VEP actually emit:

- `TestClinVarTermNormalisation`: single, duplicate, compound (`/`, `|`, `;`, `,`), case/whitespace, `pathogenic,_low_penetrance`, empty/`None`/unsupported shape.
- `TestClinVarAssertionSemantics`: single terms, duplicates, every conflicting spelling, co-occurrence, neutral/unknown terms, both low-penetrance spellings; a parity test against the pre-fix substring rule on 20 scalar inputs, and a separate test that pins the two deliberate divergences (mixed strings; benign plus `conflicting_data_from_submitters`) with the legacy result each one used to give. Two tests prove each guard clause is load-bearing on its own: `["pathogenic", "conflicting_classifications_of_pathogenicity"]` has no benign term so only the token clause can block; `["benign", "pathogenic"]` has no token so only the co-occurrence clause can.
- `TestClinVarCriteriaWithStructuredTerms`: PS1/PP5/BP6 on list and compound inputs, star gate unchanged, detail names the conflict, `source` byte-identical for scalars, and a classify-level case where the conflict token turns Likely Pathogenic into VUS.
- `TestListValuedClinSigEndToEnd`: `_extract_evidence_from_vep` -> `classify_variant` -> `generate_report` on the release-116 payload shape (list `clin_sig`, string `clin_sig_allele`), plus the dict-stars variants, asserting the report and TSV contain `uncertain_significance; pathogenic` and no `['`.

Mutation check, deleting or weakening one guard at a time and re-running the suite:

| Mutant | Result |
|---|---|
| drop explicit-token clause | 16 failed |
| drop `conflicting` stem match (exact tokens only) | 1 failed |
| drop co-occurrence clause | 11 failed |
| split on `,` only (the #344 shape) | 21 failed |
| drop `/` / `|` / `;` from the separator set | 14 / 6 / 3 failed |
| conflict still supports pathogenic / benign | 22 / 13 failed |
| drop `likely_pathogenic` / `pathogenic_low_penetrance` / `likely_pathogenic_low_penetrance` from the family | 4 / 1 / 1 failed |
| no dedup / no lowercase / no underscore strip | 3 / 42 / 1 failed |
| remove PP5 star gate | 2 failed |
| display helper emits `repr` | 3 failed |
| withheld detail silent about the conflict | 5 failed |
| `TypeError` swallowed for a dict | 1 failed |

One mutant survives: replacing the exact family-set test on each already-split term with `"pathogenic" in term`. On the ClinVar and Ensembl vocabularies those are the same predicate (no real term contains `pathogenic` without belonging to the pathogenic family or the conflicting family), so I did not invent a token to kill it.

Other checks:

- Demo output is unchanged: `--demo` on `main` and on this branch produce byte-identical `result.json`, `acmg_classifications.tsv`, `secondary_findings.tsv` and `database_versions.json`; `report.md` and `commands.sh` differ only in the input/output paths.
- `clawbio-bench --smoke` at the pinned revision, harnesses `clinical_variant_reporter` / `cvr_identity` / `cvr_correctness`: 80.0 / 100.0 / 69.2 on both `main` and this branch, matching `bench_baseline.json`. Verdicts differ only in temp paths and wall time.
- `uvx --from 'skills-ref>=0.1.1,<1' agentskills validate skills/clinical-variant-reporter`: valid. SKILL.md is 253 lines.
- `ruff check` on the three changed files reports the same pre-existing findings as `main` and nothing new. `python -m compileall` and `git diff --check` clean.
- `uv run python clawbio.py run acmg --demo` exits 0.

## Out of scope, noted for later

In all 24 variants I sampled from rest.ensembl.org release 116, `clin_sig_allele` came back as a string (`"G:conflicting_classifications_of_pathogenicity;T:pathogenic;G:uncertain_significance"`), so the `isinstance(..., dict)` guard from #322 leaves `clinvar_review_stars = 0` and PS1/PP5/BP6 never fire in live mode today. That string is also the only place the per-allele conflict flag lives; the aggregate list drops it. Reading review status and the alt-allele assertion out of `clin_sig_allele` would make live mode use ClinVar at all, and would let the engine distinguish "this allele is conflicting" from "another allele at this site is benign". That is a separate change with its own tests and I have not started it.

## Checklist

- [x] Demo data included for reviewers to test (existing `--demo` panel; unchanged)

### DIFF
--- a/skills/clinical-variant-reporter/SKILL.md
+++ b/skills/clinical-variant-reporter/SKILL.md
@@ -174,6 +174,8 @@ The classification engine implements the ACMG/AMP 2015 framework (Richards et al
 
 ClinVar significance is parsed into terms before PS1, PP5 or BP6 read it; the rules never substring-match a joined string. VEP REST returns `clin_sig` as a list aggregated over every ClinVar record at the site, and ClinVar's own strings join terms with `/`, `|`, `;` or `,`. Both shapes bucket the same way.
 
+For live VEP REST extraction, the site-level `clin_sig` aggregate is not used as evidence because it can mix assertions from different alternate alleles. The extractor keeps only the `clin_sig_allele` entry that exactly matches the queried ALT. Missing, malformed, non-string or unmatched allele-specific payloads are treated as absent evidence. VEP REST does not pair that allele-specific assertion with independently verifiable review stars, so live extraction records zero stars and withholds PS1, PP5 and BP6. Cached or directly constructed evidence that pairs a ClinVar assertion with a trustworthy review-star value still follows the table below.
+
 | ClinVar value | PS1 / PP5 | BP6 |
 |---------------|-----------|-----|
 | `Pathogenic`, `Likely pathogenic`, `Pathogenic/Likely pathogenic`, `Pathogenic\|risk_factor`, `Pathogenic, low penetrance`, `pathogenic_low_penetrance`, `likely_pathogenic_low_penetrance` | eligible | no |
--- a/skills/clinical-variant-reporter/SKILL.md
+++ b/skills/clinical-variant-reporter/SKILL.md
@@ -170,6 +170,20 @@ The classification engine implements the ACMG/AMP 2015 framework (Richards et al
 - **BP4**: CADD < 15
 - **ClinVar minimum stars for PS1/PP5/BP6**: ≥ 2
 
+### ClinVar Assertion Handling
+
+ClinVar significance is parsed into terms before PS1, PP5 or BP6 read it; the rules never substring-match a joined string. VEP REST returns `clin_sig` as a list aggregated over every ClinVar record at the site, and ClinVar's own strings join terms with `/`, `|`, `;` or `,`. Both shapes bucket the same way.
+
+| ClinVar value | PS1 / PP5 | BP6 |
+|---------------|-----------|-----|
+| `Pathogenic`, `Likely pathogenic`, `Pathogenic/Likely pathogenic`, `Pathogenic\|risk_factor`, `Pathogenic, low penetrance`, `pathogenic_low_penetrance`, `likely_pathogenic_low_penetrance` | eligible | no |
+| `Benign`, `Likely benign`, `Benign/Likely benign` | no | eligible |
+| `Conflicting_interpretations_of_pathogenicity`, `Conflicting_classifications_of_pathogenicity`, `conflicting_data_from_submitters` (alone or alongside any other term) | withheld | withheld |
+| pathogenic-family and benign-family terms together, e.g. `["benign", "pathogenic"]` | withheld | withheld |
+| `Uncertain significance`, `drug_response`, `risk_factor`, `not_provided`, unrecognised terms | no | no |
+
+"Withheld" means the rule does not fire and records the conflict as its reason in the criterion's detail field. A variant whose ClinVar records disagree therefore loses the ClinVar-backed criteria rather than being promoted on one side of the disagreement; the remaining criteria still combine as usual. Review-star gating (≥ 2) applies on top of this in every case.
+
 ## Example Queries
 
 - "Classify the variants in this exome VCF according to ACMG guidelines"
--- a/skills/clinical-variant-reporter/acmg_engine.py
+++ b/skills/clinical-variant-reporter/acmg_engine.py
@@ -13,6 +13,8 @@ References:
 
 from __future__ import annotations
 
+import re
+from collections.abc import Sequence
 from dataclasses import dataclass, field
 from typing import Literal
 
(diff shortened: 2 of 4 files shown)


manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

**Domain review: clinical-variant-reporter**

This is a good fix. Parsing significance into terms instead of substring matching is the right shape, the conflicting vocabulary covers the 2023 rename as well as `conflicting_data_from_submitters`, and keeping the old rule as a parity oracle with the two divergences pinned separately makes the behaviour change auditable. The 2 star gate is untouched, and pathogenic plus benign co-occurring now withholds rather than promotes.

One question before this can merge.

**`clin_sig` is site level, not allele level.** `clinical_variant_reporter.py:231` stores the first colocated variant's whole list, and `acmg_engine.py:186` only withholds when both families co-occur. Your own fixture shows the gap: `clin_sig_allele` is `G:conflicting_classifications_of_pathogenicity;T:pathogenic;G:uncertain_significance` (tests/test_clinical_variant_reporter.py:1334). Query the G allele at that site and the aggregate is `["uncertain_significance", "pathogenic"]`, so PS1 and PP5 fire on T's assertion; tests/test_clinical_variant_reporter.py:1351 pins that result as Likely Pathogenic. Could you filter terms by the queried alt using `clin_sig_allele`, or withhold when a site carries assertions for more than one allele?

**Non-blocking suggestions**
- `CLINVAR_TERM_SEPARATORS` (acmg_engine.py:105) omits `&`, the delimiter VEP uses in VCF and tabular `CLIN_SIG`. It fails closed today, so no wrong call comes of it.
- PP5 and BP6 were withdrawn by ClinGen SVI (Biesecker and Harrison, Genet Med 2018, PMID 29543229). Pre-existing, but worth a line in SKILL.md.

---

krudo-taco:
Addressed the reviewer concern by keeping live VEP ClinVar evidence allele-specific:\n\n- _extract_evidence_from_vep() now reads only the clin_sig_allele entry for the queried ALT and never falls back to site-level clin_sig for live REST extraction.\n- Unmatched, malformed, non-string, or star-only allele payloads fail closed with zero ClinVar stars.\n- Added regression coverage for multi-allelic sites, unsafe payload shapes, and the site-level aggregate leaking across alleles.\n\nValidation: python -m pytest skills/clinical-variant-reporter/tests/test_clinical_variant_reporter.py -k 'ClinVar or allele_specific or ListValuedClinSigEndToEnd' -q (111 passed).

manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Thanks, this is a careful fix and the code comments made it easy to verify.

What is right, specifically:

- Parsed terms replace the substring test, so `"Benign; Pathogenic"` no longer reads as pathogenic support and `pathogenicity` no longer matches `pathogenic` (`acmg_engine.py:105`).
- The vocabulary is what the sources really emit: both conflicting spellings, VEP's `conflicting_data_from_submitters`, and `/ | ; , &` as separators.
- Dropping site level `clin_sig` (`clinical_variant_reporter.py:255`) stops a multi allelic locus lending another ALT's assertion to the queried one, and the removed `review_status_stars` read was a field VEP REST does not return.

Merging; the points below are for later.

**Non-blocking suggestions**

1. Live runs are now star blind, so PS1, PP5 and BP6 can never fire on VEP REST data (`clinical_variant_reporter.py:255`). That is the safe direction and SKILL.md says so, but reading review status from the ClinVar API would restore the evidence.
2. `_clinvar_significance_for_alt` (`clinical_variant_reporter.py:193`) matches the padded VCF ALT against VEP's allele; for indels VEP uses trimmed alleles or `-`, so the ClinVar column will read N/A. Display only while stars are zero.
3. PP5 and BP6 were withdrawn by ClinGen SVI (Biesecker and Harrison, Genet Med 2018;20:1687-8, PMID 29543229), and PS1 and PP5 both fire from one assertion (`acmg_engine.py:476`, `:520`). Pre-existing; this PR narrows both, so a good moment to plan removal.

---
Approved and merged.

PR #383, opened 2026-09-03
gi-expression: read the submitted length from meta.sequence_length

boldakov:
The gi-expression report's "of N bp submitted" clause and the submitted_sequence_length summary field read data.input.submitted_sequence_length, an untyped echo field the API is removing. They now read meta.sequence_length, which the API publishes on every response, sync and polled, and which is typed and required. The clause renders under an isinstance guard, so losing the echo would have dropped it from the report without an error. Fixtures carry the meta field instead, and there is one new test that a response without the echo still renders the clause.


manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Thanks for this. Moving off an untyped echo onto a required field is the right instinct, and I like that you pinned both windowing paths rather than just the one you needed. The reasoning in the test docstring about the `isinstance` guard making the drop silent is exactly the right thing to worry about.

Two things I need confirmed before this goes in.

1. **Does `meta.sequence_length` mean submitted, or scored?** `data.input` already carries a `sequence_length` next to the old `submitted_sequence_length`, and in the only live capture every one of these is 9198 with `scored_window` `[0, 9198]`, so the fixture cannot tell the two apart. If `meta.sequence_length` is the length actually scored rather than the length posted, then `gi_runner.py:309-310` will print the window size as the submitted length precisely when a locus is windowed, which is the one case the clause exists to reveal. A live capture where the two differ would settle it.

2. **The swap can be silent in the other direction.** If the API has not yet started emitting `meta.sequence_length`, `.get` returns `None`, the guard fires and the clause vanishes, which is the failure you are trying to prevent. Reading meta with a fallback to the echo would cover both deploy orders.

### Non-blocking suggestions

`LIVE_EXPRESSION_BODY` is documented as a verbatim capture, but `test_gi_response_shape.py:289` and `:317` now describe a response that has not shipped. Worth re-capturing once the change lands.

The static scan reported nothing here and I found nothing for it to have missed.

---

boldakov:
Both points taken, and both are now on the branch.

1. meta.sequence_length is the submitted length, not the scored one. Live capture from the production API today, a 25,000 bp sequence with tss_index 12500: meta.sequence_length 25000, meta.task_specific_counts.scored_window [7901, 17099], and the old echo data.input.sequence_length 9198 (the window). The fixture now carries that capture, so the two lengths differ in the test body and a reader that confuses them fails here.

2. The runner reads meta.sequence_length and falls back to data.input.submitted_sequence_length when meta has no length, so the clause survives either deploy order. Two tests pin it: meta wins when both are present, and the echo is used when meta is missing.

For the record, the API side of the removal is sequenced after this merges and reaches a release, not before.

manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Thanks, this is a careful migration and the new fixture is better than the one it replaces. The arithmetic checks out: tss_index 12500 with a window of [7901, 17099) is exactly the documented 9,198 bp server-side cut, so the capture now separates the submitted length from the scored window, which the old 9,198 bp capture could not. Spotting that the `isinstance` guard in `gi_runner.py:390` would have made the field's removal silent is the right diagnosis, and the fallback keeps either deploy order working.

One thing to confirm before this merges.

1. Nothing in this repo names `meta.sequence_length`, and `skills/gi-expression/SKILL.md:177` still tells readers that `data.input.scored_window` "also carries `submitted_sequence_length`". The only evidence that the new key is the submitted length rather than the 9,198 bp scored window is the fixture added here (`clawbio/tests/test_gi_response_shape.py:294`). Could you cite the published schema or contract revision that defines it, and update SKILL.md:177 in the same PR? If the field turns out to be the window length, `gi_runner.py:310` preferring meta over the echo would turn a correct clause into a wrong one, and that clause is the user's only check on windowing.

**Non-blocking suggestions**

- `_write_report` is already passed the locally counted submitted length (`gi_runner.py:311`). Comparing the server's number against it and warning on a mismatch would make the field's meaning self-checking.
- The fixture now mixes a fresh capture with prediction values from the old one. You flag that clearly, so it is fine; a second named fixture may age better.

---

boldakov:
On the citation: meta.sequence_length is the submitted length, not the scored window. It is typed in the published schema as Meta.sequence_length, "Length of the submitted sequence (bp)", and it is required on every response, sync or job result.

The revision that settles it is 13 (2026-09-03), which removed data.input.sequence_length from the echo on all seven predict operations precisely because that field meant two different things: the 9,198 bp scored window on expression and the submitted length everywhere else. meta.sequence_length was already the typed, unambiguous one, so it survived and the echo field went.

Measured against api.genomicintelligence.ai just now, 12,000 bp submitted with tss_index 6000:

  data.input.scored_window       [1401, 10599]   (9,198 bp)
  data.input.submitted_sequence_length  12000
  meta.sequence_length           12000

So gi_runner.py:310 preferring meta over the echo is correct, and the clause it feeds is safe.

On SKILL.md:177: you are right that the wording was wrong, and I have pushed a fix to this branch. submitted_sequence_length is a sibling of scored_window inside data.input, not something the window carries. The line now says the submitted length is meta.sequence_length, echoed alongside the window as data.input.submitted_sequence_length.

On the first non-blocking suggestion, comparing the server number against the locally counted length and warning on a mismatch: worth doing, but it is a behaviour change rather than a doc fix, so I would rather not fold it into this PR. Happy to open it separately.

The second one is fair. A second named fixture captured end to end would age better than the current mix; I left the mix because it keeps the prediction values comparable with the fixture it replaces.

boldakov:
One more commit on the branch, dc1e1aa, correcting something I got wrong earlier in this thread.

Three comments I wrote said data.input.submitted_sequence_length is being removed from the API. It is not. Revision 13 removed data.input.sequence_length, which is a different key, and submitted_sequence_length stayed. Production still returns it. Checked against api.genomicintelligence.ai today, 25,000 bp with tss_index 12500, running 2026.09.04.4:

    data.input = {"sequence_name": "probe-25kb", "description": "K562 cells",
                  "tss_index": 12500, "scored_window": [7901, 17099],
                  "submitted_sequence_length": 25000}
    meta.sequence_length = 25000

No behaviour changes and no assertions move. Reading meta.sequence_length first is still the right call, but the reason that survives is which side of the pin the field sits on: data.input is an untyped object that no published schema and no consumer pin covers, while meta.sequence_length is typed and required. A comment predicting a removal that never happened would send the next reader looking for a contract change that does not exist.

The capture comment now also states why that body leaves the echo key out, which is so the tests below it can only pass by reading meta instead of falling through to the fallback.

212 passing across the four gi test files, same as before the commit.

boldakov:
One more, d8c3a15, finishing what the comment fix started. The body labelled a live capture was not one any more: it left out data.input.submitted_sequence_length, which production does return, and its prediction numbers came from an older capture. It is now one whole response verbatim, taken today against 2026.09.04.4, 25,000 bp with tss_index 12500 and the same description the old fixture used.

Recapturing it turned up a third length the old body did not carry, data.summary.sequence_length, which is the 9,198 bp window and not the submission. All three are pinned now, so a reader who reaches for the sequence_length in summary fails here.

This also fixes a test that did not test its own name. test_the_submitted_length_clause_comes_from_meta_not_the_echo forced the meta read by leaving the echo out, but with the key absent a runner preferring the echo falls back to meta and is right by accident. I checked by flipping the preference in _summarize locally: the old test passed and only test_meta_wins_over_the_echo caught it. The echo now holds 9198, the number someone confusing the two would produce, so the wrong value has to reach the report for the test to pass. Same flip, the rewritten test fails. The report leg is unchanged.

That covers your second non-blocking suggestion as well, since the fixture is no longer a mix. No runner changes, 212 passing across the four gi test files.

manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

## Domain review: SAFE

Read the full diff plus `_summarize` and `_write_report` in the checkout. The fixture work is the strong part: recapturing one whole PROD response so the submitted length (25,000) and the scored window (9,198) are *different* numbers is what makes the two readers distinguishable at all, and `test_the_submitted_length_clause_comes_from_meta_not_the_echo` pins it by putting a wrong value in the echo rather than deleting the key. An absent key would have passed under either reader, so that choice matters. The new fixture is internally consistent: 12500 ± 4599 gives [7901, 17099), 9,198 bp, and ln(1.035) = 0.0344 matches the stated `log(TPM+1)` unit.

Nothing here changes the prediction, and the fallback at `gi_runner.py:321` preserves the old behaviour when `meta.sequence_length` is absent, so neither deploy order drops the clause. The SKILL.md line now matches the code. The static scan reported no findings and I found none it missed. This is merging.

### Non-blocking suggestions

- `_write_report` already receives the locally counted length as `sequence_length` (`gi_runner.py:311`) and prints it in the header. Comparing it against `summary["submitted_sequence_length"]` would turn the one remaining failure mode, meta ever carrying a length that is not the submission, into a visible warning instead of a quiet wrong clause.
- The clause renders under `isinstance(submitted, int)` (`gi_runner.py:390`), so a future string valued length would still vanish in silence. Worth the same treatment later.

Thanks for the careful write up in the fixture header; it made this quick to check.

---
Approved and merged.

PR #385, opened 2026-09-04
Document test verification results and clarify optional dependencies

chronicgiardia:
## Summary

- Adds `docs/testing.md`: a full per-path test verification run (4,290 passed, 55 failed, 19 errors, 98 skipped across 79 `pytest.ini` test paths), documents why a bare `uv run pytest` fails at collection time (a `conftest.py` module-name collision under `--import-mode=importlib` when two skills' `tests/` packages both resolve to the same module name), and why CI runs pytest per-path instead of once for the whole repo.
- Triages all 13 failing suites: 9 are caused purely by missing *optional* per-skill dependencies (not installed by base `uv sync`, by design), 1 (`proteomics-clock`) has both a missing dependency and an unrelated real bug, and 3 (`clawbio/common`, `bgpt-mcp`, `bigquery-public`) are genuine code/test issues unrelated to any dependency.
- Updates `README.md`: fixes the "Run tests" section (previously recommended a bare `uv run pytest`, which fails), and expands the optional-dependencies table from 4 to 11 entries based on this run, each with the exact install command sourced from that skill's own `SKILL.md`.
- Updates `CONTRIBUTING.md`: fixes AI-agent guidance that told agents to run a bare `pytest -v` across the whole repo.
- Flags one gap found along the way: `robotary`'s `fastapi` dependency isn't declared in any `SKILL.md`, requirements file, or `pyproject.toml`.

## Why

Running `uv sync` followed by the documented `uv run pytest` currently fails immediately for any new contributor or agent, with a confusing plugin-registration error unrelated to their actual change. This documents the real cause and the per-path workaround CI already uses, and gives an accurate, sourced list of which test failures are expected (missing optional deps) vs. real.

## Testing

This PR is documentation-only; no source code changed. All claims in `docs/testing.md` are backed by an actual `uv sync` + per-path `pytest` run performed today, with root causes traced to specific `ModuleNotFoundError`/`ImportError` tracebacks and cross-checked against each skill's declared `SKILL.md` `install:` block.

Co-Authored-By: Warp <agent@warp.dev>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -174,7 +174,7 @@ AI coding agents (Codex, Devin, Claude Code, Cursor, etc.) should follow the sam
 1. Read [`AGENTS.md`](AGENTS.md) for setup, commands, code style, and project structure
 2. Read the target skill's `SKILL.md` before modifying any code
 3. Use `python clawbio.py list` to verify skills still load after changes
-4. Run `python -m pytest -v` to confirm all tests pass
+4. Run `python -m pytest skills/your-skill-name/tests/ -v` to confirm your skill's tests pass. Do not run a bare `pytest -v` across the whole repo — it fails at collection time on a `conftest.py` module-name collision between skills; see [docs/testing.md](docs/testing.md)
 5. Regenerate `skills/catalog.json` if you changed any SKILL.md YAML frontmatter: `python scripts/generate_catalog.py`
 
 ### SKILL.md Quality Checklist
--- a/README.md
+++ b/README.md
@@ -479,8 +479,15 @@ python clawbio.py run methylation --geo-id GSE139307 --output results_methylatio
 ### Run tests
 
 ```bash
-uv run pytest                # or: pip install pytest && python -m pytest
+uv run pytest <path>          # one test path at a time, e.g.:
+uv run pytest skills/pharmgx-reporter/tests/ -v
 ```
+
+Running a single bare `uv run pytest` (no path) fails at collection time on
+a `conftest.py` module-name collision between skills. CI works around this
+by invoking pytest separately per test path; do the same locally. See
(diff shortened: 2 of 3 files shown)


manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  
Static scan: 0 critical, 15 warning. Domain review: complete.

Thanks, this is a genuinely useful piece of maintenance. Documentation only: `CONTRIBUTING.md`, `README.md` and a new `docs/testing.md`. No skill code, no clinical logic, no data-source vocabulary, so there is no path here to a wrong scientific result.

I checked the substantive claims against the checkout rather than taking them on trust:

- The collision cause is right. `skills/eqtl-catalogue-region-fetch/tests/` and `skills/gwas-catalog-region-fetch/tests/` each carry `__init__.py` and `conftest.py`, with no package marker at the skill level, exactly as `docs/testing.md:80` describes.
- The per-path workaround matches reality: `.github/workflows/ci.yml:123` onwards invokes pytest once per path and never bare.
- `robotary/server.py:17` does import `fastapi` directly, and it is absent from `pyproject.toml`.

This is merging. The points below are for later.

**Non-blocking suggestions**

1. `pytest.ini` currently lists 80 `testpaths` entries, not 79. `docs/testing.md:111` and `docs/testing.md:157` both say 79, though your own arithmetic (67 clean plus 13 failing) sums to 80.
2. `docs/testing.md:126` says fastapi is "not declared in any SKILL.md". It is declared in `skills/turingdb-graph/SKILL.md:52`, just not for `robotary/`. Narrowing the wording would keep the claim exact.
3. For the record, the static scanner raised 14 "unpinned pip install" warnings on this PR. They are false positives: your tables quote install lines each skill already declares in its own `SKILL.md`, and the diff adds no dependency. Nothing to change.

CI has not run yet; it is queued for maintainer approval.

---

Issue #367, opened 2026-08-26
Tutorial notebook teaches a frontmatter schema that fails skill-lint and cannot be routed


manuelcorpas:
`docs/tutorial-build-a-skill.ipynb` still teaches the pre-`metadata:` frontmatter schema, so a contributor who follows it end to end lands on a PR that fails `skill-lint` and cannot be routed to by the orchestrator.

Two concrete defects:

1. **Inline flow style fails the validator.** The tutorial's frontmatter uses `format: [vcf, csv, tsv, txt]`. PyYAML accepts it; the pinned `strictyaml` validator rejects it with `Found ugly disallowed JSONesque flow mapping`. This is now the third occurrence of this class in the repo:
   - 2026-08-11, `os: [darwin, linux]` in a workflow, which left `main` red for three days
   - #345, `format: [vcf, csv, tsv, txt]`, straight from this tutorial
   - #364, `packages: []`, from a contributor who was not following the tutorial

2. **No `metadata.openclaw.trigger_keywords`.** The tutorial uses the old flat layout with top-level `version`, `author`, `inputs`, `outputs` and `tags`. Current skills nest these under `metadata:` and declare `metadata.openclaw.trigger_keywords`. Without those keywords the orchestrator cannot route to the skill at all, so a tutorial-shaped skill is unreachable even once it lints clean.

`templates/SKILL-TEMPLATE.md` was already migrated to block style and the nested layout. The fix landed in the template but not in the notebook, so the notebook is now the stale copy.

#345 was closed as a direct consequence of this: a contributor completed the tutorial in good faith and opened a PR on its output, which then failed two CI checks for reasons that were entirely our doing.

Suggested fix: regenerate the notebook's frontmatter block from the current template, and add a final tutorial cell that runs `skill-lint` locally so the failure surfaces before a PR is opened rather than after.

PR #389, opened 2026-09-04
feat(skill-builder): write reproducibility bundles via clawbio.common

krudo-taco:
Addresses #372 (the remaining skill-builder scaffolder item after #374).

## Summary

- `skill-builder` no longer hand-rolls `commands.sh` / `environment.yml` with `write_text`. Its own report bundle and the generated skill skeleton both go through `clawbio.common.reproducibility`, including `checksums.sha256` with `anchor=output_dir`.
- Generated `SKILL.md` frontmatter now matches `templates/SKILL-TEMPLATE.md`: top-level `name` / `description` / `license`, everything else under `metadata:`, block lists instead of `os: [darwin, linux]` and `tags: [demo]`. Those flow-style lists fail `agentskills validate` even though PyYAML accepts them (same class as #367).
- Tests pin the shared-layer import, the checksum labels, a real `--demo` run of a generated skill, and the absence of flow-style lists in frontmatter.

## Skill affected

`skill-builder`

## Checklist

## Test output

```text
$ python -m pytest skills/skill-builder/tests/test_skill_builder.py -q
122 passed in 13.50s

$ python -m pytest tests/test_scaffold_skill.py -q
6 passed in 0.96s

$ python skills/skill-builder/skill_builder.py --demo --output /tmp/skill_builder_repro_demo
# Demo complete. Checksums verify:
report.md: OK
result.json: OK

$ uvx --from 'skills-ref>=0.1.1,<1' agentskills validate /tmp/skill_builder_repro_demo/hello-bioinformatics
Valid skill: /tmp/skill_builder_repro_demo/hello-bioinformatics
```

This does not migrate the remaining bundle-less compute skills listed in #372. Those stay one-skill-per-PR as the issue asked.

### DIFF
--- a/skills/skill-builder/SKILL.md
+++ b/skills/skill-builder/SKILL.md
@@ -146,7 +146,7 @@ Full spec with all optional fields:
 3. **Fill defaults**: domain → "bioinformatics", version → "0.1.0", capabilities/triggers → generic placeholders
 4. **Render SKILL.md**: Fill YAML frontmatter + all 13 required body sections from template
 5. **Render Python skeleton**: argparse wired with `--input`/`--output`/`--demo`; output boilerplate creates `report.md`, `result.json`, reproducibility bundle
-6. **Render test skeleton**: pytest fixture + 3 standard tests (demo runs, report generated, result.json valid)
+6. **Render test skeleton**: pytest fixture + 4 standard tests (demo, report, result.json, reproducibility bundle)
 7. **Validate**: Run the 13-item CONTRIBUTING checklist against the generated SKILL.md before writing
 8. **Register**: Append catalog entry; patch `clawbio.py` SKILLS dict via targeted string replacement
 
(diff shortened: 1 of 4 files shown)


manuelcorpas (approved):
Thanks, this is careful work. Two things stand out. The frontmatter now matches `templates/SKILL-TEMPLATE.md` exactly (version, author, domain and tags under `metadata:`, block-style `os`, `install` without the stray `bins: []`). And `test_skill_builder.py:424` actually executes the scaffolded skill instead of grepping the template, so the bundle claim is genuinely tested; checksum labels are anchored to `output_dir` and the verify line in `commands.sh` cds to the same place, so `sha256sum -c` resolves. This is merging.

**Scanner false positive.** The static scan raised a CRITICAL on `tests/test_skill_builder.py` for referencing `hello_bioinformatics.py`, "not present in the repo". That file is generated at run time from the `hello-bioinformatics` name in `demo_spec.json:2` into `tmp_output`, and the test reads it from there. Dismissed; all three test matrix jobs pass on this head.

**Non-blocking suggestions**

1. `SKILL.md:180` still advertises "stdlib only, zero install", but `write_skill_builder_report` now calls `_repro_helpers()` (`skill_builder.py:56`), and importing `clawbio.common.reproducibility` runs `clawbio/common/__init__.py`, which pulls numpy and pandas via `scrna_io`. The docstring blames OpenTelemetry; that lives only in `clawbio/common/audit.py` and is not imported here. Worth correcting both.
2. The generated `environment.yml` gets only the spec's pip deps, so an env rebuilt from the bundle lacks numpy, pandas and the repo itself, and the `python <script>.py` line in `commands.sh` would fail on import. A prerequisite comment would close that.

Approved and merged.

_Automated domain review; parsed by hand today because the agent's output parser dropped this verdict (fixed)._

PR #390, opened 2026-09-04
feat(wgs-prs): add reproducibility bundle

krudo-taco:
Addresses #372 (compute-and-report skill with no bundle; sibling of #379).

## Summary

- Add a reproducibility bundle to `wgs-prs` via `clawbio.common.reproducibility`: portable `commands.sh`, `environment.yml`, `provenance.json`, and `checksums.sha256`.
- Record SHA-256 digests of FASTQ/VCF inputs without storing paths or contents. Non-demo replay requires `INPUT_VCF` or `FASTQ_R1`/`FASTQ_R2`. A free-text `--trait` query is stored only as a SHA-256 fingerprint.
- Implement `--demo` (synthetic VCF, skip sarek) so the catalog command `python skills/wgs-prs/wgs_prs.py --demo` actually runs. Demo defaults to the offline `CLAWBIO-T2D-8` panel.
- Document the output contract, bump the skill to 0.2.0, and run `wgs-prs` tests on the main CI allowlist.

## Skill affected

`wgs-prs`

## Checklist

## Test output

```text
$ python -m pytest skills/wgs-prs/tests -q
50 passed in 12.91s

$ python -m pytest tests/test_generate_catalog.py -q
30 passed in 1.05s

$ uvx --from 'skills-ref>=0.1.1,<1' agentskills validate skills/wgs-prs
Valid skill: skills/wgs-prs

$ python skills/wgs-prs/wgs_prs.py --demo --output-dir <temp>
# writes bridge_report.md, bridge_report.json, and reproducibility/
```

This does not migrate the remaining bundle-less skills in #372. One skill per PR.


manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Thanks, the privacy design here is the strongest part. `provenance.json` stores SHA-256 digests with no paths or contents, the free-text trait is fingerprinted rather than stored (repro_bundle.py:397), replay demands `INPUT_VCF`/`FASTQ_R1` through preflight guards, and `--pgs-id` is shell quoted. `test_selection_precedence_mirrors_stage3` reading the real `_run_stage3` source is a nice touch.

Two things to confirm before this lands.

**1. `--demo` cannot produce a score.** `write_demo_vcf` (wgs_prs.py:578) emits `rs0` to `rs199`, but the default demo panel CLAWBIO-T2D-8 carries rs7903146, rs1801282 and six other real dbSNP IDs. Scoring is by rsID, so overlap is 0/8. gwas-prs skips the panel below the default `--min-overlap 0.5` and exits 0 (gwas_prs.py:1567), which `_run_stage3` records as `success` while no PRS exists. `test_demo_cli_writes_the_documented_output_contract` only asserts that files exist, so it passes on that outcome. Seeding the demo VCF with the eight panel rsIDs, and asserting `variants_used > 0`, would fix both.

**2. `--demo` plus a selector replays as a different score.** The CLI permits `--demo --pgs-id ...` (wgs_prs.py:685), but `_selection` short-circuits on demo (repro_bundle.py:154), so `commands.sh` emits bare `--demo` and provenance records only `{"mode": "demo"}`. Replay then scores CLAWBIO-T2D-8, not what was run. Either reject that combination in the CLI or record the selector alongside demo.

**Non-blocking suggestions**

- The report looks for `prs_output/report.md` (wgs_prs.py:449) while gwas-prs writes `prs_report.md`. Pre-existing, worth a follow-up.

---

manuelcorpas:
Second read of head fc46cffe, nothing has moved since the 5 September comment, so restating what blocks and adding one point that comment missed.

**Blocking**

1. `--demo` produces no score. `wgs_prs.py:578-590` writes synthetic IDs `rs0` to `rs199`. The default panel `CLAWBIO-T2D-8_GRCh37.txt` carries eight real rsIDs (`rs7903146`, `rs1801282`, ...). `gwas_prs.py:656` matches on rsID, so overlap is 0/8, the score is skipped at the 0.5 threshold, and `wgs_prs.py:381-382` records `success`. The demo test at `test_repro_bundle.py:772-797` asserts only that files exist. Seed the demo VCF with the eight panel rsIDs and assert `variants_used > 0`.
2. `--demo --pgs-id X` replays as a different run: `repro_bundle.py:154-155` returns `("demo", None)` before reading the selector, so `commands.sh` is bare `--demo`. Reject the combination or record it.

**New**

3. `sample_id` is stored in plaintext in `provenance.json` (`repro_bundle.py:266`) and in `commands.sh` (line 219). SKILL.md:112 calls these "safe parameters". A lab accession or patient identifier is not safe. Fingerprint it the way `--trait` is, or route it through an env var like the input paths.
4. The new `--panel-id` replay branch at `wgs_prs.py:376-377` has no test; the only path that reaches it is the demo CLI test, which passes on a skipped score.

**Verified fine**

Inputs recorded as role plus SHA-256 only, trait hashed, `${VAR:?}` preflights with quoted expansion, selectors `shlex.quote`d, every stage-3 flag replayed. The ci.yml change is one extra pytest step and cannot weaken a gate. Pre-existing and not yours: `wgs_prs.py:466` reads `prs_output/report.md` but gwas-prs writes `prs_report.md`, so the PRS excerpt never embeds.

Fix 1 to 3 and this merges.

manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Good work, and the replay design is genuinely careful. `repro_bundle.py:340` shell quotes `--pgs-id`, and `tests/test_repro_bundle.py:699` proves it with a `; echo unsafe` payload rather than asserting on the implementation. Sample IDs, trait queries and input paths stay out of the bundle and are re-supplied as environment variables. `tests/test_repro_bundle.py:845` runs the real `--demo` CLI end to end and asserts `variants_used > 0`, so the new `--panel-id` route is exercised against gwas-prs itself, not a mock. I checked the eight demo rsIDs against `skills/gwas-prs/data/CLAWBIO-T2D-8_GRCh37.txt`; positions and effect alleles line up exactly, so the demo scores what it claims. This is merging.

**Non-blocking suggestions**

1. In `write_demo_vcf` (`wgs_prs.py:606`) all four filler allele pairs are transitions, so the demo VCF sits near Ti/Tv 66 against a 2.5 ceiling. Stage 2 will always report QC FAIL, which is why `--demo` forces `fail_fast=False` (`wgs_prs.py:719`). Mixing in transversions would let the demo pass QC honestly.
2. The demo variants at `wgs_prs.py:578` are GRCh37, correctly matched to the GRCh37 panel, but SKILL.md still tells VCF-entry users to supply GRCh38 and the file carries no `##reference` line. One header line would stop anyone reusing it as a GRCh38 template.
3. SKILL.md does not say that `--demo` overrides fail-fast. Worth a sentence.

---
Approved and merged.

PR #392, opened 2026-09-04
fix(deps): bump pillow and aiohttp off known advisories

camlloyd:
## Summary

- Bump two transitive dependencies off known advisories: **pillow 12.2.0 → 12.3.0** and **aiohttp 3.13.5 → 3.14.3**.
- Both sit on the RoboTerri bot's attack surface. The upload allowlist admits `.png/.jpg/.jpeg/.tif/.tiff/.heic/.heif`, which go straight to Pillow's decoders, and the bot runs in public mode — so an unauthenticated stranger can hand a malformed image to those decoders. aiohttp is the Discord adapter's network path.
- Lockfile only (`uv lock --upgrade-package`); neither package is a direct dependency (pillow via matplotlib, aiohttp via discord.py), so `pyproject.toml` is untouched.

## Skill affected

None directly. `cell-detection` and the plotting skills are the consumers of Pillow, and their tests were run explicitly (below).

## Checklist

- [x] SKILL.md is present and complete (YAML frontmatter + methodology) — n/a, no skill added or modified
- [x] Tests included and passing (`pytest -v`) — no new tests; existing suites run against both old and new versions to prove no regression
- [x] Demo data included for reviewers to test — n/a
- [x] No patient/sensitive data committed
- [x] Follows [CONTRIBUTING.md](../CONTRIBUTING.md) conventions

## What this clears

Pillow 12.3.0 fixes ten advisories published July 2026, confirmed against the upstream advisory list, including:

- `GHSA-9hw9-ch79-4vh6` — controlled heap out-of-bounds write in `ImageCmsTransform.apply()` (High)
- `GHSA-6r8x-57c9-28j4` — heap out-of-bounds write in `Image.paste()` / `Image.crop()` (High)
- several decompression-bomb check bypasses in font and image loaders

`uv audit` before and after:

```
before:  Found 107 known vulnerabilities ... in 185 packages   (9 packages)
after:   Found  53 known vulnerabilities ... in 185 packages   (7 packages)
```

**This does not take the report to zero**, and is not meant to. What remains is
build and tooling transitives — gitpython, dulwich, cryptography, pyasn1, idna,
msgpack, setuptools — none of which are on the message-handling path. They are
left for a separate pass so this change stays reviewable.

For reference, the two tools disagree on how to count the same findings
(`uv audit` lists GHSA and PYSEC records for one underlying issue; pip-audit
dedupes across sources but repeats rows per requirement entry). Both flag the
identical set of packages:

```
[...]

Issue #342, opened 2026-08-13
[Bug] scientific-audit CI job cannot fail: bench verdict is discarded by || true


manuelcorpas:
The `scientific-audit` job has reported SUCCESS on every PR for months and the green tick carries no information. It is structurally incapable of failing.

`.github/workflows/ci.yml:268-297`:

```yaml
      - name: Run scientific correctness audit (smoke)
        run: |
          mkdir -p results
          uv run clawbio-bench --smoke --repo . || true   # <- exit code discarded
```

Every subsequent step is `if: always()` or `|| true`. Nothing reads the verdict and nothing exits non-zero.

## What the bench has actually been reporting

Downloaded the `scientific-audit-verdicts` artifact from the most recent main run (suite `20260804_085043`, clawbio commit `6fd9d9f`). The bench own top-level verdict:

```json
"overall": {
  "pass": false,
  "total_cases": 183,
  "total_pass": 160,
  "total_pass_rate": 87.9,
  "blocking_skills": ["bio-orchestrator", "nutrigx-advisor", "pharmgx-reporter",
                      "clawbio-finemapping", "clinical-variant-reporter",
                      "gwas-prs", "cvr-acmg-correctness"]
}
```

Per harness:

| Harness | Pass rate |
|---|---|
| equity-scorer | 100.0 |
| claw-metagenomics | 100.0 |
| cvr-variant-identity | 100.0 |
| bio-orchestrator | 98.1 |
| pharmgx-reporter | 95.5 |
| clawbio-finemapping | 95.0 |
| clinical-variant-reporter | 80.0 |
| cvr-acmg-correctness | 69.2 |
| gwas-prs | 62.5 |
| **nutrigx-advisor** | **0.0** (0/10, `harness_errors: 0`) |

`pass: false`, seven blocking skills, and CI green throughout.

The nutrigx 0% is not a harness error, it is ten genuine failures under `score_incorrect` and `snp_invalid`. #237 (open since 7 May) documents the cause: the rs4988235 lactase polarity is inverted, so the skill gives lactase-persistent users advice to avoid dairy. **A skill scoring 0% on scientific correctness has been shipping in the catalog, with the bench saying so on every single CI run for at least nine days, because one `|| true` swallowed it.**

## Why this matters beyond the one skill

[...]

PR #393, opened 2026-09-04
ci: report dependency advisories on every PR

camlloyd:
## Summary

- Add a `dependency-audit` job to CI that runs `uv audit` against the lockfile on every push and PR, writing the report to the step summary and uploading it as an artifact.
- **Report-only**: findings never fail the build. But a *missing* report does — if `uv audit` stops producing a recognisable report, the job fails loudly rather than rendering an empty summary that reads as "clean".
- One tool, not two. `uv audit` reads `uv.lock` natively with no export step; pip-audit was evaluated and independently confirmed the identical set of affected packages, so running both would be duplicated cost for one signal.

## Skill affected

None — CI configuration only (`.github/workflows/ci.yml`).

## Checklist

- [x] SKILL.md is present and complete (YAML frontmatter + methodology) — n/a, no skill added or modified
- [x] Tests included and passing (`pytest -v`) — n/a, no Python changed; the job's guard clause was exercised directly (below)

## Why this was missing

There was no dependency audit anywhere in CI — `scientific-audit` is the
correctness benchmark, not a security check — and no Dependabot config. So
advisories against locked dependencies went unnoticed until someone ran a scan
by hand. That is how pillow sat at 12.2.0 with ten July 2026 advisories against
it, while decoding images uploaded to the public RoboTerri bot (fixed
separately in #392).

## Why report-only rather than gating

Two reasons, both verified against uv 0.10.12 and recorded in a comment on the
job:

1. `uv audit` **exits 0 even when it reports findings**, so gating would mean
   parsing its output ourselves. Not worth doing while there is a backlog of
   affected packages to clear — and a gate born red is how the
   `scientific-audit` job below ended up carrying `|| true` for months (#342).
2. The subcommand is still **experimental** and warns that it may change
   without notice, so its exit codes and output are not a stable contract to
   build a gate on yet.

Promote to a gate once the backlog is cleared and the subcommand settles.

## Why the guard clause

`|| true` is what makes the job report-only, but on its own it would also mask
the tool breaking — a removed or renamed subcommand exits 0, the summary
renders empty, and the job still shows green. The step therefore greps for the
`Found N known vulnerabilities` line that `uv audit` always prints, and fails
if it is absent.

## Test output

The guard clause, exercised against real and broken outputs:

```
1. real audit output:        -> passes (real report)
2. subcommand removed:       -> FAILS (reports broken)
3. empty output:             -> FAILS (reports broken)
4. a genuinely clean report: -> passes (real report)
```

Case 4 is the one that matters for false alarms: `Found 0 known
vulnerabilities` passes rather than being mistaken for a broken tool.

Validated end to end with `act`, which caught a real bug before it reached CI:
`setup-uv` installs whichever uv is latest (0.12.9 today), and that version
rejects `--all-groups` outright.

```
[CI/dependency-audit] ❗ ::error::uv audit produced no recognisable report - the audit step is broken, not clean
[CI/dependency-audit]  | error: unexpected argument '--all-groups' found
[CI/dependency-audit] ❌ Failure - Main Audit locked dependencies
```

That is the guard clause working as intended: a broken step failed loudly
instead of rendering an empty summary that reads as clean. The command is now
unflagged, which is the only form valid on both uv versions in play (the
preview flag was also renamed, `audit` -> `audit-command`). Re-run under `act`:

```
[CI/dependency-audit]  | Successfully installed uv version 0.12.9
[CI/dependency-audit] ✅ Success - Main Audit locked dependencies
[CI/dependency-audit] ✅ Success - Main Post summary
```

The only remaining `act` failure is `Upload audit report`, which cannot run
locally: act's container has no Node for JS actions
(`exec: "node": executable file not found in $PATH`). GitHub's runners do.

The command the job runs, today, on this branch:

```
$ uv audit
Resolved 186 packages in 7ms
Found 107 known vulnerabilities and no adverse project statuses in 185 packages
```

(That count is against this branch's lockfile. With #392 merged it drops to 53;
note that `uv audit` lists GHSA and PYSEC records separately for the same
underlying issue, so the raw number roughly double-counts.)

YAML validated with `yaml.safe_load`; the job list parses as expected.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,6 +74,14 @@ jobs:
 
       - name: Audit locked dependencies
         id: audit
+        # Deliberately unflagged. setup-uv installs whichever uv is latest
+        # (no version input, no required-version), and this preview subcommand
+        # has already renamed things across versions: `--all-groups` does not
+        # exist on 0.12.9, and `--preview-features audit` became
+        # `audit-command`. Auditing every group is the default, so the bare
+        # command is the one form that works on both. The experimental warning
+        # is cosmetic and lands in the report.
+        #
         # `|| true` keeps findings from failing the build (this job is
         # report-only), but it would also hide the tool itself breaking. So
         # assert the report is a report: `uv audit` always prints a "Found N
@@ -81,7 +89,7 @@ jobs:
         # changed, was removed, or died -- which must fail loudly rather than
         # render an empty summary nobody reads.
         run: |
-          uv audit --all-groups --preview-features audit > audit.txt 2>&1 || true
+          uv audit > audit.txt 2>&1 || true
           if ! grep -qE '^Found [0-9]+ known vulnerabilit' audit.txt; then
             echo "::error::uv audit produced no recognisable report - the audit step is broken, not clean"
             cat audit.txt
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,7 +74,19 @@ jobs:
 
       - name: Audit locked dependencies
         id: audit
-        run: uv audit --all-groups --preview-features audit > audit.txt 2>&1 || true
+        # `|| true` keeps findings from failing the build (this job is
+        # report-only), but it would also hide the tool itself breaking. So
+        # assert the report is a report: `uv audit` always prints a "Found N
+        # known vulnerabilities" line, and if that is missing the subcommand
+        # changed, was removed, or died -- which must fail loudly rather than
+        # render an empty summary nobody reads.
+        run: |
+          uv audit --all-groups --preview-features audit > audit.txt 2>&1 || true
+          if ! grep -qE '^Found [0-9]+ known vulnerabilit' audit.txt; then
+            echo "::error::uv audit produced no recognisable report - the audit step is broken, not clean"
+            cat audit.txt
+            exit 1
+          fi
 
       - name: Post summary
         if: always()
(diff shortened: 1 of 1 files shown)


manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Thanks for this. CI only; no skill code, no clinical logic and no data-source literals are touched, so there is no path here to a wrong scientific result. The details are right where it matters: actions pinned by SHA with version comments, `persist-credentials: false`, the workflow-level `contents: read` inherited, `pull_request` rather than `pull_request_target`, and no secrets in the job. The best part is ci.yml:93, where you refuse to let `|| true` hide a broken tool: asserting that the report is actually a report is exactly the lesson of #342, and the reasoning for staying report-only for now is written down rather than assumed. Merging.

**Non-blocking suggestions** (for later, none of these block the merge)

1. ci.yml:104 tells the reader "this job never fails the build", but ci.yml:96 can exit 1 when the report is unrecognisable. The guard is the right call; it is the summary line that is slightly off. Something like "report-only: findings never fail the build" would match the behaviour.
2. ci.yml:107 runs `cat audit.txt` under `if: always()`. If the job dies before the audit step, for example on a setup-uv failure, that step fails on a missing file and the upload artifact step warns too. `cat audit.txt 2>/dev/null || echo "no report produced"` keeps the real failure visible.
3. setup-uv has no `version:`, so a future rename of the subcommand or of the "Found N known vulnerabilities" line turns this red on unrelated PRs. You have already weighed this; worth a pin if it starts costing anyone time.

The static scan reported nothing on this PR and I agree with it.

---
Approved and merged.

PR #398, opened 2026-09-05
gi-expression: read the scored window from meta, not data.input

boldakov:
Follow-up to #383, same shape of change one field along.

The GI API is normalising `data.input` to the inputs the caller sent, so the server-derived `scored_window` is leaving it. Its typed home is `meta.task_specific_counts.scored_window`, which is returned on every expression response today with the same value (checked against PROD on 2026-09-05: a 25,000 bp submission with `tss_index` 12500 reports `[7901, 17099]` in both places).

The runner now reads meta first and falls back to the `data.input` echo, exactly as it already does for the submitted length. Without this, the day the echo goes, the "Scored window" line in `report.md` would disappear silently, and that line is the only place a user sees which window was scored.

Also relabels a diagnostic in `_write_report` that named `data.input.scored_window` for a value that comes from the summary dict, and corrects one stale sentence in the gi-expression SKILL.md. No skill frontmatter changes, so the catalog is unchanged.

Tests: seven new or rewritten cases in `test_gi_response_shape.py` covering meta-only, echo-only, neither, and wrong-typed in each source. `pytest clawbio/tests/test_gi_response_shape.py clawbio/tests/test_gi_request_construction.py clawbio/tests/test_gi_contract.py clawbio/tests/test_runner.py -m "not integration"` gives 218 passed; four of the new tests fail against the current runner.

We will hold the API change until this is in a published release, as with #383.


manuelcorpas (approved):
Good change, and the vocabulary checks out. `meta.task_specific_counts.scored_window` is a field this API really emits: it is in the recorded live body in `test_gi_response_shape.py` (captured 2026-08-21 against PROD, kept verbatim), and the pre-existing test pins it equal to the `data.input` copy, so the switch moves between two paths the repo has evidence for rather than to an invented one. Moving the `_as_bounds` guard to the read is the right call; the refusal now names the field the value actually came from, and a wrong-typed value in `meta` raises instead of silently falling through to the echo, which is what you want. The fallback keeps either deploy order working, and `test_the_window_survives_the_echo_being_removed` and `test_the_echo_is_the_fallback_when_meta_has_no_window` pin both directions. Merging.

## Non-blocking suggestions

For later, not for this PR.

1. `gi_runner.py:302` still reads `tss_index` from `data.input` alone. By this PR's own argument the index is server-derived when `--tss-index` is omitted, so the same narrowing would render `TSS index None` inside the window line. `meta.task_specific_counts.tss_index` is already pinned equal in the live body.

2. When the window is in neither place, `gi_runner.py:422` drops the line under a truthiness guard. SKILL.md calls that line the only defence against an offset that "does not error, it lies", and a missing line reads the same as a correct one. An explicit "not reported by the API" would be louder.

3. The comments cite a narrowing in the API half of GH#123; that is not verifiable from this repo, but nothing here depends on it being true.

Approved and merged.

_Automated domain review; parsed by hand today because the agent's output parser dropped this verdict (fixed)._

PR #399, opened 2026-09-05
feat(gwas-prs): validate evidence before percentile interpretation

manuelcorpas:
## Summary

Real-input GWAS-PRS reports could emit percentiles without evidence that the exact score, reference and target context support them. Add a deterministic evidence gate before interpretation.

- Require a local hash-bound evidence manifest for real-input research percentiles. Check score/input identity, declared build and strand, compatible complete variants, independent validation and a matching reference; withhold unsupported percentiles and risk labels with structured reasons.
- Preserve raw sums for audit, separate prediction from heritability/causal evidence, carry decisions into every report, and record evidence hashes and portable replay inputs.
- Add a 14-case offline synthetic benchmark and migration/schema documentation. Built-in demo estimates remain explicitly illustrative.

## Skill affected

gwas-prs (0.3.0), with its public CLI flag registration and generated catalogue entry.

Companion #400 preserves the research-only scope in unified reports. Merge #400 first. The actual PRS CLI result envelope was also passed through that consumer successfully: withheld interpretations remained withheld.

## Checklist

- [x] Tests included and passing (focused skill suite)
- [x] Demo data included for reviewers to test (deterministic synthetic fixtures)

## Test output

Test-first development: 167 existing tests passed; 26 initial new contract tests failed before implementation. Additional raw-sum, strand and replay tests also demonstrated failures before their fixes.

Final focused suite:
- Python 3.10: 200 passed
- Python 3.11: 200 passed
- Python 3.12: 200 passed

Public CLI smoke: `clawbio.py run prs --demo --evidence-json ...` completed successfully and withheld unsupported results. A real generated replay command produced byte-identical prs_results.json in a fresh directory. Skill validation and git diff --check passed.

Reviewer demonstration:

```bash
python skills/gwas-prs/benchmark_evidence.py --demo --output /tmp/prs-evidence-review
```

Output: 14/14 cases passed; 2 supported, 12 withheld; false_supported=0, false_withheld=0.

These are synthetic software contracts, not independent scientific validation or a reproduction of Schwaba et al. (Nature 2026; doi:10.1038/s41586-026-10992-9). That paper motivates keeping heritability, prediction and portability distinct.

## Behaviour and limits

Real-input percentiles now default to null without sufficient evidence. Even supported results provide a research percentile, never a clinical risk category. Complete biallelic diploid SNVs and a documented normal reference are required in policy v1; partial sums remain audit-only. Existing output artifacts are protected; replay requires PRS_REPLAY_OUTPUT and, when applicable, PRS_EVIDENCE_FILE.

Evidence is caller-curated: hash consistency and declarations do not independently verify publications, population applicability or clinical utility. There are no new dependencies, scientific algorithms or patient-data network calls.

### DIFF
--- a/clawbio/cli.py
+++ b/clawbio/cli.py
@@ -449,7 +449,7 @@ SKILLS = {
         "description": "GWAS Polygenic Risk Score calculator (PGS Catalog, 3000+ scores)",
         "allowed_extra_flags": {
             "--trait", "--pgs-id", "--panel-id", "--min-overlap",
-            "--max-variants", "--build",
+            "--max-variants", "--build", "--evidence-json",
         },
         "accepts_genotypes": True,
     },
--- a/skills/catalog.json
+++ b/skills/catalog.json
@@ -2265,7 +2265,7 @@
       "name": "gwas-prs",
       "cli_alias": "prs",
       "description": "Calculate polygenic risk scores from DTC genetic data using the PGS Catalog",
-      "version": "0.2.0",
+      "version": "0.3.0",
       "status": "mvp",
       "maturity_tier": "ci-validated",
       "maturity_evidence": {
--- a/skills/gwas-prs/SKILL.md
+++ b/skills/gwas-prs/SKILL.md
@@ -3,7 +3,7 @@ name: gwas-prs
 description: Calculate polygenic risk scores from DTC genetic data using the PGS Catalog
 license: MIT
 metadata:
-  version: 0.2.0
+  version: 0.3.0
   openclaw:
     requires:
       bins:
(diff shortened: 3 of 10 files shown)


manuelcorpas:
Not merging yet: `scientific-audit` is red on this head and the failure is caused by this PR, not by the baseline. The bench reports `gwas-prs: FAIL, coverage_not_flagged: 3` on `gwas_03_no_matching_variants`, `gwas_04_below_coverage_threshold` and `gwas_08_single_variant_below_threshold`. That is the regression the domain review filed as non-blocking suggestion 1: `gwas_prs.py:1595` makes `--min-overlap` inert on every real-input run, so a sub-threshold overlap now produces a raw sum instead of a coverage skip. The bench expects the skip. Restore the coverage gate (or make the withheld-percentile path also emit the coverage flag the bench looks for) and the audit should go green.

Job: https://github.com/ClawBio/ClawBio/actions/runs/33971797680/job/101321854974

manuelcorpas (commented):
**Verdict: NEEDS REVIEW** (audited at 6b53e3c5)

**The record on this PR misstates why CI is red, and so did my first reading of it.** The log line about a missing `aggregate_report.json` is the echoed step script, not the error. The report was found and the baseline check ran. The actual failure is:

```
Bench regressed against the committed baseline:
  - gwas-prs: pass rate fell from 62.5% to 37.5% (-25.0 points)
```

The `fm_12_susie_nonconvergence` harness error is pre-existing and unrelated to this skill; the baseline file already records it as excluded on main.

**The 25-point drop is by design and restoring `--min-overlap` will not clear it.** Per case, the three coverage cases fail on main too (`missing_output`), so they move nothing. The loss is `gwas_01` and `gwas_06`, both flipping to `percentile_incorrect`. The pinned harness invokes the skill with no `--evidence-json`, so this PR correctly withholds the percentile, and the harness scores a withheld percentile as a wrong answer. Its ground truth for `gwas_01` expects 78.8 from a curated EUR reference, which is precisely the un-evidenced percentile this PR argues must not be printed. That is a policy collision between the PR and the pinned bench, not a defect in either. The workflow names the remedy: lower `bench_baseline.json` in this PR and say why in the description, or teach the bench the withheld contract.

**The science is conservative and correct.** Failure on any gate leaves percentile, z-score and risk category null with `method: "withheld_evidence"`; the report prints "Withheld" rather than a caveated number, and a mismatched-ancestry reference is withheld rather than hedged. The gate chain is thorough (hash identity for both scoring file and genotypes, build, strand, duplicates, variant count, complete coverage, allele compatibility, raw-score recompute to 1e-12, validation completeness, reference completeness). No invented performance floor, and the reference doc says so explicitly.

Three asks:

1. **`--min-overlap` is now inert on every real-input run** (`if ... and not evidence_gate:`), while still registered with its old help text. A 3-of-8 coverage raw sum is written to the results JSON, the CSV and the summary with only a JSON label protecting it. Honour it, or retire it explicitly in `--help`, and add a test either way; this is the exact behaviour the bench caught.
2. **Genotype case.** `assess()` compares raw genotype strings against uppercased alleles while `calculate_prs` uppercases and the shared parser does not normalise, so lowercase calls are withheld as `ALLELE_INCOMPATIBLE`. Fails closed, low practical impact for consumer exports, still worth normalising.
3. **Rebase onto the post-#423 baseline** before editing it; the branch is 248 commits behind.

Palindromic A/T and C/G flips remain undetectable by construction, since forward strand is a caller declaration. The doc says "no strand guessing" and should name palindromic SNPs as the residual risk.

PR #401, opened 2026-09-06
ci: move pinned actions to node24 runtimes

camlloyd:
## Summary

- Bump the four pinned GitHub Actions that still declare a `node20` runtime to majors that declare `node24`, silencing the runner's Node 20 deprecation warning on every job.
- Runners already force these actions onto `node24`; this only makes the declared runtime match what is happening, so there is no change in execution environment.
- Verified each new pin's `using:` field and checked every major's breaking changes against how this repo actually uses the action.

| Action | Was | Now |
|---|---|---|
| `actions/checkout` | v4.3.1 | `3d3c42e` v7.0.1 |
| `astral-sh/setup-uv` | v6.8.0 | `20cfd1b` v10.0.1 |
| `actions/upload-artifact` (ci.yml) | v4.6.2 | `043fb46` v7.0.1 |
| `peter-evans/create-pull-request` | v6.1.0 | `5f6978f` v8.1.1 |

`actions/download-artifact` (v8.0.1) and `upload-artifact` in `publish.yml` were already on `node24`; `pypa/gh-action-pypi-publish` is a composite action and unaffected. No `node20` pins remain in `.github/`.

Breaking changes checked against actual usage:

- **checkout v7** blocks fork PR checkout under `pull_request_target` / `workflow_run`. Neither trigger is used here. Only inputs used are `persist-credentials` and `fetch-depth`, both unchanged.
- **create-pull-request v7** renamed `git-token` to `branch-token`. Not set in this repo.
- **setup-uv v9** changed `prune-cache` to default `false`. Caches will be somewhat larger; this is upstream's recommended default to reduce PyPI load.
- **setup-uv v10** skips the cache under the default `enable-cache: auto` for `pull_request_target`, `workflow_run`, `release` and tag-push events, to guard against cache poisoning. The two jobs that rely on caching set `enable-cache: true` explicitly and are unaffected; `publish.yml` takes a cold cache on tag pushes, which costs seconds and cannot fail the run.

## Skill affected

None. CI configuration only, no skill code or `SKILL.md` touched.

## Checklist

- [x] Demo data included for reviewers to test — n/a, no skill changed

## Test output

This change touches no Python, so the relevant checks are that the workflows still parse and that the new pins resolve to `node24` action definitions. Both were verified, along with a representative slice of the suite:

```
$ uv run pytest -q skills/pharmgx-reporter/tests/test_pharmgx.py \
    skills/equity-scorer/tests/test_equity_scorer.py \
    skills/nutrigx/tests/test_nutrigx.py
122 passed in 2.70s
```

```
$ python3 -c "import yaml,glob;[yaml.safe_load(open(f)) for f in glob.glob('.github/workflows/*.yml')]"
all workflows parse OK
```

Each new pin's runtime was confirmed directly from its `action.yml` at the pinned SHA (`using: node24` for all four).

Note for after merge: `bench-leaderboard.yml` and `approve-skills-only.yml` are schedule/dispatch driven, so this PR will not exercise them. A manual `workflow_dispatch` run is worth doing to confirm the `create-pull-request` v8 bump behaves.

### DIFF
--- a/.github/workflows/approve-skills-only.yml
+++ b/.github/workflows/approve-skills-only.yml
@@ -40,7 +40,7 @@ jobs:
   approve:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           persist-credentials: false
 
--- a/.github/workflows/bench-leaderboard.yml
+++ b/.github/workflows/bench-leaderboard.yml
@@ -16,12 +16,12 @@ jobs:
     timeout-minutes: 30
     steps:
       - name: Checkout
-        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           fetch-depth: 1
 
       - name: Install uv
-        uses: astral-sh/setup-uv@d0cc045d04ccac9d8b7881df0226f9e82c39688e # v6.8.0
+        uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
         with:
           enable-cache: true
 
@@ -70,7 +70,7 @@ jobs:
 
       - name: Open PR with refreshed leaderboard
         if: steps.diff.outputs.changed == 'true'
-        uses: peter-evans/create-pull-request@c5a7806660adbe173f04e3e038b0ccdcd758773c # v6.1.0
+        uses: peter-evans/create-pull-request@5f6978faf089d4d20b00c7766989d076bb2fc7f1 # v8.1.1
         with:
           branch: bot/refresh-benchmark-leaderboard
           delete-branch: true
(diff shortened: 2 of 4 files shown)


manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Thanks, this is exactly the right shape for a runtime bump. Every action stays pinned to a full commit SHA with the version in a trailing comment, no tags or floating refs sneak back in, and `persist-credentials: false` is preserved on every checkout that already had it, including the one in `approve-skills-only.yml:43`. Nice cross-check: the `actions/upload-artifact` SHA you introduce in `ci.yml` is the same one already pinned for v7.0.1 at `publish.yml:46`, so that pin is internally consistent with what is on main.

No skills, scripts or reference data are touched, so there is no path here to a wrong scientific or clinical result. The static scan reported nothing and I found nothing to dispute. Verdict is SAFE and it is merging. Note that SHA to tag mapping cannot be verified from this offline checkout; CI is green on this head, which shows the refs resolve and the steps run.

**Non-blocking suggestions**

1. `bench-leaderboard.yml:67` moves `peter-evans/create-pull-request` from v6.1.0 to v8.1.1, two major versions, and that workflow only runs on schedule or dispatch, so PR CI never exercises it. The inputs you pass (`add-paths`, `commit-message`, `title`, `body`) look unchanged, but a manual `workflow_dispatch` after merge would confirm it before the next scheduled refresh.
2. The `astral-sh/setup-uv` jump from v6.8.0 to v10.0.1 is similarly large; worth a glance at the cache key behaviour if you ever see a surprise cache miss.

---
Approved and merged.

PR #402, opened 2026-09-06
ci: let scientific-audit start without waiting for test

camlloyd:
## Summary

- Remove `needs: test` from the `scientific-audit` job so it starts immediately instead of queuing behind `test`.
- Takes CI wall clock from roughly 9 minutes to roughly 6.5, with no change to what any job does.

`scientific-audit` is the longest job in every run. It checks out, installs its own toolchain and `clawbio-bench`, runs the smoke bench and writes its own `results/`. It reads no artifact and no output from `test`, so the dependency was ordering only.

Measured against run `34035644836`:

| Job | Duration |
|---|---|
| `scientific-audit` | 394s |
| `test (3.12)` | 140s |
| `test (3.13)` | 124s |
| `test (3.11)` | 118s |
| `skill-lint` | 26s |
| `benchmark` | 18s |

Sum of job time was 14.2 min; wall clock was ~9 min, essentially `test` then `scientific-audit` back to back.

The trade-off, stated plainly: the bench now also runs on pushes whose tests fail, where the edge previously skipped it. That costs some runner minutes. Wall clock looks like the scarcer resource here, and the baseline-regression gate at the end of the job does not depend on `test` having passed. Happy to restore the edge if you'd rather keep the minutes.

## Skill affected

None. CI configuration only.

## Checklist

- [x] Tests included and passing (`pytest -v`) — n/a, no test or skill code changed

## Test output

One-line workflow change; the check is that it parses and that the job graph is what we intend.

```
$ python3 -c "import yaml;yaml.safe_load(open('.github/workflows/ci.yml'))"
parses OK

$ git diff --stat
 .github/workflows/ci.yml | 1 -
```

This PR's own CI run is the real verification — `scientific-audit` should now start alongside `test` rather than after it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -429,7 +429,6 @@ jobs:
 
   scientific-audit:
     runs-on: ubuntu-latest
-    needs: test
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
         with:


manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Thanks, this is a good catch on the CI graph.

I checked what `scientific-audit` actually consumes before agreeing it can float free of `test`. The job at `.github/workflows/ci.yml:429` does its own `actions/checkout`, its own `uv sync --locked` and its own pinned install of `clawbio-bench`; there is no `download-artifact` step and nothing it reads is produced by the `test` matrix. So `needs: test` was pure ordering, and dropping it cannot change what the audit measures.

It also closes a small hole rather than opening one. With `needs: test` in place, a red test matrix left the audit **skipped**, and a skipped required check does not report a failure. PR #403 in today's batch shows exactly that shape: `test (3.12)` failed and `scientific-audit` came back SKIPPED. After this change the audit runs on its own merits.

The real gate is untouched: the final "Fail on regression against the committed baseline" step is still not `if: always()`, and a missing `aggregate_report.json` still exits 2 rather than passing quietly.

No skill code changed here, so there is no clinical or vocabulary surface to review. The static scan reported nothing and I found nothing to add to it. Merging.

**Non-blocking suggestions**

For later, not for this PR: the audit will now burn a runner on PRs whose tests are already red, which is the price of the fix and is probably worth it. If that ever bites, a `concurrency` group on the job is a cheaper lever than putting the `needs` back.

---
Approved and merged.

PR #403, opened 2026-09-06
test: drop tests/__init__.py from the four region-fetch skills

camlloyd:
## Summary

- Delete the empty `tests/__init__.py` from the four region-fetch skills, which is what currently stops `pytest skills/` from collecting at all.
- No CI change and no test behaviour change; this only removes the blocker so CI *can* later run the skill tree in one pass instead of a hardcoded 12-skill allowlist.

`pytest skills/` aborts today. Each of these four skills carries an empty `tests/__init__.py`, so pytest imports all four conftests under the same module name `tests.conftest`, and the second to load kills collection:

```
ValueError: Plugin already registered under a different name:
skills/gwas-catalog-region-fetch/tests/conftest.py=<module 'tests.conftest'
from 'skills/eqtl-catalogue-region-fetch/tests/conftest.py'>
```

The four files are zero bytes — package markers only, nothing imports through them. Removing them keeps the paths unique and the whole tree collects.

## Context

`ci.yml`'s `test` job names 12 skills explicitly. 88 skills ship Python tests. The other 76 are only covered by the "skills changed in this PR" step, so on a push to main they never run — a shared dependency bump can regress them silently. Collapsing the allowlist into one whole-tree run is the fix, and this collision is the first thing in the way. That rewrite is deliberately **not** in this PR.

## Skill affected

`eqtl-catalogue-region-fetch`, `gwas-catalog-region-fetch`, `ld-1000g-region-compute`, `ukb-ppp-region-fetch` — test scaffolding only, no skill logic or `SKILL.md` touched.

## Checklist

- [x] Demo data included for reviewers to test — unchanged

## Test output

Each affected skill's suite, run on its own, is byte-identical before and after:

```
$ uv run pytest -q skills/eqtl-catalogue-region-fetch/tests/   22 passed
$ uv run pytest -q skills/gwas-catalog-region-fetch/tests/     5 failed, 10 passed, 1 skipped
$ uv run pytest -q skills/ld-1000g-region-compute/tests/       7 passed, 1 skipped
$ uv run pytest -q skills/ukb-ppp-region-fetch/tests/          28 passed, 2 skipped
```

**The 5 `gwas-catalog-region-fetch` failures are pre-existing and unrelated to this change** — verified by checking out `main` with the `__init__.py` still present and getting the identical `5 failed, 10 passed, 1 skipped`. They are left alone here rather than folded into an unrelated PR.

Collection across the whole tree, which was previously impossible:

```
$ uv run pytest -q --co skills/
4484 tests collected in ~4s
```

(One further collection error remains locally in `clinical-variant-prioritizer`, from a broken `torch`/`libgfortran` link in my conda env, not from the repo.)


manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Thanks for this. The direction looks right, and I checked the things that usually make an `__init__.py` removal bite. All four files are the empty git blob (`e69de29b`), so no code is lost. `pytest.ini:3` sets `--import-mode=importlib`, which does not need package markers to keep module names unique, and these suites reach their skill module through `sys.path.insert` rather than a package-relative import (`skills/ukb-ppp-region-fetch/tests/test_ukb_ppp_region_fetch.py:24`). It also brings the four into line with `wgs-prs` and the three `nfcore-*` wrappers, which already ship `tests/conftest.py` with no `__init__.py`. There is no scientific or clinical surface in this diff.

One thing holds it up. CI on `204bd595` has `test (3.12)` at FAILURE, with 3.11 and 3.13 cancelled behind it, and #402 from the same minute is fully green, so the base was healthy. The scanner's single warning about that run is accurate, not a false positive. Each of these suites keeps its live-network gate in `tests/conftest.py` (`skills/ld-1000g-region-compute/tests/conftest.py:13`), so a change to how the suite is imported can quietly alter which tests actually run, and "green because nothing was collected" is the failure mode worth ruling out here. Could you paste the 3.12 failure, or push a rerun once it is understood? If it turns out to be an unrelated flake, this is fine to merge.

### Non-blocking suggestions

23 other skills still carry `skills/*/tests/__init__.py`. Worth the same sweep in a follow-up, once this one is confirmed clean.

---

manuelcorpas:
Correction to the audit comment above: the red `test (3.12)` run is maintainer-side, not yours.

Your change made the "Run tests for skills changed in this PR" step run `skills/eqtl-catalogue-region-fetch/tests` for the first time. That suite imports `pysam` at setup (`test_eqtl_catalogue_region_fetch.py:131`), and the CI environment installs `uv sync --locked --all-extras` with no pysam extra in `pyproject.toml`, so every test errors before it starts. Main never runs those four suites (they are in `pytest.ini` but not in the per-path CI list), which is why main is green and this PR is red.

The diff itself is right: all four files are the empty blob, `--import-mode=importlib` needs no package marker, and the suites reach their module through `sys.path`. I have opened an issue for the missing extra. Once that lands and this is rebased, the run should go green with no change on your side. Thanks for the patience.

manuelcorpas:
#411 is merged: pysam is now the `region-fetch` extra, so `uv sync --all-extras` in CI installs it. A rerun of the old run still fails on `No module named 'pysam'` because it reuses the original merge commit. Rebase onto main and the four suites should execute and pass, as they did for me locally (22, 15, 7 and 28 passed). Nothing else needed from you.

manuelcorpas (approved):
Updated the branch onto main from the maintainer side; with pysam now in the region-fetch extra all four suites collect and pass on head fe1b9d7e (3.11 to 3.13 green). Automated audit by ClawBio pr-audit plus a maintainer read.

PR #404, opened 2026-09-06
perf(genome-compare): share parsed genomes and the demo run across tests

camlloyd:
## Summary

- `skills/genome-compare/tests/test_genome_compare.py` was the slowest step in CI's `test` job at **52s**. This removes two duplications inside it, taking local runtime from **42.2s to 11.6s** (~72%).
- Behaviour is unchanged: same 28 tests, same 47 assertions, all passing.
- No production code touched — this is a test-only change.

`--durations` showed no single slow test; the cost was spread evenly at 1.3–4s across all 28, which is the signature of repeated work rather than one expensive assertion. Two things were being redone:

| Duplication | Before | After |
|---|---|---|
| Parsing the 4.8 MB + 4.9 MB gzipped 23andMe reference files | 14 parses across 28 tests | 2, in a module-scoped fixture |
| The full `run_comparison` demo pipeline | 4 identical runs at ~4.5s each, each asserting on a different part of the same output | 1, shared |

### Why module scope is safe here

Checked rather than assumed, since a shared mutable structure across tests is how this kind of change goes wrong:

- Neither `compute_ibs` nor `estimate_ancestry` mutates any parameter it is handed — no item assignment, `.pop`, `.update`, `.setdefault`, `del`, or `.clear` on `geno_a`, `geno_b`, `positions_a`, `genotypes`, `aims_panel` or `populations`.
- No test writes to the parsed structures either.

So no defensive copies are needed. A comment in the file records this, because if it ever stops being true these fixtures must go back to per-test parses.

The shared demo run uses `tmp_path_factory` rather than `tmp_path`: the latter is function-scoped and cannot be requested from a module-scoped fixture.

### Context

This is the first of the levers found while looking at CI runtime, and the only one that is a code change. The larger one is structural and **not** in this PR: CI runs 12 separate `uv run pytest` invocations, and merging them into one saves a further ~39% (measured: 104.0s across 12 invocations vs 63.0s for the same paths in one) purely by not paying twelve interpreter cold starts.

Note this does not by itself shorten CI wall clock — `scientific-audit` at 394s is the critical path — but it cuts runner minutes across all three Python versions in the matrix, and it matters if the `test` job ever grows to cover the whole skill tree.

## Skill affected

`genome-compare` — tests only. No change to `genome_compare.py` or `SKILL.md`.

## Checklist

- [x] Demo data included for reviewers to test — unchanged

## Test output

Before, on `main`:

```
$ uv run pytest -q skills/genome-compare/tests/test_genome_compare.py
28 passed in 42.22s
```

After, on this branch:

```
$ uv run pytest -q skills/genome-compare/tests/test_genome_compare.py
28 passed in 11.58s
```

Equivalence checks, both run against `main` and this branch:

```
collected tests   28  ->  28
assert lines      47  ->  47
ruff errors        6  ->   5   (one fewer; the remainder are pre-existing)
```

Remaining `parse_23andme_extended` / `run_comparison` call sites in the file: **3** — the two fixture parses and the one shared demo run.

### DIFF
--- a/skills/genome-compare/tests/test_genome_compare.py
+++ b/skills/genome-compare/tests/test_genome_compare.py
@@ -12,6 +12,7 @@ import sys
 from pathlib import Path
 
 import numpy as np
+import pytest
 
 sys.path.insert(0, str(Path(__file__).parent.parent))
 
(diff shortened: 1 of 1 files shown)


manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

**Domain review: genome-compare**

Nicely scoped performance change, and the reasoning in the diff is sound. I checked the safety claim rather than taking it on trust: `compute_ibs` (`genome_compare.py:116-171`) and `estimate_ancestry` (`genome_compare.py:224` onwards) only read the dicts they are handed, and `_ibs_at_site` copies the allele list locally at `genome_compare.py:107`, so module scope for `reference_genome` and `demo_patient` is safe. `load_aims_panel` is still called per test, so the marker dicts are not shared either.

Every scientific assertion survives verbatim: the IBS band and the >400k overlap floor at `tests/test_genome_compare.py:253-255`, the proportions summing to 1.0, and the EUR top-ancestry check. No assertion was loosened to fit the refactor. The note at `tests/test_genome_compare.py:17-27` explaining why module scope holds, and what to do if that changes, is exactly the right thing to leave behind.

No blocking issues, so this is merging. The points below are for later.

**Non-blocking suggestions**

1. `test_end_to_end_demo` now shares `demo_comparison` with the three report tests (`tests/test_genome_compare.py:243`), so the pipeline runs once per module rather than twice. Fine for these fixed demo inputs, but if you want the e2e to prove an independent clean run, it should keep its own invocation.
2. `json` and `numpy` at the top of the file look unused. Pre-existing, worth a tidy next time you are in here.

---
Approved and merged.

PR #408, opened 2026-09-09
dnasp: VB-parity corrections + VCF input (v0.5.0)

drdaviddelorenzo:
## Summary

One combined change to the `dnasp` skill, `v0.4.0 -> v0.5.0`, in three parts.

### Part 1  -  VB-parity corrections (neutrality, divergence, HKA)

Every formula in the skill was checked module by module against the DnaSP 6
Visual Basic source (shared under MIT by J. Rozas). Six discrepancies were found
and corrected, plus one edge case; two of the six came from a Codex adversarial
review of the first fixes.

- **Fu's Fs**  -  was using the *lower* Ewens tail `P(K <= H)`; Fu (1997) and
  DnaSP use the *upper* tail `P(K >= H)`. Sign was inverted (rp49: +33.7 ->
  -22.3). Also moved the tail sums to log space (the exact Stirling numbers
  overflow `float` at n ~ 171, so `fufs` used to raise `OverflowError` on large
  alignments), and removed a "significant at 0.02" label  -  `S'` is not a
  P-value because theta_pi is estimated. `Fs` now returns n.a. for a monomorphic
  locus (was a sentinel).
- **Fu & Li D / F with outgroup**  -  replaced a `D*/F*`-style numerator with
  mismatched variance coefficients by the DnaSP `FULI.vb` forms
  (`D = (eta - a_n*eta_e)/sqrt(u_D*eta + v_D*eta^2)`, matching `c_n`, `v_D`,
  `u_D`, `v_F`, `u_F`). rp49 D: +1.41 -> -1.64, now sign-consistent with
  D* = -1.52.
- **Fu & Li F\***  -  k-bar is now computed over the same orientable-site mask as
  eta (DnaSP accumulates `rp1` only inside `SitioIesInformativo`), not over a
  wider complete-deletion mask.
- **SFS**  -  biallelic sites only, as in DnaSP (`FULI.vb` gates on
  `contot == 2`); multiallelic columns are excluded and counted.
- **Ts/Tv**  -  one change per biallelic segregating site (DnaSP
  `Mutational.vb::Mod31Compute_1`), not a sum over sequence pairs. The old count
  scaled with sample size and could not be read as mutational bias.
- **Divergence shared / private / fixed**  -  ported from
  `Divergencia.vb::Mod3BuscaShareFixDifferences`. A site polymorphic in one
  population only is now *private*, not shared; `Ss` / `Sx` are mutation counts
  (`Sf` stays a site count) and reconcile to eta exactly.
- **HKA**  -  replaced the home-grown k-locus Poisson MLE (whose bisection
  silently returned T = 500 on unbracketed input) with DnaSP's two-locus
  closed-form solver (`HKA.vb::HKAResolEcuacion` + `HKAJiCuadrado`): quadratic in
  theta_1, HKA (1987) variances, df = 1. Unsolvable input now reports an error,
  not a fabricated divergence time. The `--hka-file` format changes to
  `locus n S L_poly D [L_div] [chrom]`.

### Part 2  -  VCF input (`--vcf`)

Adds a multi-sample VCF as an input format, so `dnasp` can be run on variant
calls without an external FASTA conversion step. Follows DnaSP 6
(`Formularios/multifilefrmvcf.vb::readvcf`).

- **`--vcf FILE`**  -  converts a multi-sample VCF to one aligned MSA per `CHROM`
  and runs the requested analyses on each. `--region CHROM` restricts to one;
  `--vcf-merge` pools every CHROM into a single MSA (a deliberate genome-wide
  summary only  -  it mixes unlinked regions and is not valid for pi / Tajima's D
  / the SFS).
- **Biallelic SNPs only.** Indels, multi-base REF/ALT and multiallelic sites are
  skipped and counted in the run summary.
- FORMAT must begin with `GT`. FILTER is ignored (as in DnaSP). Ploidy is
  validated per genotype  -  a VCF mixing haploid and diploid calls, or
  containing a polyploid call, is rejected.
- **Diploid phased** (`|`): two haplotype rows per sample (`<sample>_h1`,
  `<sample>_h2`). **Unphased** (`/`): homozygous -> the allele on both rows;
  heterozygous -> both rows gap (DnaSP cannot resolve phase). `.` -> gap.
  **Haploid** (`GT` = `0`/`1`): one row per sample.
- **Per-CHROM sample set**: a sample whose GT starts with `.` at the first
  retained variant of a CHROM is dropped from that whole MSA. Population
  assignments (`--pop-file`) are resolved per CHROM from that CHROM's own
  samples; unassigned samples are reported. `load_pop_file` now accepts
  space-separated files (DnaSP's `.SG.txt`).

This runs the **standard** modules on a VCF-derived alignment. It is not a port
of DnaSP's RAD engine (no Achaz F\* variances, no per-MSA Mean row, no
`.Hetz` / `.Btw` / `.GFlow` outputs).

### Part 3  -  fixes from the DnaSP 6.12 for Windows validation run

The classic example alignments distributed with DnaSP (rp49, COII_Apes,
DmelOSRegion, Ex_n1) and a synthetic HKA case were run in DnaSP 6.12 for
Windows and in the skill under identical settings. Every mismatch was traced
to the VB source (or DnaSP's own help file) and the DnaSP behaviour reproduced;
nothing was tuned to fit.

- **Fu & Li D\*/F\* and D/F (outgroup)**  -  the singleton / external-mutation
  site counts cap at 1 per site, not 1 per derived allele
  (`FuLiIntraespecifico`, `Mod12BusqMutacExternas`); D/F scale by S, not eta
  (`14cdaf6`). Ex_n1 and rp49 now exact.
- **Rm**  -  DnaSP's own interval reduction (`CODIGO2.vb::RecombinacionRM`)
  instead of an optimal stabbing algorithm; rp49 7 -> 10, DmelOSRegion 5 -> 7,
  both exact (`14cdaf6`).
- **R2**  -  divides by S, not eta (`Dnasp_51.vb::CalculaStatisticsInArray`);
  rp49 now exact (`c657d0d`).
- **`--genetic-code {standard,vertebrate-mitochondrial}`** for `mk`, `kaks`
  and `codon` (DnaSP's "mtDNA Mammals" table); under the standard code the
  TGA columns of a mitochondrial gene were silently dropped as stops
  (`14cdaf6`).
- **McDonald-Kreitman**  -  counts per nucleotide site, not per codon
  (`3ae5c50`), then a line-by-line port of the within-species routine
  `McDonaldK.vb::BuscaSitiosReemplazamientoMK` (`b002f16`): a site that
  segregates in the ingroup and is also fixed against the outgroup counts in
  both tables; a site carrying k bases contributes k - 1 changes; DnaSP's
  "complex codons" are excluded and reported (`n_complex_codons`). An interim
  codon-level rule (`748c593`) was superseded. COII: Pn 10, Ps 69, Dn 8, Ds 56,
  NI, alpha and Fisher's P all match DnaSP 6.12.
- **Ka/Ks**  -  ingroup-vs-outgroup pairs when `--outgroup` is set
  (`ed4031b`, `2395150`); synonymous sites leave stop-codon alternatives out of
  the denominator (`SINONIMO.vb::ComputeFoldPos`), nonsynonymous sites are
  3 x codons analysed - S, and Ks/Ka are one Jukes-Cantor correction of mean
  differences over mean sites as in `EntrePobsMod.vb::PolDivergenceOut`, not a
  mean of per-pair corrected distances (`ec3d622`). COII: S 168.222, Ks
  0.83011, Ka 0.02561 match DnaSP exactly.

`SKILL.md` and `docs/index.md` are in line with all of the above; `skills/catalog.json`
is regenerated.

## Skill affected

`dnasp`

## Validation

- **DnaSP's own shipped output**  -  DnaSP 6.0.60 ships computed results for its
  RAD example (`rp49_5regions`) and its VCF examples
  (`Data_Example_Diploid{Phased,Unphased}.vcf`, `Data_Example_Haploid.vcf`).
  On the clean cases every comparable statistic matches DnaSP exactly: S, eta,
  Hap, Hd, VarHd, pi, theta_W, theta_k, Tajima's D, **Fu & Li D\***, Fu's Fs, R2,
  Z_nS. On `Scaffold_2` of the phased VCF example: pi 0.1, theta_W 0.281870,
  S 1, Hd 0.1, Tajima's D -1.164391, Fu & Li D\* -1.539595, Fu's Fs -0.879302,
  R2 0.217945, Z_nS n.a. Sample inclusion matches DnaSP on every scaffold of
  every example.
- **Fu & Li F\*** differs by design: DnaSP's RAD/multi-MSA mode uses the Achaz
  (2009) variance, the skill uses Simonsen et al. (1995) of the standard mode.
  D\*, which is variance-form-independent, matches everywhere.
- Documented limitation: scaffolds built to exercise edge cases (ALT=".",
  `A,T` multiallelic, `TG/CG` equal-length MNP) differ from DnaSP by a site or
  two, because DnaSP's RAD engine splits the MNP per position and drops the
  multiallelic column downstream while `parse_vcf` is biallelic-SNP-only.
- **DnaSP 6.12 for Windows, side by side**  -  the classic example alignments
  (rp49, 34 sequences; COII_Apes, 11 sequences + outgroup; DmelOSRegion, 17
  sequences; Ex_n1) and a synthetic HKA case, run in both programs under
  identical settings, plus the shipped RAD/VCF output above. **170 summary
  statistics compared across the polymorphism, LD, recombination, divergence,
  neutrality, McDonald-Kreitman, Ka/Ks, codon-usage and Fst groups: 163 agree
  to the precision DnaSP prints (four to five significant figures), 3 differ
  only in the last displayed digit** (Variance of Hd; COII nonsynonymous sites
  and omega, both DnaSP's own rounding), **and the remaining 4 are Fu & Li F\*
  in DnaSP's multi-alignment mode**, the variance-form difference above. Not
  counted: 19 rows where DnaSP prints nothing comparable (Ts/Tv has no panel
  in this build; ENC is per sequence; Fay & Wu H and Zeng E are normalised
  only; ZnS n.a. at small n). The full table is the manuscript's
  Supplementary Table S1 (`publication/validation/validation_table.md` in the
  DnaSP-ClawBio validation bundle, notes 1-17 record every mismatch found and
  how it was resolved).
- Documented, not changed: DnaSP keeps codons with internal stop codons as a
  21st amino acid, the skill skips them; visible only under the wrong genetic
  code (COII under the standard code: MK Pn 11 vs 13, everything else exact).

### Review follow-ups

Two review passes since the branch was opened:

- **Automated audit** (`fc372f4`): report/TSV caveat that VCF per-site π and θ_W
  are per variant site, not per base (matches DnaSP 6, which prints the same
  caveat); `SKILL.md` Fu's Fs `Fs > 0` row corrected; `HKAStats.note` split from
  `HKAStats.error`.
- **Human review** (`978ef5d`): unphased-heterozygote site loss is now counted
  (`n_unphased_het_sites`) and warned; `--window` on a VCF is flagged as a
  SNP-index window (POS is not used to place columns); `SKILL.md` states plainly
  that the unit tests are literature/VB-derived on synthetic fixtures and the
  DnaSP-output comparison lives with the manuscript; "planned for v0.5.0" wording
  removed from the v0.5.0 change.

## Checklist

- [x] SKILL.md present and complete (corrected formulas + VCF input section + flag reference)
- [x] Tests included and passing  -  `355 -> 440`, `pytest -v` green
- [x] Demo data included (`--demo` unchanged; DnaSP's own VCF examples are the
      fixtures in the manuscript validation bundle; unit tests build small
      synthetic VCFs inline; new `hka_synthetic.tsv` example in the bundle)
- [x] No patient/sensitive data
- [x] Follows CONTRIBUTING.md  -  em/en dashes replaced with the repo's spaced
      hyphen style (per `manuelcorpas`)

## Test output

```
$ cd skills/dnasp && pytest -q
440 passed in ~2s
```

New/changed tests cover: Fu's Fs tail direction and the monomorphic case;
large-n Ewens sums (n = 180/200/300); Fu & Li outgroup D/F sign and the k-bar
site mask; the SFS biallelic gate; per-biallelic-site Ts/Tv and its sample-size
invariance; the divergence state-count table and its reconciliation to eta; the
HKA two-locus solver (neutral -> chi2 = 0, selection -> chi2 >> 0, unsolvable ->
error); phased/unphased/haploid genotype decoding; one MSA per CHROM; `--region`
filter; indel / multi-base / multiallelic skipping; unphased-het -> gap;
missing-GT sample exclusion; FILTER ignored; `--vcf-merge`; mixed-ploidy and
polyploid rejection; per-CHROM sample sets; per-CHROM population map; colliding
sanitised CHROM names getting distinct output directories; singleton caps
per site for Fu & Li; DnaSP's Rm reduction; R2 over S; the mitochondrial
codon table; every McDonald-Kreitman worked example from DnaSP's help file
(codons 1-3, 7-9, 10-12, 13-15, 16-18, 22-24), dual-status sites, complex
codons; DnaSP's stop-excluded synonymous-site counts (TGT, TAT, AGT) and the
corrected-ratio Ka/Ks summary.


manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Careful VB-parity work, and several of these are genuine safety fixes: the Fu & Li outgroup numerator was inverted before and now matches Fu & Li (1993) in `compute_fu_li_outgroup` (dnasp.py:1786 onward); Fu's Fs moves to the upper tail S' = P(K ≥ H) and drops the unfounded "significant at 0.02" claim; Ts/Tv no longer scales with sample size; HKA sets an error instead of fabricating a T̂. The VCF tests read like real parser semantics, not tests written to match the code.

One thing to settle before this leaves draft.

**π and θ_W from `--vcf` are per variant site, not per base.** `parse_vcf` emits one column per retained SNP, and π is k / L_net (dnasp.py:2620), so L_net is the SNP count. The report prints that as "Nucleotide diversity (π)" (dnasp.py:2949) with no caveat, so a VCF run gives a π orders of magnitude above true per-base diversity. Tajima's D and Hd are count based and unaffected, which makes the inflated π easy to miss. Could you either take a total analysed-sites count for rescaling, or print an explicit warning on the VCF path? Either is fine.

**The Fu's Fs table is now self-contradictory.** SKILL.md:610 (on main, untouched here) still says "Fs > 0 ... More haplotypes than expected", directly below your corrected "Fs << 0 ... More haplotypes". docs/index.md has the right version.

**Non-blocking suggestions**

- `HKAStats.error` doubles as an informational note when two positive roots exist. A separate `note` field would keep `error is not None` meaning "test not run".

---

drdaviddelorenzo:
Thanks for the review — all three points are addressed in fc372f4.

**π / θ_W per variant site (the blocking one).** This matches DnaSP 6 by design rather than being an oversight. `Formularios/multifilefrmvcf.vb` builds one MSA column per retained variant and sets `netsites = nnucs` (~line 1609), then computes `ThetaWattNuc = th2 / netsites` (~1747) and `ThetaK = thktot / netsites` (~1732) — i.e. per retained variant site, exactly as `parse_vcf` does. DnaSP handles the inflation by printing a header caveat in its output ("ThetaWattNuc, ThetaWatt and TajimaD are calculated using the total number of segregating sites", ~1078 / ~3538), not by rescaling.

So rather than add a `--callable-sites` denominator that DnaSP doesn't have (which would be a deliberate divergence from the parity target), the skill now carries the same caveat: a note in `report.md` and a `#` comment row in `results.tsv` on the `--vcf` path, plus a line on stdout, with SKILL.md and docs/index.md updated. Count-based and scale-free statistics (S, η, H, Hd, Tajima's D, Fu & Li D*/F*, R2) are called out as unaffected. A `--callable-sites` opt-in can be added later if there's demand.

**Fu's Fs table contradiction.** Fixed — the `Fs > 0` row now reads "Fewer haplotypes than expected", consistent with the corrected `Fs << 0` row and with docs/index.md.

**`HKAStats.error` as an informational note.** Split out: new `HKAStats.note` carries the "more than one positive-θ solution" remark; `error is not None` now strictly means the test was not run. Report and stdout updated.

Tests 373 → 394 (388 → 394 for the VCF branch alone). Still a draft pending a side-by-side run against DnaSP 6.12 for Windows on the classic example alignments.

manuelcorpas:
Read at head 9f3ea3dd while it is still a draft, so nothing here is a formal block; it is what I would need before undrafting.

The statistic corrections are right and I checked them against the papers rather than the VB. Fu and Li D with outgroup now uses `c_n = 2(n a_n - 2(n-1))/((n-1)(n-2))`, `v_D = 1 + a_n^2/(b_n + a_n^2) (c_n - (n+1)/(n-1))`, `u_D = a_n - 1 - v_D`, and D* and F* take Simonsen 1995's coefficients; that matches Fu and Li 1993 equations 22 to 25 and 31 to 34, and the old code had been borrowing v_F from the no-outgroup case. Fu's Fs moving to the Ewens upper tail is Fu 1997's S'. Ts/Tv per biallelic site, biallelic-only SFS and the HKA b_n variance term all hold. The revert-sensitive tests are real (`test_variance_includes_second_harmonic_term`, `test_Fs_decreases_with_H`, `test_triallelic_site_excluded`, and so on).

Three things before undraft.

1. **VCF path reports per-variant, not per-base.** `parse_vcf` emits one character per retained SNP (`dnasp.py:640`), `analyse_region` then computes `Pi = k / L_net` with `L_net` from `complete_deletion` (`:3004-3013`), so `L_net` is the SNP count. The report prints `Nucleotide diversity (π)` at `:3343` with no caveat; the FASTA path's `L_net` includes invariant sites. Same column label, different denominator, and `--window` (`:4574`) becomes a SNP-index window because POS is discarded. Either rescale using the region length, or label every VCF-path number as per segregating site in the report itself.
2. **Unphased heterozygotes vanish silently.** `:786` writes two gaps for any `0/1`, and `complete_deletion` drops the whole column. On an unphased diploid VCF a site survives only when every sample is homozygous, which is a biased site set, and the summary at `:4528-4532` counts indels, multiallelics and non-GT rows but not this. Print the count.
3. **"VB parity" is a port of the source, not a comparison against DnaSP output.** No new expected value in the tests comes from a DnaSP run on a named dataset; the HKA Adh example in SKILL.md is not tested against the published chi-square. Say so in SKILL.md, or add one case with the DnaSP number and its provenance.

Small: `SKILL.md:657` still has `Fs > 0 ... More haplotypes than expected` two rows below the corrected `Fs << 0` line, and SKILL.md says coalescent simulation is "planned for v0.5.0" in the v0.5.0 PR. No `np.random`, no URLs, catalog regenerated, version consistent.

drdaviddelorenzo:
Thanks, this is a thorough read. All four points are now addressed — the automated-audit thread above covers `fc372f4`, and `978ef5d` handles the rest of yours.

**1. VCF reports per-variant, not per-base.** This is deliberate parity, not a slip: DnaSP 6 computes `netsites = nnucs` on the VCF-derived MSA (`multifilefrmvcf.vb:~1609`) and `ThetaWattNuc = th2 / netsites` (`~1747`), and prints a header caveat rather than rescaling. The skill now carries the same caveat in `report.md` and `results.tsv` and documents it (`fc372f4`). On the new sub-point — `--window` on a VCF is indeed a SNP-index window, since POS is not used to place columns. `main()` now warns when `--window` is set on the VCF path and `write_report` labels the sliding-window section; SKILL.md and docs say so (`978ef5d`). I kept `--window` working (labelled) rather than disabling it, since a run of consecutive SNPs is still a useful view; happy to disable it on the VCF path instead if you'd rather.

**2. Unphased heterozygotes vanish silently.** Fixed the "silently" part: `VCFPopulation.n_unphased_het_sites` counts retained SNPs with at least one unphased het (both haplotypes gapped, as DnaSP does), the run summary prints it with a warning that complete deletion then drops those columns and that on heterozygous unphased data most sites are lost, and SKILL.md/docs state the same. The gapping behaviour itself matches DnaSP; the transparency was the gap (`978ef5d`).

**3. "VB parity" is a port, not a comparison against DnaSP output.** Agreed. SKILL.md now says plainly that the unit suite asserts literature- and VB-derived values on synthetic fixtures and is not a bulk regression against DnaSP GUI output; the real DnaSP-output comparison (classic alignments vs DnaSP 6.12 for Windows, plus DnaSP's own shipped results for the RAD and `Data_Example_*.vcf` examples) is maintained with the associated manuscript as a supplementary table. I also removed an overclaim in an earlier draft that a DnaSP-shipped case was checked in the suite, and clarified that the `Adh`/`5flank` HKA rows are layout placeholders, not a published result (`978ef5d`).

**Smalls.** `SKILL.md` `Fs > 0` row corrected to "Fewer haplotypes than expected" (`fc372f4`). "planned for v0.5.0" is gone from the v0.5.0 change — the docs roadmap now lists VCF as delivered and simulation p-values as a future item (`978ef5d`).

Tests `355 -> 397`. Still a draft pending the DnaSP 6.12 for Windows side-by-side.

manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Careful, well documented work, and the statistical corrections hold up when checked against the primary literature rather than the VB source.

- `dnasp.py:1797` onwards: Fu & Li D = (η − aₙ·η_e)/√(u_Dη + v_Dη²), with c_n, v_D, u_D, v_F, u_F as in Fu & Li (1993). The previous numerator was sign inverted, and the interpretation tables in `SKILL.md:340` and `docs/index.md` have been flipped to match, so D < 0 now means excess singletons.
- `dnasp.py:2499`: Fu's Fs now uses the upper tail S′ = P(K ≥ H), which is Fu (1997). Dropping the 0.02 threshold instead of quoting a P-value for an estimated θ_π is the right call.
- `compute_hka`: I re-derived the quadratic from the three moment equations and coef_a, coef_b and coef_c are algebraically correct; df = 1 and the bₙ variance terms match HKA (1987). Setting `error` rather than inventing a T̂ is the right failure mode.

Static scan: two INFO notes only, both flagged as pre-existing. Nothing to dispute.

1. `compute_hka`, in the `if theta2 <= 0.0: continue` loop: θ₁ and θ₂ are constrained positive but T is not, so high S with low D can report a negative divergence time. Worth rejecting or labelling T̂ < 0.
2. `--vcf-merge`: `--help` and `SKILL.md:155` say it is not valid for π, Tajima's D or the SFS, but `report.md` prints those figures without repeating the caveat. Consider carrying it through as `variant_sites_only` does.
3. `load_hka_file` skips rows with fewer than five fields, so an old four-column file reports "requires exactly two loci" rather than a format error.

---

---

manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Careful work, and the corrections here are the substantive kind. Three I checked and agree with:

- **Fu & Li outgroup D/F**: the old numerator (`eta_e - eta/An`) carried the opposite sign to Fu & Li (1993), and the old interpretation table matched the inverted statistic. `D = (S - aₙ·η_e)` and `F = (k̄ - η_e)` with the 1993 coefficients have expectation zero, as they should.
- **Fu's Fs**: S′ is the upper tail P(K ≥ H) in Fu (1997), so the old CDF gave Fs the wrong sign. Dropping the "significant at 0.02" claim rather than faking a simulated null is the right call.
- HKA sets `error` instead of inventing a T̂ when no positive-θ root exists, and the VCF report states plainly that π and θ_W are per variant site.

The MK tests keyed to DnaSP's own help-file examples (codons 1-3, 10-12, 13-15, 16-18) come from outside the implementation, which is what makes them worth something.

One worth fixing: `dnasp.py:5112`, the `--hka-file` argparse help still reads `TSV: locus<TAB>S<TAB>D<TAB>n`, but `load_hka_file` now needs `locus n S L_poly D [L_div] [chrom]`. Anyone following `--help` writes the old order; the `len(parts) < 5` skip then drops those rows silently and the only clue is "HKA requires exactly two loci". A warning per skipped row would help.

- `docs/index.md`, "D and F statistics": "Only the 'total mutations' mode is implemented" contradicts the paragraph above it and the code, which scales by S.
- `kaks` switches to ingroup-versus-outgroup when `--outgroup` is set for another module. It is labelled in the report and on stdout; consider putting the mode in the Ka/Ks table row itself.

---
Approved and merged.

PR #409, opened 2026-09-10
fix(wgs-prs): embed the gwas-prs report in bridge output

krudo-taco:
## Summary
- read the `prs_report.md` artifact that `gwas-prs` actually writes
- add a regression test that exercises the WGS-to-PRS report handoff
- correct the documented WGS-PRS output path

## Why
The WGS-PRS bridge starts `gwas-prs` with `--output <bridge>/prs_output`. `gwas-prs` writes `<bridge>/prs_output/prs_report.md`, but the bridge report only looked for `report.md`. A successful scoring run could therefore be presented as unavailable in the aggregate medical-AI report.

## Validation
- `uv run --with pytest python -m pytest -q skills/wgs-prs/tests/` (37 passed)
- `python -m compileall -q skills/wgs-prs/wgs_prs.py`
- `uvx --from 'skills-ref>=0.1.1,<1' agentskills validate skills/wgs-prs`
- `git diff --check`

## Scope
This only aligns an existing local file handoff and its declared output name. It does not change PRS calculation, clinical interpretation, QC thresholds, or patient-data handling.

### DIFF
--- a/skills/wgs-prs/SKILL.md
+++ b/skills/wgs-prs/SKILL.md
@@ -80,7 +80,7 @@ metadata:
     description: Ti/Tv ratio, Het/Hom ratio, variant counts, pass/fail
   - name: vcf_qc/canonical_pass.vcf.gz
     description: Normalised, filtered canonical VCF ready for PRS scoring
-  - name: prs_output/report.md
+  - name: prs_output/prs_report.md
     description: PRS narrative report from gwas-prs
   - name: prs_output/tables/scores.csv
     description: Per-trait PRS scores, percentiles, and risk categories
--- a/skills/wgs-prs/tests/test_wgs_prs.py
+++ b/skills/wgs-prs/tests/test_wgs_prs.py
@@ -18,7 +18,7 @@ SKILL_DIR = Path(__file__).resolve().parent.parent
 if str(SKILL_DIR) not in sys.path:
     sys.path.insert(0, str(SKILL_DIR))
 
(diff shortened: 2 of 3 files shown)


manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Good catch, and the fix is right. `gwas-prs` writes its narrative report to `prs_report.md` (`skills/gwas-prs/gwas_prs.py:1584`, and the same name is documented in `skills/gwas-prs/SKILL.md:114`), so the bridge's old lookup for `report.md` could never match. Before this change the bridge fell through to the `_PRS scores not available (scoring stage did not complete)._` line at `skills/wgs-prs/wgs_prs.py:459` even on a run where scoring had completed normally, which is exactly the kind of silent failure that reads as a result. Updating `SKILL.md:83` in the same PR keeps the declared outputs honest.

Merging.

**Non-blocking suggestions**

1. The new test at `skills/wgs-prs/tests/test_wgs_prs.py:32` writes `prs_report.md` itself, so it pins the bridge to its own spelling rather than to what `gwas-prs` emits. It would still pass if both sides drifted together. Lifting the filename into one shared constant that both skills import, or asserting against the path `gwas-prs` actually produces, would make the coupling real.

2. The `else` branch at `wgs_prs.py:459` infers stage failure from a missing file. Keying that message off the recorded stage status instead would stop a future rename from being reported to the user as a failed scoring stage.

Nothing from the static scan to answer on this one; it reported no findings, and I found none it missed.

---
Approved and merged.

PR #410, opened 2026-09-10
fix(qc): enforce VCF missing-call limits on both execution paths

krudo-taco:
## Summary
- Compute the configured missing-call rate after both bcftools and Python statistics paths, persisting the metric in JSON.
- Count unreadable and partial genotype calls as missing while preserving complete haploid and polyploid calls.
- Strip VCF record line endings before parsing terminal GT fields, with a regression asserting 2 heterozygous and 1 homozygous-alt calls. This addresses the GT-only fixture audit finding.
- Document that missingness is measured across sample calls in the raw input VCF before filtering, as requested in the non-blocking audit note.

## Skill affected

WGS-PRS VCF QC, through `clawbio/common/vcf_qc.py`.

## Checklist

- [x] Existing SKILL.md remains complete; this enforces its configured missing-call limit.
- [x] Regression tests included; the affected WGS-PRS suite passes (48 tests).
- [x] Synthetic GT-only fixture included in the regression test (2 het / 1 hom-alt).
- [x] Focused change follows contribution conventions; local full-suite limitations are recorded below.

## Test output

- New GT-only regression before fix: failed, expected 2 heterozygous calls but got 0.
- `uv run --with pytest python -m pytest -q skills/wgs-prs/tests/test_vcf_qc.py`: 27 passed.
- `uv run --with pytest python -m pytest -q skills/wgs-prs/tests`: 48 passed.
- `uv run python clawbio.py list`, Python compile check, and `git diff --check`: passed.
- Broader WGS-PRS plus common tests: 170 passed, 1 failed (`test_generate_report_header_requires_skill_version`). The same failure reproduces on unchanged PR head `6a9ce97` (7 passed, 1 failed in that file).
- Full local collection also stops at an existing duplicate `tests.conftest` registration between eqtl-catalogue and gwas-catalog region-fetch tests. A passing full local suite is not claimed.

## Scope

This enforces an existing QC limit and corrects missing-call accounting; it does not alter filtering thresholds, PRS scoring, or clinical interpretation. The follow-up fixes valid GT-only VCF records without widening the QC change.

### DIFF
--- a/clawbio/common/vcf_qc.py
+++ b/clawbio/common/vcf_qc.py
@@ -305,7 +305,7 @@ class VcfQC:
             for line in fh:
                 if line.startswith("#"):
                     continue
-                parts = line.split("\t")
+                parts = line.rstrip("\r\n").split("\t")
                 if len(parts) < 10:
                     continue
                 ref, alt = parts[3], parts[4]
@@ -354,7 +354,7 @@ class VcfQC:
             for line in fh:
                 if line.startswith("#"):
                     continue
-                parts = line.rstrip("\n").split("\t")
+                parts = line.rstrip("\r\n").split("\t")
                 if len(parts) < 10:
                     continue
                 format_fields = parts[8].split(":")
--- a/clawbio/common/vcf_qc.py
+++ b/clawbio/common/vcf_qc.py
@@ -41,6 +41,11 @@ from typing import Optional
 log = logging.getLogger(__name__)
 
 
+def _is_complete_genotype(value: str) -> bool:
+    """Whether a GT contains one or more fully specified numeric alleles."""
+    return bool(re.fullmatch(r"\d+(?:[|/]\d+)*", value))
+
+
 # ---------------------------------------------------------------------------
 # Configuration
 # ---------------------------------------------------------------------------
@@ -92,6 +97,7 @@ class QcResult:
     hom_alt_count: int = 0
     het_hom_ratio: Optional[float] = None
     filtered_out: int = 0
+    missing_genotype_rate: float | None = None
 
     def summary(self) -> str:
         lines = [
@@ -167,6 +173,7 @@ class VcfQC:
             stats = self._python_stats(input_vcf)
 
         self._populate_result(result, stats)
+        result.missing_genotype_rate = self._missing_genotype_rate(input_vcf)
         self._evaluate_pass_fail(result)
 
         metrics_path = output_dir / "qc_metrics.json"
@@ -302,8 +309,10 @@ class VcfQC:
                 if len(parts) < 10:
                     continue
                 ref, alt = parts[3], parts[4]
-                fmt_idx = parts[8].split(":").index("GT") if "GT" in parts[8] else -1
-                gt_raw = parts[9].split(":")[fmt_idx] if fmt_idx >= 0 else "."
+                format_fields = parts[8].split(":")
+                fmt_idx = format_fields.index("GT") if "GT" in format_fields else -1
+                sample_fields = parts[9].split(":")
+                gt_raw = sample_fields[fmt_idx] if 0 <= fmt_idx < len(sample_fields) else "."
 
                 # SNP vs indel
                 is_snp = len(ref) == 1 and len(alt) == 1 and alt != "."
@@ -318,8 +327,8 @@ class VcfQC:
                     metrics["number of indels:"] += 1
 
                 # Het / hom
-                gt = re.split(r"[|/]", gt_raw.replace(".", "0"))
-                if len(gt) == 2:
+                gt = re.split(r"[|/]", gt_raw)
+                if _is_complete_genotype(gt_raw) and len(gt) == 2:
                     if gt[0] != gt[1]:
                         metrics["number of heterozygous SNPs:"] += 1
                     elif gt[0] != "0":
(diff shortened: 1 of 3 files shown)


manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

**Domain review: wgs-prs / vcf_qc**

Good change on the whole. Replacing the `"GT" in parts[8]` substring test with a lookup in the split FORMAT list (`clawbio/common/vcf_qc.py:312`) fixes a real bug, since `PGT` and similar keys used to be read as `GT`. Guarding against a short sample column (`vcf_qc.py:315`) is also right: VCF allows trailing FORMAT fields to be dropped. And `max_missing_rate` was dead config until this PR.

**Blocking**

`_python_stats` still splits the raw line without stripping the newline (`vcf_qc.py:308`), so when `GT` is the last FORMAT field the genotype string is `"0/1\n"`. The new `_is_complete_genotype(gt_raw)` guard at `vcf_qc.py:331` then returns False for every call and het/hom counts fall to 0 for any GT-only VCF, which is exactly what `plink2 --export vcf` writes. `_missing_genotype_rate` does strip (`vcf_qc.py:357`), so the same file would report 0.000 missing alongside 0 hets. The two new assertions at `skills/wgs-prs/tests/test_vcf_qc.py:157` and `:168` expect 0 and pass either way, so they do not cover this. `line.rstrip("\r\n")` in `_python_stats` should be enough.

**Non-blocking suggestions**

- Missingness is measured on `input_vcf` (`vcf_qc.py:176`) while the other stats come from the filtered canonical VCF on the bcftools path, and every sample column counts. A joint-called VCF is therefore gated on cohort missingness though the bridge scores one sample. Worth a line in the docstring.
- The mocked stats dict at `test_vcf_qc.py:198` uses `number of heterozygous SNPs:` as an SN key. bcftools stats reports het and hom counts in the PSC section (`nHets`, `nNonRefHom`), not SN, so `_parse_bcftools_stats` never sees them on the real path. Pre-existing, but worth a follow-up.

---

manuelcorpas:
Read at head 6a9ce979. The gate is right: one call site after the two paths merge (`vcf_qc.py:176`), one comparison (`:398`, strict `>`), equality tested at `test_vcf_qc.py:224-235`, structured reason plus `passes_qc` false, and both `test_missing_calls_fail_the_configured_missingness_limit` and `test_missingness_just_over_the_threshold_fails_on_bcftools_path` go red if the enforcement is deleted. Thank you for closing a config field that was dead on main.

One thing blocks merge. `vcf_qc.py:308` does `parts = line.split("\t")` without stripping the newline, while `:357` does `line.rstrip("\n")`. For a GT-only FORMAT column the genotype arrives as `"0/1\n"`, `re.fullmatch(r"\d+(?:[|/]\d+)*", ...)` is False, and `:331` skips every call, so het and hom counts are zero on the Python path for any plink-style VCF. For the record, main was wrong on the same input in the other direction: `re.split` produced `["1", "1\n"]` and every GT-only call counted as heterozygous. Your change trades all-het for all-zero, and the het/hom-ratio gate is blind either way. Every current fixture uses `GT:DP` (`test_vcf_qc.py:24-28`), which is why nothing caught it.

Fix: `line.rstrip("\r\n")` at `:308`, plus a GT-only fixture asserting the het and hom counts, written red first.

Non-blocking: missingness is measured on the raw input, not the filtered canonical VCF, and on a multi-sample file it is cohort missingness. Worth one sentence in SKILL.md.

The audit comment above cites `:157` and `:168` for the zero-expecting assertions; the lines are `:210` and `:221`.

manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Good catch, and the fix is in the right direction on both counts.

`max_missing_rate` (`vcf_qc.py:59`) was declared as a threshold but never read anywhere in the tree, so the advertised missingness gate was a no-op. Wiring it into `_evaluate_pass_fail` closes a silently skipped check.

The genotype parsing fix matters more than the title suggests. The old `gt_raw.replace(".", "0")` turned `1/.` into `1/0` and counted it as heterozygous, and a GT-only final column kept its newline, so `1/1` split to `["1", "1\n"]` and every homozygote was counted as a het. Both inflated het/hom and could flip the QC verdict. `_is_complete_genotype` (`vcf_qc.py:44`) and the guard at `vcf_qc.py:331` fix that, and the tests cover haploid, polyploid, partial, `PGT`-only and truncated FORMAT cases rather than only the happy path.

Computing missingness on the raw input rather than the filtered canonical VCF is a defensible choice and you have documented it in `SKILL.md`, so a reader will not misread the number.

**Non-blocking suggestions** (for later, not for this PR):

1. `_missing_genotype_rate` returns `None` when no sample columns are present (`vcf_qc.py:371`), and unlike Ti/Tv that produces no warning. A sites-only VCF would skip the gate quietly. A `warnings.append` would match the existing pattern.
2. Separately, `_populate_result` (`vcf_qc.py:377`) reads `number of heterozygous SNPs:` and `number of homozygous SNPs:` from `bcftools stats`. The SN section does not emit those keys; het counts live in PSC. If so, the het/hom gate has never fired on the bcftools path. Worth its own issue.

The static scan reported no findings here and I found nothing to add to it.

---
Approved and merged.

PR #412, opened 2026-09-10
Add a keyless Colab agent demo and repair the guided tutorial

manuelcorpas:
## Summary

The first-skill Colab notebook installs into one environment but runs with another, expects the removed summary.json output, and starts online enrichment during an introductory demo. Replace that route with one pinned, key-free PharmGx lesson using bundled synthetic data and explicit --no-enrich execution.

The learner runs the baseline, inspects CYP2C19 phase ambiguity, removes its three markers from a copy of the teaching input, verifies NOT_TESTED while DPYD stays unchanged, and downloads reports plus checksummed tutorial provenance. Setup failures stop execution and repeated runs preserve earlier outputs. Add a notebook execution check on relevant PRs and weekly to catch dependency and instruction drift.

## Skill affected

Existing PharmGx tutorial only. No scientific calling logic or skill contract changed. The notebook uses ClawBio commit 7290841dfc9c7e817c12af38a2dd1479f0babe8e and its lockfile. The displayed/exported report has presentation separators removed and the standard research-use notice included.

## Checklist

- [x] Existing SKILL.md retained; no skill API changes
- [x] Notebook tests included and passing
- [x] Existing synthetic demo data used
- [x] No patient data or credentials committed
- [x] Existing notebook URL preserved

## Test output

Red: original notebook failed both new contract checks (unpinned bootstrap and decorative rules). The first execution also exposed the report notice mismatch.

Green: CLAWBIO_TEST_COLAB=1 uv run --no-project --python 3.12 --with pip --with nbformat --with nbclient --with ipykernel python -m unittest discover -s tests -p test_colab_tutorial.py -v

2 tests passed. Notebook execution includes setup, both actual skill runs, uncertainty/missing-evidence assertions and ZIP creation. First setup used a fresh pinned checkout and environment on macOS. Fresh Linux notebook execution also passed: https://github.com/ClawBio/ClawBio/actions/runs/34468499522. The repaired notebook loads successfully in the Colab UI.

Real Google Colab acceptance remains pending Google sign-in, so this PR stays draft. No Colab runtime success or startup-time guarantee is claimed. Paired docs PR: https://github.com/ClawBio/docs/pull/4. Land the notebook first, then the docs entry.

## Live agent prototype

Adds a terminal-style notebook with keyless google.colab.ai, three restricted tools, an eight-step model loop, real action logs and evidence downloads. The existing deterministic tutorial remains available.

Validation: four local agent tests pass; actual baseline and missing-CYP2C19 analyses pass through the new tool adapters. In Colab, Gemini selected read_skill and received the real skill text; its next call returned HTTP 429 heavy load. A complete live agent conversation remains unverified. Remains draft pending that acceptance check.


manuelcorpas (commented):
**Verdict: NEEDS REVIEW** — policy, not code (audited at 1e3df6c7)

Keyless as claimed: no `sk-`, key, token or Bearer anywhere in the diff, and model access goes through the viewer's own Google session. The install path is properly pinned (fetch by 40-character SHA, detached checkout, `uv` pinned to an exact version, `uv sync --locked`), with no `curl|sh` and no unpinned git URL, and the workflow pins its actions by SHA with `persist-credentials: false`. No `np.random` and no saved outputs; a test enforces that cells ship unexecuted.

`tutorial-first-skill.ipynb` is clean end to end: bundled demo patient only, `--no-enrich` on every run, and an assertion that the enrichment count is zero. Nothing genomic leaves the runtime.

**The one thing to decide consciously:** `tutorial-agent.ipynb` loads your real published 23andMe file and the agent loop sends the conversation history to Gemini on every turn, with that history containing the inspection result including full per-gene profiles. It is disclosed twice in the notebook and the data is public and yours, so this is not a leak. It is still the first ClawBio notebook that ships a named individual's genotype-derived output to a third-party model by default, and the pinned PharmGx cannot read that file's coordinates, so the demo's content is the agent explaining a withheld result. Worth an explicit sign-off rather than an auto-merge.

**Do not close this without merging or repointing docs.** The published lesson at docs.clawbio.ai already links the notebook at a commit that exists only inside this branch, because the docs entry merged on 10 September and this did not. If the branch goes away, that public link 404s.

Two smaller notes: `tutorial-agent.ipynb` is never executed by any workflow, only compiled and string-matched, and it is linked from nowhere on main. The pinned commit is valid but 199 commits behind, which the notebook documents as deliberate.

PR #416, opened 2026-09-11
fix(omics): checksum every advertised output artifact

krudo-taco:
## Summary

Write the environment and RO-Crate artifacts before the checksum manifest, so the five files already advertised by the omics evidence mapper are actually covered. Previously the manifest was generated while the final two files did not exist. The checksum manifest intentionally does not hash itself.

## Skill affected

`omics-target-evidence-mapper`.

## Checklist

- [x] Existing SKILL.md output contract is preserved and now fulfilled.
- [x] Regression included; the complete skill suite passes (2 tests).
- [x] Deterministic synthetic output fixture included.
- [x] Uses the existing shared reproducibility helpers and focused contribution conventions.

## Test output

- Regression independently recomputes each SHA-256 and requires exactly: `evidence.json`, `report.md`, `metadata.json`, `reproducibility/environment.yml`, and `ro-crate-metadata.json`.
- `uv run --with pytest python -m pytest -q skills/omics-target-evidence-mapper/tests`: 2 passed.
- Python compile and `git diff --check`: passed.
- The regression uses a deterministic RO-Crate writer stub to isolate the checksum contract; it does not assert third-party RO-Crate serialization behavior. No full repository test pass is claimed.

### DIFF
--- a/skills/omics-target-evidence-mapper/omics_target_evidence_mapper.py
+++ b/skills/omics-target-evidence-mapper/omics_target_evidence_mapper.py
@@ -360,14 +360,13 @@ def main() -> None:
     write_json(evidence_path, evidence)
     write_text(report_path, report)
     write_json(metadata_path, metadata)
-    write_checksums([evidence_path, report_path, metadata_path], output_dir, anchor=output_dir)
-    write_environment_yml(
+    environment_path = write_environment_yml(
         output_dir,
         env_name="clawbio-omics-target-evidence-mapper",
         pip_deps=["requests", "rocrate"],
         python_version="3.11",
     )
-    write_ro_crate(
+    ro_crate_path = write_ro_crate(
         output_dir,
         skill_name="omics-target-evidence-mapper",
         skill_version="0.1.0",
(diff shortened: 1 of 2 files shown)


manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Thanks, this is a clean fix to a real gap. Before this change `write_checksums` ran at `omics_target_evidence_mapper.py:363`, ahead of `write_environment_yml` and `write_ro_crate`, so two advertised artefacts shipped without a checksum. Moving the call to last and threading the returned paths through is the right shape: both helpers already return a `Path` inside `output_dir` (`clawbio/common/reproducibility.py:176` and `:350`), so the labels in the manifest stay correctly relative to the anchor.

The new test replaces a bare `assert True` and checks the manifest by exact set equality, then re-hashes each file, which is the assertion that actually matters. Good that it also asserts the manifest does not list itself.

No clinical or classification logic is touched, nothing leaves the machine, and the static scan reported no findings that I could contradict.

1. `write_ro_crate` packages `output_dir` via `rglob` at `reproducibility.py:312`. Because the crate is now built before the manifest exists, `reproducibility/checksums.sha256` is no longer listed in the crate's `hasPart`. That trade is unavoidable in one pass and you picked the better side of it, but it may be worth a line in SKILL.md.
2. The test stubs `write_ro_crate` at `tests/test_omics_target_evidence_mapper.py:82`, so the real rocrate path is not exercised. The stub matches the real return value today; a note saying it exists to avoid the rocrate dependency would keep that honest.
3. `write_checksums` skips missing files silently (`reproducibility.py:153`), so a failed write shrinks the manifest quietly at runtime. Pre-existing, worth raising upstream.

---
Approved and merged.

PR #423, opened 2026-09-11
refactor(fine-mapping): delegate the SuSiE engine to sushie 0.20

camlloyd:
## Summary

- Replace the hand-rolled numpy IBSS in `fine_mapping_core/susie.py` with an adapter around `sushie.infer_ss.infer_sushie_ss` (mancusolab/sushie 0.20, single-ancestry summary-stats mode). ABF is unchanged. **SuSiE-inf is also unchanged and still runs this skill's own numpy IBSS** — sushie has no infinitesimal-component model to delegate to, so two engines with different priors now coexist under one skill. SKILL.md says so explicitly and the PR title has been corrected accordingly.
- Fixes the phantom-credible-set pathology: on the demo locus (2 causal variants, L=10) `main` reports **10** credible sets, 8 of them 200-variant purity-0 phantoms, and 8 variants at PIP ≥ 0.1; this branch reports exactly **2** single-variant sets on the causal indices. Cost: ~0.06 s → ~4 s per fit (JAX JIT) and a heavy dependency set behind a new optional `fine-mapping` extra; ABF still runs without it.
- Keeps the `run_susie` contract that the external `biostochastics/clawbio_bench` fine-mapping driver imports directly (`w`, `max_iter`, `tol`, `min_purity`, `mu`/`mu2`), pinned by `TestSuSiEBenchmarkContract`. `w` seeds sushie's `effect_var`; `min_purity`/`coverage` are forwarded to sushie's `purity`/`threshold` so fit-time pruning and `credible_sets.py` agree; non-convergence warns and flags like `susieR::susie_rss`.

## Skill affected

fine-mapping (SKILL.md 0.2.0 → 0.3.0; `skills/catalog.json` regenerated)

## Review round 2 — what changed

Addressing the maintainer review and the automated audit. Merged `main` (the branch is now `MERGEABLE`; `uv.lock` was regenerated rather than hand-resolved).

**1. Conflicts with main.** Resolved. `pyproject.toml` auto-merged; `uv.lock` was re-cut from `main`'s lockfile with `uv lock`, so CI has a merge ref to build.

**2. The bench gate.** `ci.yml` (`scientific-audit`) and `bench-leaderboard.yml` now `uv sync --locked --all-extras`, so sushie is present and SuSiE cases run instead of erroring. `bench_baseline.json` is re-cut for `clawbio-finemapping`: **95.0 → 90.0**, measured against this head with clawbio-bench 0.1.1 (18/20 pass, 1 harness error), not estimated. `scripts/check_bench_baseline.py` reports delta 0.0 against it.

The baseline file asks that a lowered number be justified rather than quietly updated, so both moved cases are recorded in its `_comment`:

- `fm_12_susie_nonconvergence` — was `harness_error` (excluded) on `main`, now a scored fail. Its own rubric accepts "a top-level status flag that downstream callers must consult", which is exactly what the adapter returns (`converged=False` plus a `RuntimeWarning`, mirroring `susieR::susie_rss`). The grader only accepts a raise.
- `fm_14_susie_moment_mislabeled` — was a pass, now `harness_error`. Its locus is `z=[5,5,0]` under identity LD, so the single credible set holds two uncorrelated variants, purity 0, and is pruned; PIPs are reported over surviving sets only, so the case gets no `mu` to score. It also expects susieR's z-unit `r·z` moments where sushie reports standardised effect-size moments.
- `fm_20_susie_inf_est_tausq_ignored` fails identically on `main` — pre-existing, untouched here.

Both are grading gaps rather than engine defects, and restoring 95.0 once either is regraded upstream is noted in the file. Filing them with the bench maintainer is left to you.

**3. Silent project-wide `cryptography` downgrade.** Fixed, not justified — the downgrade was not safe. `liknorm 2.0.0` pins `cffi<2.0` and `cryptography>=48` needs `cffi>=2.0`, so pulling liknorm in through `glimix-core` dragged cryptography from **48.0.0 to 45.0.7 for every user of the lock, extras or not**, and `pip-audit` reports advisories against 45.0.7 that 48.0.0 does not carry (PYSEC-2026-2141, PYSEC-2026-35/36, GHSA-537c-gmf6-5ccf). sushie's summary-statistics path imports `glimix_core` but never `liknorm` — verified, glimix-core only loads it for the GLMM/exponential-family models the individual-level path uses. So liknorm is dropped from the resolution via a `[tool.uv] override-dependencies` entry with an unsatisfiable marker, and cryptography stays at 48.0.0 with cffi 2.0.0. The explicit `glimix-core` pin existed only to force `liknorm>=2.0` and is gone.

**4. No numerical known-answer test.** Added `TestSuSiEKnownAnswer` plus `tests/fixtures/demo_susie_known_answer.json`, recording from the seeded demo locus at this commit: engine version, convergence and iteration count, credible-set count and per-row leads, PIP magnitudes by index, credible-set membership and purity, and the lead `alpha`/`mu`/`mu2` moments. A sushie bump that reorders alpha rows against `CSIndex`, shifts PIP magnitudes or rescales the moments now fails loudly instead of passing the shape-only assertions. `test_two_signals_recovered` required only one of the two causal variants (`60 in top5 or 140 in top5`); it now requires both, each at PIP > 0.8 and each leading its own credible set.

**Smaller points from the review**

- **Loci with fewer variants than L.** `run_susie` now clamps `L` to the variant count and warns, so the CLI default `--max-signals 10` runs on a 9-variant locus as the old IBSS did, instead of tripping sushie's `min_snps` guard. `min_snps` is also held at or above `L` for `L > 100`. Test and SKILL.md gotcha added.
- **`jax.config.update("jax_enable_x64", True)` at call time.** Now read, set only if off, and restored in a `finally`, so a shared interpreter's other JAX users keep their precision. Pinned by `test_x64_flag_is_restored`.
- **SKILL.md purity wording.** Corrected to the **minimum** absolute pairwise `|r|` (Wang 2020 §3.2), which is what `credible_sets._purity` has always computed; the table said "average pairwise r²".
- **The 0.80 → 0.87 internal composite.** Agreed, that is a model change (EM-estimated effect variance replacing a fixed Wakefield prior), not a bug fix. Removed from the summary above rather than left standing as a correctness claim.

**From the automated audit**

- `CSIndex k → alpha row k-1` is now pinned by `test_credible_set_membership_matches_recording`, asserting the demo's credible-set leads are indices 60 and 140 with size 1 and purity 1.0.
- `report.md` now carries a provisional banner **above** the credible-set tables when `converged` is false, per the skill's own Gotcha 4, with the iteration budget in it. Four tests cover present/absent/ABF/budget.
- `run_finemapping` now rejects `coverage`/`min_purity` outside the open interval on the SuSiE path before loading sumstats or LD, instead of failing inside `run_susie`. ABF still accepts `coverage=1.0`.

**Also in this branch**

- CodeQL (`security/code-scanning/778`): dropped the unused `import fine_mapping_core.susie as susie_mod` that duplicated the `from ... import` in the same test.
- `plt.cm.get_cmap` was removed in matplotlib 3.9, so `--demo` crashed while writing figures. This is pre-existing on `main` and not caused by this branch, but it made the demo this PR asks reviewers to run fail, so it is fixed here (3 call sites → `plt.get_cmap`).

## Checklist

- [x] SKILL.md is present and complete (YAML frontmatter + methodology) — methodology, dependencies and CLI updated; `## Gotchas` now has 6 entries (absent on `main`, conformance checklist requires ≥3); 271 lines, under the 500 limit
- [x] Demo data included for reviewers to test — `uv sync --extra fine-mapping && uv run --extra fine-mapping python skills/fine-mapping/fine_mapping.py --demo --output /tmp/finemapping_demo`

## Test output

```
$ uv run --extra fine-mapping --with pytest python -m pytest skills/fine-mapping/tests/ tests/ -q
418 passed in 45.51s
```

Demo, end to end (all figures written, no traceback):

```
$ uv run --extra fine-mapping python skills/fine-mapping/fine_mapping.py --demo --output /tmp/fm_demo_check
  SuSiE (sushie 0.20) converged in 176 iterations
  Results:
    Lead variant: rs_demo_0061 (PIP = 1.0000)
    Credible sets: 2
      L1: 1 variants, coverage=100.0%, purity=1.000
      L2: 1 variants, coverage=99.2%, purity=1.000
    Variants with PIP >= 0.1: 2
```

Bench gate, against this head:

```
$ uv run clawbio-bench --smoke --repo . --harness finemapping
  Overall: 18/20 pass (90.0%) [1 harness errors]

$ uv run python scripts/check_bench_baseline.py <report> --baseline bench_baseline.json
| Harness             | Baseline | This run | Delta |
| clawbio-finemapping | 90.0%    | 90.0%    | 0.0   |
```

`uv sync --locked --all-extras` succeeds, so the lockfile matches `pyproject.toml` as CI checks it.

## Behaviour changes to note

- `--prior-variance` is an EM seed under SuSiE, not a fixed prior (ABF unchanged).
- `--coverage` and `--min-purity` must lie strictly in (0, 1) for the SuSiE path (sushie constraint); `1.0`/`0` now raise a clear `ValueError` at the CLI. ABF still accepts `coverage=1.0`.
- `--max-signals` above the variant count is clamped to the variant count with a `RuntimeWarning`.
- `alpha`/`mu`/`mu2` contain only signals sushie kept as credible sets; a null locus returns zero rows, and a locus whose only set is impure returns all-zero PIPs.
- `run_susie` now also returns `max_iter`, so a report can say "did not converge in 1 of 500".
- SuSiE runs need `uv sync --extra fine-mapping` (`uv run --extra fine-mapping ...`); CLAUDE.md and SKILL.md commands updated accordingly.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -283,14 +283,16 @@ python skills/recombinator/recombinator.py \
   --father einstein-g0 --mother anning-g0 --offspring 3 --generation 1
 
 # SuSiE fine-mapping — credible sets and PIPs from GWAS summary stats
+# SuSiE (with --ld) delegates to the sushie package: uv sync --extra fine-mapping
+# ABF (no --ld) needs no extra
 python skills/fine-mapping/fine_mapping.py \
   --sumstats locus.tsv --output <report_dir>
-python skills/fine-mapping/fine_mapping.py \
+uv run --extra fine-mapping python skills/fine-mapping/fine_mapping.py \
   --sumstats locus.tsv --ld ld_matrix.npy --output <report_dir>
-python skills/fine-mapping/fine_mapping.py \
+uv run --extra fine-mapping python skills/fine-mapping/fine_mapping.py \
   --sumstats gwas_full.tsv --chr 1 --start 109000000 --end 110000000 \
   --ld ld_matrix.npy --output <report_dir>
-python skills/fine-mapping/fine_mapping.py --demo --output /tmp/finemapping_demo
+uv run --extra fine-mapping python skills/fine-mapping/fine_mapping.py --demo --output /tmp/finemapping_demo
 
 # CellposeSAM — cell segmentation from fluorescence microscopy images
 # cpsam is channel-order invariant; pass greyscale or up to 3 channels directly
@@ -526,8 +528,8 @@ python skills/recombinator/recombinator.py --demo
 # Labstep demo
 python skills/labstep/labstep.py --demo --output /tmp/labstep
 
-# SuSiE fine-mapping demo
-python skills/fine-mapping/fine_mapping.py --demo --output /tmp/finemapping_demo
+# SuSiE fine-mapping demo (needs the sushie engine: uv sync --extra fine-mapping)
+uv run --extra fine-mapping python skills/fine-mapping/fine_mapping.py --demo --output /tmp/finemapping_demo
 
 # CellposeSAM demo
 python skills/cell-detection/cell_detection.py --demo --output /tmp/cell_detection_demo
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,6 +65,13 @@ just-prs = [
     "fastmcp>=3.4.4,<4",
     "typer>=0.27,<1",
 ]
+# SuShiE fine-mapping engine (jax-based) for skills/fine-mapping.
+# glimix-core floor keeps liknorm at >=2.0 (wheels); 1.x is sdist-only and
+# fails to build on macOS.
+fine-mapping = [
+    "sushie>=0.20,<0.21",
+    "glimix-core>=3.1.14,<4",
+]
 
 [project.scripts]
 clawbio = "clawbio.cli:main"
--- a/skills/catalog.json
+++ b/skills/catalog.json
@@ -1584,7 +1584,7 @@
       "name": "fine-mapping",
       "cli_alias": null,
       "description": "Statistical fine-mapping of GWAS loci using SuSiE, SuSiE-inf, and Approximate Bayes Factors to identify credible sets and posterior inclusion probabilities (PIPs) for causal variant discovery. SuSiE-inf adds an infinitesimal polygenic component for improved calibration at well-powered loci.",
-      "version": "0.2.0",
+      "version": "0.3.0",
       "status": "planned",
       "maturity_tier": "tested",
       "maturity_evidence": {
--- a/skills/fine-mapping/SKILL.md
+++ b/skills/fine-mapping/SKILL.md
@@ -5,7 +5,7 @@ description: Statistical fine-mapping of GWAS loci using SuSiE, SuSiE-inf, and A
   component for improved calibration at well-powered loci.
 license: MIT
 metadata:
-  version: 0.2.0
+  version: 0.3.0
   author: ClawBio
   tags:
   - gwas
@@ -34,6 +34,8 @@ metadata:
       package: pandas
     - kind: pip
       package: matplotlib
+    - kind: pip
+      package: sushie
     trigger_keywords:
     - fine-mapping
     - finemapping
@@ -64,12 +66,12 @@ GWAS identifies associated loci, not causal variants. A single GWAS signal can c
 
 - **Without it**: Researchers must manually triage 10–200 correlated SNPs per locus with no principled prioritisation
 - **With it**: A ranked credible set with PIPs and 95% credible set boundaries in seconds
-- **Why ClawBio**: Runs locally without uploading individual-level data; implements ABF natively and wraps SuSiE (via polyfun) when available — no R dependency required
+- **Why ClawBio**: Runs locally without uploading individual-level data; implements ABF natively and delegates SuSiE to the published [sushie](https://github.com/mancusolab/sushie) package — no R dependency required
 
 ## Core Capabilities
 
 1. **Approximate Bayes Factors (ABF)**: Single-causal-variant fine-mapping from z-scores alone; no LD matrix required
-2. **SuSiE (Sum of Single Effects)**: Multi-signal fine-mapping with LD using the iterative Bayesian stepwise selection algorithm; pure-Python implementation, no R dependency
+2. **SuSiE (Sum of Single Effects)**: Multi-signal fine-mapping with LD, delegated to the sushie package (mancusolab/sushie, JAX-based) via its summary-statistics interface run with a single ancestry; requires the `fine-mapping` extra (`uv sync --extra fine-mapping`)
 3. **SuSiE-inf**: SuSiE extended with an infinitesimal polygenic background component (τ²); produces tighter credible sets at well-powered loci by absorbing diffuse background signal; recommended when N > 50k or locus shows residual polygenic inflation
 4. **Swappable benchmark**: `tests/benchmark/finemapping_benchmark.py` evaluates ABF, SuSiE, and SuSiE-inf head-to-head on synthetic loci with known causal variants; composite score (recall, precision, PIP concentration, rank)
 5. **Credible sets**: 95% and 99% credible sets computed from PIPs; reports size, coverage, and lead variant
@@ -99,38 +101,39 @@ When the user asks for fine-mapping:
 ## CLI Reference
 
 ```bash
-# ABF single-signal fine-mapping (no LD needed)
+# ABF single-signal fine-mapping (no LD needed; no extra required)
 python skills/fine-mapping/fine_mapping.py \
   --sumstats locus.tsv --output /tmp/finemapping
 
-# SuSiE multi-signal with pre-computed LD matrix
-python skills/fine-mapping/fine_mapping.py \
+# SuSiE multi-signal with pre-computed LD matrix (sushie engine:
+# install once with `uv sync --extra fine-mapping`)
+uv run --extra fine-mapping python skills/fine-mapping/fine_mapping.py \
   --sumstats locus.tsv --ld ld_matrix.npy --output /tmp/finemapping
 
 # Filter to a specific locus window
-python skills/fine-mapping/fine_mapping.py \
+uv run --extra fine-mapping python skills/fine-mapping/fine_mapping.py \
   --sumstats gwas_full.tsv --chr 1 --start 109000000 --end 110000000 \
   --ld ld_matrix.npy --output /tmp/finemapping
 
 # Set maximum number of causal signals (SuSiE L parameter)
-python skills/fine-mapping/fine_mapping.py \
+uv run --extra fine-mapping python skills/fine-mapping/fine_mapping.py \
   --sumstats locus.tsv --ld ld_matrix.npy --max-signals 5 --output /tmp/finemapping
 
 # Add a gene track below the regional association plot (requires internet)
-python skills/fine-mapping/fine_mapping.py \
+uv run --extra fine-mapping python skills/fine-mapping/fine_mapping.py \
   --sumstats locus.tsv --ld ld_matrix.npy --gene-track --output /tmp/finemapping
 
 # Demo mode (synthetic 200-variant locus, two causal signals)
-python skills/fine-mapping/fine_mapping.py --demo --output /tmp/finemapping_demo
+uv run --extra fine-mapping python skills/fine-mapping/fine_mapping.py --demo --output /tmp/finemapping_demo
 ```
 
 ## Demo
 
 ```bash
-python skills/fine-mapping/fine_mapping.py --demo --output /tmp/finemapping_demo
+uv run --extra fine-mapping python skills/fine-mapping/fine_mapping.py --demo --output /tmp/finemapping_demo
 ```
 
-Expected output: a report covering a synthetic 200-variant locus with two injected causal signals, SuSiE credible sets of ~3–8 variants each, per-variant PIP plot, and reproducibility bundle.
+Expected output: a report covering a synthetic 200-variant locus with two injected causal signals, two single-variant SuSiE credible sets pinpointing the causal variants (indices 60 and 140), per-variant PIP plot, and reproducibility bundle.
 
 ## Algorithm / Methodology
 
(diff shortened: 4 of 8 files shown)


manuelcorpas:
Closing and reopening to trigger CI: GitHub created no workflow run for this head (zero check runs, none parked). Nothing to do on your side.

manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Thank you for this. Moving SuSiE onto a maintained implementation is the right call. The adapter is careful. It rejects NaN z-scores before sushie can return degenerate PIPs (`susie.py:391`). It forwards `coverage` and `min_purity` so fit-time pruning and `credible_sets.py` agree. It keeps only the retained signals, so a null locus gives zero alpha rows and no phantom PIPs. Non-convergence now warns and sets `converged: False`, which matches `susieR` and still appears in the report's parameter table. The benchmark call in `tests/benchmark/finemapping_benchmark.py:177` still matches the kept keyword set. The Gotchas section is useful.

Static scan: the CI failure warning is accurate for this commit and needs a look. The four SKILL.md missing-section notes predate this PR.

### Non-blocking suggestions
These are for later.

1. `susie.py:554` maps `CSIndex` k to alpha row k-1. No test pins that mapping. `test_two_signal_demo_keeps_two_alpha_rows` only checks the shape. An assertion that the demo's credible-set leads, built from `alpha`, are indices 60 and 140 would catch an ordering change in a future sushie release.
2. With `converged: False`, `report.md` still presents PIPs as results. A one-line provisional banner in `report.py` would follow your own Gotcha 4.
3. `fine_mapping.py:179` still accepts `coverage=1.0` and `min_purity=0`, and these then fail inside `run_susie`. Tightening the CLI check for the SuSiE path would give a clearer error earlier.

---

manuelcorpas (commented):

**Verdict: NEEDS REVIEW.** The engine swap is the right direction and the adapter's mapping onto sushie 0.20 checks out against the sdist on every parameter I could trace (`L`, `effect_var`, `max_iter`, `min_tol`, `threshold`, `purity`, `min_snps`; single-ancestry `rho` ignored, `resid_var` defaults to ones, z rescaled by sqrt(n)). It cannot merge as it stands, for four reasons.

1. **The branch conflicts with main** on `pyproject.toml` and `uv.lock`, which is why GitHub has never created a CI run for it (a pull_request workflow needs a merge ref; closing and reopening did not help). Rebase first.
2. **The bench gate will fail.** `sushie` is an optional extra, but the `scientific-correctness` job runs `uv sync --locked` without `--all-extras` (`ci.yml:380`) and then the bench (`:394`). `run_susie` now raises `ImportError` at `susie.py:82-88` when sushie is absent, so every SuSiE case becomes `harness_error`, and `scripts/check_bench_baseline.py:87-91` fails on any rise from the recorded 1. Same gap in `bench-leaderboard.yml:33`. Add `--all-extras` to both jobs or make the extra a hard dependency, and re-cut `bench_baseline.json` in this PR, since your own 18/20 vs 19/20 exceeds `TOLERANCE = 0.05`.
3. **Silent project-wide `cryptography` downgrade** 48.0.0 to 45.0.7 (uv.lock hunk around patch line 1084). `liknorm 2.0.0` requires `cffi<2.0`, `cryptography 48` requires `cffi>=2.0`, so the resolver dropped cryptography three majors for every user of the lock, extras or not. `glimix-core`/`liknorm` are only needed by sushie's individual-level path, not `infer_ss`; either isolate them or justify the downgrade in the PR body.
4. **No numerical known-answer test.** The old engine's PIPs and credible sets are never pinned; the new tests assert shapes, `argmax` and `> 0.8`. `test_elbo_tracked` loosened from `>= 2` to `>= 1`; `test_alpha_rows_sum_to_at_most_one` and `test_null_locus_without_null_weight_has_phantom` were deleted; `test_two_signals_recovered` still passes when one of two causal variants is missed. Please add a fixture with recorded PIPs and CS membership from this commit so the next sushie bump cannot drift silently.

Smaller: loci with fewer variants than L now raise (`min_snps=min(p, 100)` against sushie's `if min_snps < L`), so the CLI default `--max-signals 10` errors on a 9-variant locus where IBSS ran; no test, no SKILL.md gotcha. `jax.config.update("jax_enable_x64", True)` at call time (`susie.py:93`) flips precision for every other JAX user in the process. SuSiE-inf keeps the hand-rolled IBSS with `null_weight`, so two engines with different priors now coexist under one skill while the title says the hand-rolled engine is gone; SKILL.md still describes purity as mean pairwise r² where sushie prunes on min |r|. The 0.80 to 0.87 composite in the description is a model change (EM-estimated effect variance replacing a fixed Wakefield prior), not a bug fix.

Licence is MIT, no fabricated-data tells, no network fetch added, NaN and shape guards ahead of sushie are welcome.

camlloyd:
Thanks — all four blockers are addressed, plus the smaller points and the three audit suggestions. Pushed as `e2c28a8`; the PR body has the full breakdown and the title no longer claims the hand-rolled engine is gone.

**1. Conflicts.** Merged `main`; `uv.lock` re-cut with `uv lock` rather than hand-resolved. GitHub now reports the PR `MERGEABLE`, so there is a merge ref for CI.

**2. Bench gate.** Both jobs now `uv sync --locked --all-extras`. `bench_baseline.json` re-cut for `clawbio-finemapping`, 95.0 → 90.0, measured against this head (clawbio-bench 0.1.1: 18/20, 1 harness error) — `check_bench_baseline.py` reports delta 0.0. The file asks for a lowered number to be justified rather than quietly updated, so its `_comment` records both moved cases and what restores 95.0. Filing `fm_12` and `fm_14` upstream is left with you.

**3. cryptography.** You were right to push on this, and it turned out not to be justifiable — `pip-audit` reports advisories against 45.0.7 that 48.0.0 does not carry (PYSEC-2026-2141, PYSEC-2026-35/36, GHSA-537c-gmf6-5ccf). So it is isolated, not justified: `glimix_core` is imported by `infer_ss` but `liknorm` never is (glimix-core only loads it for the GLMM/expfam models), so liknorm is dropped from the resolution via a `[tool.uv] override-dependencies` entry with an unsatisfiable marker. cryptography stays at 48.0.0, cffi at 2.0.0. The explicit `glimix-core` pin only existed to force `liknorm>=2.0` and is gone.

**4. Known-answer test.** `TestSuSiEKnownAnswer` plus a recorded `tests/fixtures/demo_susie_known_answer.json`: engine version, convergence, iteration count, credible-set count and per-row leads, PIP magnitudes by index, credible-set membership and purity, and the lead `alpha`/`mu`/`mu2`. `test_two_signals_recovered` now requires both causal variants rather than either.

**Smaller.** `L` is clamped to the variant count with a warning, so a 9-variant locus runs under the default `--max-signals 10` (test + gotcha added). The x64 flag is restored in a `finally` and pinned by a test. SKILL.md now says purity is the **minimum** `|r|`, and states plainly that two engines coexist — only the `--ld` path is sushie, SuSiE-inf is still the numpy IBSS with a different prior. The 0.80 → 0.87 composite is a model change, not a bug fix; it is out of the summary.

**Audit suggestions.** All three done: the `CSIndex k → alpha row k-1` mapping is pinned by an assertion on the demo leads (60/140); `report.md` carries a provisional banner above the credible-set tables when `converged` is false; and `run_finemapping` rejects out-of-range `coverage`/`min_purity` on the SuSiE path before loading any data.

**One thing you did not flag.** `plt.cm.get_cmap` was removed in matplotlib 3.9, so `--demo` crashed while writing figures. It is pre-existing on `main`, not caused by this branch, but it made the demo this PR asks reviewers to run fail, so I fixed the three call sites here. Happy to split it out if you would rather keep the diff clean.

`418 passed` across `skills/fine-mapping/tests/` and `tests/`; demo runs end to end with 2 single-variant credible sets on indices 60/140.

camlloyd:
Follow-up: **CI is fully green on `1f6fbad3`** — 13 checks pass, 0 open code-scanning alerts, `mergeable=MERGEABLE`, `state=CLEAN`. This is the first run GitHub has produced for this PR at all, since there was no merge ref before the rebase.

The bench gate confirms in CI what I measured locally:

```
| Harness             | Baseline | This run | Delta |
| clawbio-finemapping | 90.0%    | 90.0%    | 0.0   |
No regression against bench_baseline.json.
```

Two commits landed after my previous comment, both self-inflicted and worth flagging rather than burying:

1. **`84ff9e03` — the `pip-audit` job.** My liknorm override broke it. `pip-audit -r` shells out to pip, pip does not honour `[tool.uv] override-dependencies`, so it re-resolved glimix-core's metadata, put liknorm back, and died on `ResolutionImpossible` against the locked cffi 2.0.0. There is no version of that requirement pip can satisfy: liknorm 2.0.0 pins `cffi<2.0`, and liknorm 1.2.10 leaves cffi unconstrained but its sdist cannot generate metadata at all (its `pyproject.toml` declares no version). So the job now audits the venv `uv sync --locked --all-extras` builds instead of an exported requirements file. That is narrower in platform coverage than the old lockfile export — worth a second opinion if you would rather solve it another way — but it has no resolution step and covers what we actually ship. It reports 53 findings across 6 packages, cryptography at 48.0.0, no liknorm. `uv audit` (the `dependency-audit` job) was unaffected throughout, since uv honours the override.

2. **`1f6fbad3` — a new CodeQL alert I introduced.** `py/redundant-comparison` on the SuSiE-path coverage guard from the round-2 work: `coverage <= 0` already raises a few lines above, so the `0 < coverage` half of `not 0 < coverage < 1` could never fire. Now `coverage >= 1`. The `min_purity` guard beside it is genuinely two-sided (the shared check above permits `0`) and is unchanged. The original alert 778 (the duplicate import) is cleared.

Nothing else changed. Ready for your read whenever you have time.

manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Thank you for this. Moving SuSiE onto sushie in `fine_mapping_core/susie.py` is a real improvement, and the adapter is careful. You check for NaN z-scores before sushie sees them, you pass `coverage` and `min_purity` straight through to sushie's pruning so fitting and set building use the same thresholds, and you restore the jax x64 flag in a `finally`. Pruned signals are dropped rather than kept at near-uniform values, which removes the phantom credible sets on the demo locus. The known-answer fixture that pins PIPs, alpha row order and moment scale is exactly the right guard against engine drift. Showing a provisional banner in `report.py` when the fit did not converge, instead of raising, is a sound choice, and it is tested.

The static scan reported no findings for this PR, so there are no false positives to record.

1. `fine_mapping.py:385`: the `--min-purity` help text still says "Min average pairwise |r|". The code and SKILL.md now use the minimum |r| (Wang 2020 §3.2), so the help should say the same.
2. `bench_baseline.json`: the clawbio-finemapping baseline drops from 95.0 to 90.0. Your note on fm_12 and fm_14 explains it well. It is worth linking the upstream issues so the 95.0 can be restored once those cases are regraded.
3. `tests/benchmark/finemapping_benchmark.py:177` still passes `max_iter=100`, while the new default is 500. On harder loci the head-to-head benchmark may now report SuSiE as not converged when the CLI would converge.

---
Approved and merged.

PR #424, opened 2026-09-11
ci: widen the Dependabot cooldown to 7 days

camlloyd:
## Summary

- Sets an explicit 7-day `cooldown` on both the `uv` and `github-actions` Dependabot ecosystems in `.github/dependabot.yml`.
- Dependabot already waits 3 days by default before proposing a version update. A compromised release is usually pulled within days of publication, and 3 sits at the edge of that window, so this widens it to a week. Security updates bypass cooldown either way.
- Resolves the two `dependabot-cooldown` findings zizmor reports against `main`.

## Skill affected

None. CI configuration only (`.github/dependabot.yml`).

## Checklist

- [x] SKILL.md is present and complete (YAML frontmatter + methodology) — n/a, no skill touched
- [x] Tests included and passing (`pytest -v`) — n/a, no Python changed; verified with zizmor below

## Test output

```
$ uvx zizmor --no-progress --format plain .github/dependabot.yml
No findings to report. Good job!
```

Before this change, zizmor 1.30.1 reported on `main`:

```
warning[dependabot-cooldown]: insufficient cooldown in Dependabot updates
 --> .github/dependabot.yml:7:5   (package-ecosystem: "uv")
warning[dependabot-cooldown]: insufficient cooldown in Dependabot updates
 --> .github/dependabot.yml:16:5  (package-ecosystem: "github-actions")
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,12 @@ updates:
     schedule:
       interval: "weekly"
       day: "monday"
+    # Dependabot already waits 3 days by default before proposing a version
+    # update. Widen that to a week: a compromised release is usually pulled
+    # within days of publication, and 3 sits at the edge of that window.
+    # Security updates bypass cooldown either way (zizmor: dependabot-cooldown).
+    cooldown:
+      default-days: 7
     open-pull-requests-limit: 5
     groups:
       python-minor-and-patch:
@@ -18,4 +24,10 @@ updates:
     schedule:
       interval: "weekly"
       day: "monday"
+    # Dependabot already waits 3 days by default before proposing a version
+    # update. Widen that to a week: a compromised release is usually pulled
+    # within days of publication, and 3 sits at the edge of that window.
+    # Security updates bypass cooldown either way (zizmor: dependabot-cooldown).
+    cooldown:
+      default-days: 7
     open-pull-requests-limit: 5


manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Thanks for this. Config only, and it does what the title says.

What I checked:

- No skill code is touched. Nothing here affects variant classification, ClinVar or VEP vocabulary, population frequencies, effect sizes or any analysis output, so there is no route from this change to a wrong scientific result.
- `cooldown` with `default-days` is applied consistently to both ecosystem entries in the file (`.github/dependabot.yml:16` and `.github/dependabot.yml:31`), not just the first one. That is the easy thing to get half right, and you got it right.
- Security updates are not gated by cooldown, so the vulnerability response window is unchanged. Worth stating plainly, since that is the main worry a reviewer would have about slowing updates down.

The static scan reported no findings on this PR and I found nothing it missed.

**Non-blocking suggestions** (for later, nothing to fix now)

1. The comment at `.github/dependabot.yml:12` and `.github/dependabot.yml:27` says Dependabot "already waits 3 days by default". I could not confirm a non-zero default cooldown from anything in the diff. If the real default is no delay, this PR introduces a cooldown rather than widening one, and the comment would be worth rewording so future readers are not misled about the baseline.
2. The two comment blocks are duplicated verbatim. A single note at the top of `updates:` would drift less as the file grows.

---
Approved and merged.

PR #425, opened 2026-09-12
dnasp: internal stop codons analysed as a 21st amino acid, as in DnaSP

drdaviddelorenzo:
## Summary

Follow-up to #408 (merged as `a4a9579`). One commit, `d4a90a2`: internal stop
codons follow DnaSP's rule in the `dnasp` skill's McDonald-Kreitman test and
Ka/Ks. It was pushed to the same branch minutes after #408 was merged, so it
needs its own PR.

DnaSP's help file: "if DnaSP finds stop codons (in the middle of coding
regions) they will be considered as if they would code for a new amino acid
(the amino acid 21; for example Selenocysteine)", while pathways through stop
codons are excluded. In the VB, `PosicionesSynonimas` and `NumSynonEntreCodons`
compare `GCode` strings (`* = *` synonymous, `*` vs a sense codon a
replacement) and `GeneticCodeFold` gives a stop codon zero synonymous and three
nonsynonymous sites; `ComparaSegregatingCodonesMK` sets `STOPinMiddle` and the
output warns. The skill used to skip any codon containing a stop.

- `compute_mk` and `compute_ka_ks` no longer skip stop codons;
  `_classify_codon_pair` compares stop endpoints as an amino acid and still
  drops orderings through a stop; the MK fixed side now labels pairs with the
  same ported path scorer as the within-species side (`_mk_pathway` removed).
- A final codon that is a stop in every sequence is the terminal stop and is
  left out (`_terminal_stop_codon`): DnaSP's own `COII_Apes.nex` annotates
  positions 682-684 as noncoding (`CODONPOSSET N: 682-684`) and codes 1-681,
  hence its 227 analysed codons.
- `MKStats` and `KaKsStats` gain `n_internal_stop_codons`; the report shows it
  and the CLI prints DnaSP's warning when it is non-zero, since an internal
  stop usually means a frame or genetic-code problem.

## Skill affected

`dnasp`

## Validation

- COII under the vertebrate mitochondrial code (the right table for the gene;
  no internal stops) is unchanged: MK Pn 10, Ps 69, Dn 8, Ds 56; Ka/Ks S
  168.222, Ks 0.83011, Ka 0.02561; 227 codons analysed.
- COII under the standard code, which turns four TGA/TGG codons into stops,
  now reproduces DnaSP 6.12 as well: MK Pn 13, Ps 66, Dn 9, Ds 55, Fisher's P
  0.816904, all exact; Ka/Ks S 177.278 and N 503.722 exact, raw Ks 0.46511 vs
  DnaSP's 0.46512, raw Ka 0.02960 exact, JC omega 0.0416 vs 0.042. (DnaSP's
  standard-code Ks/Ka in the validation table were its uncorrected values; its
  panel prints raw and JC side by side.)
- Recorded as note 18 of `validation_table.md` in the DnaSP-ClawBio validation
  bundle; `SKILL.md` and `docs/index.md` describe the rule.

## Checklist

- [x] SKILL.md updated (mk and kaks paragraphs, the mk gotcha about TGA)
- [x] Tests included and passing  -  `440 -> 444`, `pytest -v` green; the three
      tests that encoded the old skip are rewritten with the VB and help-file
      justification in their comments
- [x] Demo data unchanged (`--demo`)
- [x] No patient/sensitive data
- [x] Follows CONTRIBUTING.md  -  spaced hyphens, no em/en dashes

## Test output

```
$ cd skills/dnasp && pytest -q
444 passed in ~2s
```


manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Thank you for this change. DnaSP's rule is now applied the same way everywhere. A stop inside the coding region counts as a 21st amino acid. It is synonymous with another stop and a replacement against a sense codon, and a path through a stop as an intermediate codon is still dropped. I checked the tests by hand. TAA/AAG gives 1 synonymous and 1 replacement change, and TAA/ATG gives 0.5 and 2.5 over the four surviving orderings. Both are right. It is also good that the retired `_mk_pathway` is gone, so the MK and Ka/Ks paths share one scorer. Leaving out the terminal stop only when every sequence has a stop there is a sound guard. Counting internal stops, rather than silently dropping those codons, makes a wrong frame or genetic code easier to see.

The static scan reported no findings, so there were no false positives to check.

### Non-blocking suggestions
These are for later:

1. `dnasp.py:4961` prints a warning for MK only. The Ka/Ks summary at `dnasp.py:4966` stays silent even when `kaks.n_internal_stop_codons` is non-zero. The report table does show the count, but a matching CLI warning would help.
2. In Ka/Ks, a stop codon adds zero synonymous sites (`dnasp.py:2929-2933`), yet a TAA/TAG pair still adds one synonymous difference (`dnasp.py:2287-2290`). If that matches DnaSP, a one-line note in `docs/index.md` would stop someone reading it as a bug.
3. The "Group D: Fu's Fs and Site Frequency Spectrum" section header was deleted near `tests/test_dnasp.py:2686`. That looks accidental.

---
Approved and merged.

PR #427, opened 2026-09-12
fix(nutrigx): stream the AncestryDNA parser instead of loading the file into memory

drdaviddelorenzo:
## What

`parse_ancestry` read every line of the input into a list before parsing, so memory use scaled with the size of the genetic file. Consumer exports run to hundreds of thousands of rows, which makes this a denial-of-service risk on untrusted input and simply wasteful on trusted input.

`csv.DictReader` accepts any iterable of strings, so comment lines can be filtered by a generator and the file streamed.

`detect_format`, `parse_23andme` and `parse_vcf` already stream — this brings `parse_ancestry` into line with them.

## Dead code removed in the same function

```python
with open(filepath, encoding="utf-8", errors="replace") as f:
    reader = csv.DictReader(f, delimiter="\t")
    raw = f.read() if not hasattr(reader, 'fieldnames') else ""
```

`csv.DictReader` always has a `fieldnames` attribute, so this branch never ran, `raw` was always `""` and never used, and the file was opened a second time to do the real work.

## Testing

Output is unchanged: comment lines skipped, header respected, only `rs`-prefixed identifiers collected. 24 nutrigx tests pass.

## Context

Found via a security audit of [drdaviddelorenzo/nutrigenomics](https://github.com/drdaviddelorenzo/nutrigenomics), which shares this parser's ancestry. The same fix is applied there.

### DIFF
--- a/skills/nutrigx/parse_input.py
+++ b/skills/nutrigx/parse_input.py
@@ -53,27 +53,23 @@ def parse_23andme(filepath: str) -> dict:
 
 
 def parse_ancestry(filepath: str) -> dict:
-    """Parse AncestryDNA raw data file. Returns {rsid: genotype}."""
+    """
+    Parse an AncestryDNA raw data file. Returns {rsid: genotype}.
+
+    The file is streamed. csv.DictReader accepts any iterable of strings, so
+    comment lines are filtered by a generator rather than by first building a
+    list of every line in the file: consumer genetic files are large, and
+    materialising one made memory use scale with input size.
+    """
     genotypes = {}
     with open(filepath, encoding="utf-8", errors="replace") as f:
-        reader = csv.DictReader(f, delimiter="\t")
-        # Handle comment lines
-        raw = f.read() if not hasattr(reader, 'fieldnames') else ""
-    
-    # Re-read skipping comments
-    lines = []
-    with open(filepath, encoding="utf-8", errors="replace") as f:
-        for line in f:
-            if not line.startswith("#"):
-                lines.append(line)
-    
-    reader = csv.DictReader(lines, delimiter="\t")
-    for row in reader:
-        rsid = row.get("rsid", "").strip()
-        allele1 = row.get("allele1", "").strip()
-        allele2 = row.get("allele2", "").strip()
-        if rsid.startswith("rs"):
-            genotypes[rsid] = allele1 + allele2
+        rows = (line for line in f if not line.startswith("#"))
+        for row in csv.DictReader(rows, delimiter="\t"):
+            rsid = row.get("rsid", "").strip()
+            allele1 = row.get("allele1", "").strip()
+            allele2 = row.get("allele2", "").strip()
+            if rsid.startswith("rs"):
+                genotypes[rsid] = allele1 + allele2
     return genotypes
 
 


manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Thanks for this, a tidy cleanup. I compared the new `parse_ancestry` against the version on main. Its output is the same. On main the first `with` block (`parse_input.py:58-61`) did nothing useful: `hasattr(reader, 'fieldnames')` is always true, so `raw` was never used. The file was then read a second time into a list. Your generator gives `csv.DictReader` the same non-comment lines in the same order. The `rsid`, `allele1` and `allele2` keys still match the AncestryDNA header, and the rows are filtered and joined in the same way. No genotype calls change, nothing new leaves the machine and there is no clinical logic in the diff. The docstring explains the reason for the change clearly.

The static scan raised one warning: CI failed on this commit (`test (3.13)`; 3.11 and 3.12 were cancelled). That is true, but from the diff alone I cannot link the failure to this change, because the parser's behaviour is unchanged. It is worth checking the job log to confirm the failure comes from elsewhere.

### Non-blocking suggestions
These are for later.

1. None of the nutrigx tests use the AncestryDNA path. Every call in `tests/test_nutrigx.py` passes `fmt="23andme"`. A small test that parses an AncestryDNA fixture with a `#` comment header would protect this refactor. You could reuse `clawbio/tests/fixtures/mock_ancestry.txt`.
2. This problem was already there before your change. If a row has fewer columns than the header, `DictReader` fills the missing fields with `None`, so `row.get("allele2", "")` returns `None` and `.strip()` raises (`parse_input.py:69`). Passing `restval=""` would handle truncated lines safely.

---

manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Thank you for this. It is a tidy fix. The old `parse_ancestry` opened the file twice and kept every non-comment line in a list. The first open did nothing, because `hasattr(reader, 'fieldnames')` is always true. Your version reads the file once, line by line, with a generator. For a well-formed AncestryDNA file it gives exactly the same `{rsid: allele1+allele2}` result as before, so no call changes.

The `restval=""` change is a real fix too. On main, a row missing `allele2` gives `None`, and `.strip()` then raises an error. With your change, that row gives a one-letter genotype. `extract_genotypes.py:57` already treats anything shorter than two letters as `no_call`, so a cut-off row can never be scored as a risk allele. I found nothing in this diff that could give a wrong answer. It sends no data anywhere and cites no sources.

The static scanner reported no findings, so there were no claims to check.

**Non-blocking suggestions** (for later, not needed for this PR)

1. `tests/test_ancestry_parser.py:97` checks that the parser returns `"A"`. The result that matters for safety is the next step. A second assertion that `extract_snp_genotypes` marks that SNP as `no_call` would guard that path.
2. AncestryDNA writes `0` for a failed allele, so a no-call comes through as `"00"`. Downstream this becomes `allele_mismatch` with a warning rather than `no_call` (`extract_genotypes.py:125`). It is never scored, and the behaviour predates this PR. Dropping `0` alleles in `parse_input.py:45` would label it correctly.
3. `parse_input.py:3` still describes AncestryDNA files as `.csv`. The parser reads them as tab-separated.

---
Approved. Merge deferred to the maintainer.

manuelcorpas (approved):
Approved after maintainer read on head 53092a2665d82e6ee5c17a72fbddd3a9bcdbafb4: the daily audit's central claim verified against main, both follow-up suggestions addressed, CI green on this exact head. Automated audit by ClawBio pr-audit plus a maintainer read.

PR #428, opened 2026-09-12
fix(nutrigx): write output without following symlinks, keep paths out of artefacts, sanitise the input filename

drdaviddelorenzo:
## What

Three hardening changes to the output path, all scoped so they do **not** change *where* output is allowed to go — `api.py` accepts an arbitrary `output_dir` from its caller, and that should keep working.

### 1. Symlink-safe writes

Output filenames are deterministic (`nutrigx_report.md`, `nutrigx_radar.png`, `nutrigx_heatmap.png`), so anyone able to write into the output directory before a run can pre-create one of those names as a symlink and have a genetic report written through it, to a location the user never chose. `Path.write_text` and matplotlib's `savefig` both follow symlinks.

A new `path_safety.py` guards **both** components:

- the final component with `O_NOFOLLOW`, so the file itself cannot be a symlink
- the parent directory opened with `O_DIRECTORY | O_NOFOLLOW` and used as a `dir_fd`, so the write is anchored to a real directory inode and swapping the parent afterwards cannot move it

`O_EXCL` is deliberately **not** used — overwriting a regular file on a re-run is expected. Verified against both attack shapes; neither reaches the target.

These helpers are policy-free: they decide only that a path is not a symlink, never where output may live. **They may belong in `clawbio.common` rather than in the skill — happy to move them if you prefer.**

### 2. Paths kept out of artefacts

`provenance.json` deliberately records only the input *basename*, but `"args": args` wrote the full argument namespace beside it, putting the absolute path of the user's genetic data file into the artefact anyway. The same namespace was expanded into the reproduction command in `commands.sh`.

Both now use an explicit allowlist (`format`, `no_figures`, and a `custom_panel` boolean), and `commands.sh` shows a placeholder:

```
python nutrigx.py --input <your_genetic_file> --format 23andme
```

### 3. Input filename sanitised in the report

The report header writes the filename into a Markdown code span. A backtick in the name closes that span and lets arbitrary Markdown or HTML into a document that reads as a generated clinical report. The name is attacker-influenced wherever the file did not come from the person reading the report — a shared drive, an upload endpoint, or `api.py` driven by an orchestrator.

Now filtered to a conservative allowlist, truncated to 80 characters, placeholder when absent. Ordinary names like `AncestryDNA (2026).txt` are unchanged.

## Deliberately not included

- **Workspace confinement and an input extension allowlist.** The skill currently applies no path validation at all, but `api.py` takes an arbitrary `output_dir` and `panel_path` from its caller, so a "must live under the working directory" policy could break legitimate programmatic use. That is a design decision for you rather than something to slip into a hardening PR.
- **`input_sha256`, and the retained input basename in `provenance.json`.** Both are re-identification vectors for a genetic file, but they read as considered decisions — the comment says inputs are attested in provenance rather than in the manifest. Raised in a separate issue instead.

## Testing

24 nutrigx tests pass.

## Context

From a security audit of [drdaviddelorenzo/nutrigenomics](https://github.com/drdaviddelorenzo/nutrigenomics), which shares this code's ancestry; the equivalent fixes are applied there.

### DIFF
--- a/skills/nutrigx/generate_report.py
+++ b/skills/nutrigx/generate_report.py
@@ -7,6 +7,8 @@ import json
 from datetime import datetime, timezone
 from pathlib import Path
 
+from path_safety import safe_open_write, safe_write_text
+
 
 DOMAIN_LABELS = {
     "folate": "Folate / B-Vitamins",
@@ -228,7 +230,7 @@ def generate_report(snp_calls, risk_scores, snp_panel, output_dir, figures=True,
 
     report_text = "\n".join(lines)
     report_path = output_dir / "nutrigx_report.md"
-    report_path.write_text(report_text)
+    safe_write_text(report_path, report_text)
 
     if figures:
         _generate_figures(risk_scores, output_dir)
@@ -276,7 +278,8 @@ def _generate_figures(risk_scores: dict, output_dir: Path):
 
     plt.title("NutriGx Nutrient Risk Profile", size=14, fontweight="bold", pad=20)
     plt.tight_layout()
-    fig.savefig(output_dir / "nutrigx_radar.png", dpi=150, bbox_inches="tight")
+    with safe_open_write(output_dir / "nutrigx_radar.png", binary=True) as _fh:
+        fig.savefig(_fh, format="png", dpi=150, bbox_inches="tight")
     plt.close()
 
     # ── Heatmap ───────────────────────────────────────────────────────────────
@@ -304,7 +307,8 @@ def _generate_figures(risk_scores: dict, output_dir: Path):
                         cbar_kws={"label": "Risk Score (0=Ref, 0.5=Het, 1=Hom Risk)"})
             ax.set_title("Gene × Nutrient Risk Heatmap", fontsize=13, fontweight="bold")
             plt.tight_layout()
-            fig.savefig(output_dir / "nutrigx_heatmap.png", dpi=150, bbox_inches="tight")
+            with safe_open_write(output_dir / "nutrigx_heatmap.png", binary=True) as _fh:
+                fig.savefig(_fh, format="png", dpi=150, bbox_inches="tight")
             plt.close()
     except ImportError:
         pass
--- a/skills/nutrigx/generate_report.py
+++ b/skills/nutrigx/generate_report.py
@@ -5,6 +5,7 @@ generate_report.py — Markdown report + matplotlib figures for NutriGx Advisor
 import os
 import json
 from datetime import datetime, timezone
+import re
 from pathlib import Path
 
 from path_safety import safe_open_write, safe_write_text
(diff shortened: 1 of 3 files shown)


manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Thanks, David. This is a careful hardening pass. Anchoring writes to a directory fd opened with `O_NOFOLLOW` (`path_safety.py:126`) closes the parent swap as well as the final component. The allowlist in `safe_display_filename` (`generate_report.py:21`) stops a backtick escaping the Input code span. Keeping full paths out of `provenance.json` and `commands.sh` matches the existing basename-only intent. No scoring or genotype logic is touched, so I see no risk of a wrong nutrigenomic result.

Both scanner findings are accurate. CI `test (3.11)` failed on this commit, and `path_safety.py` has no test. I could not see the CI log, so please check whether the failure comes from this change.

### Non-blocking suggestions
These are for later.

1. `repro_bundle.py:62`: bash reads `<your_genetic_file>` as two redirections. The line runs as `--input auto` and sends stdout to a file named `--format`. It fails loudly, but `"${INPUT_FILE:?set INPUT_FILE}"` would make the script runnable.
2. `repro_bundle.py:42`: `--panel` is no longer emitted. A custom-panel run replayed from `commands.sh` would therefore use the default panel. When `custom_panel` is true, a variable for the panel would keep that visible.
3. A small test would lock this in: a symlinked file or parent should raise `ValueError`. Note that `path_safety.py:126` also refuses an output directory the user deliberately symlinked, and it needs `O_DIRECTORY`, which Windows lacks. A line in SKILL.md would help.

---

manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Thank you for this. It is careful hardening, and it leaves scoring, the SNP panel and risk categories untouched, so the report's content cannot change.

What is right:

- `path_safety.py` guards both the final component and the parent. It opens the parent with `O_DIRECTORY | O_NOFOLLOW` and passes it as `dir_fd`, which is the correct fix. The `ELOOP`/`EMLINK` handling covers Linux, macOS and BSD.
- The test covers a symlinked file, a symlinked parent and an ordinary overwrite. It also checks that nothing escaped to the target.
- `provenance.json` and `commands.sh` now hold only `format` and `no_figures`. Replaying with a custom panel requires `PANEL_FILE`, so a replay cannot silently fall back to the default panel.
- The backtick breakout fix in `safe_display_filename` is sound.

The static scan reported no findings, so there were none to recheck.

### Non-blocking suggestions

These are for later.

1. `repro_bundle.py:72` no longer passes `--output`. A replay therefore writes to `./nutrigx_results`, while step 3 still checks the original output directory. Adding `--output "$(dirname "$0")/.."` would point the check back at the replayed files.
2. `path_safety.py:76` creates new files as `0o600`, where before they followed the umask. That is reasonable for genetic data. It is worth one line in `SKILL.md`, since people sharing an output directory may be surprised.

---
Approved and merged.

manuelcorpas (approved):
Approved after maintainer read on head 3ef3d2774172471f0639f4fa590544ae25ef14af: the daily audit's central claim verified against main, both follow-up suggestions addressed, CI green on this exact head. Automated audit by ClawBio pr-audit plus a maintainer read.

PR #430, opened 2026-09-12
fix(nutrigx): validate genotype calls before scoring or rendering them

drdaviddelorenzo:
## The problem

Genotype values are read from a user-supplied file and written into a Markdown code span in the report:

```python
f"| {s['gene']} | {s['rsid']} | `{s['genotype']}` "
```

Nothing validated them. A backtick in the genotype column closes that span and the rest renders as Markdown, so a crafted 23andMe or MyHeritage export can inject headings, links or raw HTML into a document that reads as a generated clinical report, and can add table rows that look like genuine findings.

## The fix

Only nucleotide calls are accepted: `A`, `C`, `G`, `T`, plus `D` and `I` for 23andMe deletions and insertions. A run is allowed so VCF indels still parse. Anything else is treated as a **failed call** and never scored.

This is applied on **every entry path**, not only in this skill's parsers:

| Path | How it parses | Validation |
|---|---|---|
| `nutrigx.py` CLI | `clawbio.common.parsers.parse_genetic_file` | `clean_genotype_table()` before scoring |
| `api.py` | caller-built dict | `clean_genotype_table()` before scoring |
| `parse_input.py` parsers | this skill | `clean_genotype()` per call |

`clawbio.common.parsers` itself is left unchanged, since other skills depend on it.

Rejected calls are kept as `""`, so they report as `no_call` rather than `not_tested`; a failed call stays distinguishable from an SNP that was never typed.

`generate_report` guards the value again where it is written, and no longer lets `|` through, since a pipe splits a GFM table cell even inside a code span.

**Rejecting is also the scientifically correct behaviour.** A value that is not a genotype cannot be biologically meaningful and should never contribute to a nutrient risk score.

## Verification

- 29 nutrigx tests pass, including tests for `clean_genotype`, `no_call` status on rejected calls, the CLI and API paths, and the display helper.
- A CLI run on the synthetic patient still reports 28/28 panel coverage.

## Note on overlap

This touches `parse_input.py` and `generate_report.py`, which #427 and #428 also modify. Kept separate so each can be reviewed on its own; I will resolve any conflict in whichever lands last.

## Context

Found by ClawHub's security audit of [drdaviddelorenzo/nutrigenomics](https://github.com/drdaviddelorenzo/nutrigenomics) 0.3.7, where the equivalent fix is applied. The CLI and API coverage, the `no_call` status and the `|` exclusion come from the automated review on this PR.

### DIFF
--- a/skills/nutrigx/generate_report.py
+++ b/skills/nutrigx/generate_report.py
@@ -5,6 +5,7 @@ generate_report.py — Markdown report + matplotlib figures for NutriGx Advisor
 import os
 import json
 from datetime import datetime, timezone
+import re
 from pathlib import Path
 
 
@@ -86,6 +87,22 @@ RECOMMENDATIONS = {
         "Elevated": "Multiple antioxidant enzyme variants detected. Consider assessed supplementation: selenium, CoQ10, and magnesium. Limit pro-oxidant exposures (smoking, excess alcohol).",
     },
 }
+
+
+_SAFE_GENOTYPE_CHARS = re.compile(r"[^A-Za-z0-9/|]")
+
+
+def safe_display_genotype(genotype) -> str:
+    """
+    Render a genotype call safely for Markdown.
+
+    parse_input rejects non-nucleotide calls, but generate_report is also reached
+    directly by api.py and by any caller passing a genotype dict, so the value is
+    guarded again where it is written into a code span.
+    """
+    if not genotype:
+        return "--"
+    return _SAFE_GENOTYPE_CHARS.sub("_", str(genotype))[:32]
 
 
 def generate_report(snp_calls, risk_scores, snp_panel, output_dir, figures=True, input_file=""):
@@ -181,7 +198,7 @@ def generate_report(snp_calls, risk_scores, snp_panel, output_dir, figures=True,
             for s in data["contributing_snps"]:
                 effect = s["effect_direction"].replace("_", " ").title()
                 lines.append(
-                    f"| {s['gene']} | {s['rsid']} | `{s['genotype']}` "
+                    f"| {s['gene']} | {s['rsid']} | `{safe_display_genotype(s['genotype'])}` "
                     f"| {s['risk_count']}/2 | {effect} |"
                 )
             lines.append("")
--- a/skills/nutrigx/parse_input.py
+++ b/skills/nutrigx/parse_input.py
@@ -9,6 +9,26 @@ import re
 from pathlib import Path
 
 
+# Genotype calls are read from a user-supplied file and later rendered into a
+# Markdown report, so they are untrusted text until proven otherwise. Only
+# nucleotide calls are accepted: A/C/G/T plus D and I, which 23andMe uses for
+# deletions and insertions. A run is allowed because a VCF call joins REF/ALT
+# alleles and an indel can be several bases. Anything else is discarded rather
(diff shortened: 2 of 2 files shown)


manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Thanks for this. It is a well-aimed hardening change. `clean_genotype` at `parse_input.py:19` only accepts A/C/G/T/D/I after uppercasing, which suits every risk and ref allele in `data/snp_panel.json`. Dropped calls end up with no score in `score_variants.py:67`, so no nutrient score changes for valid input. Guarding the value again where it is rendered (`generate_report.py:201`) is the right call, because `api.py` skips the parser.

One thing to check: `test (3.11)` failed on this commit and the other test jobs were cancelled. I could not see the log. The fixture `tests/synthetic_patient.csv` holds only A/C/G/T calls, so I could not tie the failure to this diff.

### Non-blocking suggestions

These are for later.

1. The CLI does not use this parser. `nutrigx.py:21` imports `parse_genetic_file` from `clawbio/common/parsers.py`, so the whitelist only covers the tests and direct imports. Rendering is still safe thanks to the report guard. The same check in `clawbio/common/parsers.py:175` would make the title's "before scoring" true for the CLI too.
2. No-calls such as `--`, Ancestry `00` or VCF `./.` are now left out of the dict. Their status therefore changes from `no_call` (`extract_genotypes.py:57`) to `not_tested` (`extract_genotypes.py:43`). Scores are the same, but the label no longer tells a failed call apart from an untyped SNP.
3. `safe_display_genotype` keeps `|` (`generate_report.py:92`), and a pipe splits a GFM table cell even inside a code span. A few unit tests for both helpers would also lock in the behaviour.

---

manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Thank you for this. It is a careful fix, and it closes every way a genotype can get into nutrigx: the skill parsers (`parse_input.py:192`), the shared-parser CLI path (`nutrigx.py:95`) and caller-built dicts in `api.py:67`.

I checked it for scoring changes. It does not change any valid call. `clean_genotype` only strips whitespace and dashes and uppercases, so a real call reaches `extract_snp_genotypes` as before. `score_variants.py:67` only scores calls with status `found`. That means a rejected call counts as missing, as `allele_mismatch` did, and no domain score moves. Two edge cases now come out more accurately. An AncestryDNA `00` no-call is reported as `no_call`, not `allele_mismatch`. A lowercase call is scored, not discarded. Mapping rejected calls to `""` rather than dropping them is right: they show as failed calls, not untested SNPs.

The static scan reported no findings for this PR, so there were none to re-check.

1. `tests/test_genotype_validation.py:185` only checks that a line of source text exists in `nutrigx.py`. A small run of `main()` on a hostile file in `tmp_path` would test what the CLI actually does, and it would survive a harmless refactor.
2. `generate_report.py:311` puts the helper at the end of the module to avoid merge conflicts. Once the other nutrigx PRs have settled, it could move up next to the other helpers.

---
Approved and merged.

manuelcorpas (approved):
Approved after maintainer read on head 1528817c431a08014efacce6debc20290abdc932: the daily audit's central claim verified against main, both follow-up suggestions addressed, CI green on this exact head. Automated audit by ClawBio pr-audit plus a maintainer read.

PR #426, opened 2026-09-12
fix(nutrigx): correct 23 panel citations and score rs953413's A allele as lower DHA

drdaviddelorenzo:
## What

Two panel corrections, both verified against the primary literature.

1. **23 of 28 citations pointed at unrelated papers.** Replaced with checked sources. No score changes.
2. **rs953413 (ELOVL2) scored the wrong allele as risk.** Fixed, with qualified omega-3 advice. **Omega-3 scores change for GG and AA carriers.**

---

## 1. Citations

Every PMID in `skills/nutrigx/data/snp_panel.json` was checked against PubMed. 23 of the 28 cited a paper unrelated to the variant or trait:

| SNP | Domain | Cited PMID | What that paper actually is |
|---|---|---|---|
| rs1801133 MTHFR | folate | 9042914 | an FGFR3 craniosynostosis syndrome |
| rs1801282 PPARG | fat | 11773551 | "do multiple vaccines overwhelm or weaken the infant's immune system?" |
| rs671 ALDH2 | alcohol | 12042820 | genes involved in *Drosophila melanogaster* geotaxis |
| rs33972313 SLC23A1 | vitamin C | 25498559 | PCB contamination in bluefin tuna |
| rs4680 COMT | antioxidant | 9463411 | Ly49 receptor expression on NK cells |
| rs1050450 GPX1 | antioxidant | 11786716 | open window thoracostomy for thoracic empyema |

Only `rs9939609` (FTO), `rs7903146` (TCF7L2), the two BCMO1 entries and `rs4988235` cited the paper they claimed. The panel's biology is not in question; its provenance was.

Where one exists, each replacement is a study with a catalogued genome-wide association for that exact variant and trait (GWAS Catalog API). Eleven entries have no such association for their domain and cite functional or candidate-gene work instead: BCMO1 `rs7501331`/`rs12934922`, VDR `rs2228570`/`rs731236`, MTR `rs1805087`, CYP1A2 `rs762551`, SOD2 `rs4880`, GPX1 `rs1050450`, NQO1 `rs1800566`, COMT `rs4680`. Judgement calls about panel membership are in #429.

The automated review asked whether the two shared PMIDs report each exact rsID. **28334899** reports both `rs662799` (APOA5) and `rs7412` (APOE). **21829377** (Lemaitre 2011) reports ELOVL2 through `rs2236212`, `rs3734398` and `rs3798713`, not `rs953413`, so `rs953413` now cites the paper that does report it (below).

---

[...]

PR #431, opened 2026-09-12
chore: correct the demo-data count in the README headline

drdaviddelorenzo:
## Why

`tests/test_public_claims.py::test_demo_and_cli_counts_match_catalog` is **failing on `main`**, so every open PR inherits a red `test (3.11 / 3.12 / 3.13)`:

```
AssertionError: ["README.md: '91 with runnable demo data' but catalog has_demo=92"]
```

`skills/catalog.json` has `has_demo=92`. The README states the figure twice — line 696 already says **92**, the headline on line 71 still says **91**. One occurrence was updated and the other missed.

## The change

One digit in the headline.

## Verification

- `tests/test_public_claims.py` — 15 passed
- full repo suite — **306 passed, 0 failed** (was 1 failed, 305 passed)

## Context

Found while opening #426, #427, #428 and #430 against nutrigx. I confirmed the failure reproduces on unmodified `upstream/main` and that none of those four branches introduces it — each shows the same single pre-existing failure and nothing else — so I have kept the fix out of them rather than bundling an unrelated README edit into a citations or parser PR.

If a catalog automation job is meant to keep this figure in step, this is presumably a case it missed rather than something to fix by hand each time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ uv run python clawbio.py run pharmgx --demo
 
 ## What ClawBio Does Today
 
-**97 skills (91 with runnable demo data) + 8,182 Galaxy tools + 4,820 tests + benchmark validation. Local-first by default. Reproducible. No guessing.**
+**97 skills (92 with runnable demo data) + 8,182 Galaxy tools + 4,820 tests + benchmark validation. Local-first by default. Reproducible. No guessing.**
 > **v0.5.0 released** (4 Apr 2026): Validation and Benchmark Infrastructure. AD ground truth benchmark, mock API server for offline testing, swappable fine-mapping pipeline (SuSiE vs ABF), 74 benchmark tests, red/green TDD mandate. [Release notes](https://github.com/ClawBio/ClawBio/releases/tag/v0.5.0). DOI: [10.5281/zenodo.19420648](https://doi.org/10.5281/zenodo.19420648).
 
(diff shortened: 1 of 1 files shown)


manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Thanks for this. It is a small, careful fix, and the figure checks out.

The headline change from 91 to 92 skills with runnable demo data (`README.md:71`) matches the source it is generated from. In `skills/catalog.json` on main, `skill_count` is 97. The `has_demo` field is true for 92 skills and false for 5, which adds up to 97. The same 92 already appears in `index.html:262` and in the later README paragraph (`README.md:671`). So your edit removes an inconsistency and does not add one. The wording "with runnable demo data" is also the correct term for that field, as explained in the CHANGELOG.

The change touches no skill code, clinical logic, data source vocabulary or citations, so it cannot change any analysis result. The static scanner reported no findings for this PR, and I agree with it.

### Non-blocking suggestions

These are for later.

1. The test count in the same headline (`README.md:71`, 4,820 tests) differs from the checkout of main I compared against. That copy of the README says 4,605 tests, and `index.html:262` says 4,623. These counts drift easily. Regenerating all the headline figures from the commands recorded in `llms.txt` in one pass would keep the README, `llms.txt` and `index.html` in step.
2. If the demo count is edited by hand, a light check that compares the README headline against `catalog.json` would stop it going stale again.

---
Approved and merged.

PR #433, opened 2026-09-13
fix(mendelian-randomisation): estimators say when they do not apply; weighted mode and Steiger follow the cited methods term by term

madaraviv:
## TL;DR

Housekeeping on the `mendelian-randomisation` skill to bring three estimators in line with the TwoSampleMR conventions they cite. Below 3 instruments, MR-Egger, weighted median and weighted mode now report "not applicable" instead of a number, as TwoSampleMR does; the weighted mode and the Steiger test follow their papers term by term; and four PubMed links in the citations list are corrected. The 30-instrument demo is unchanged except for the weighted mode's standard error, which is now the paper's bootstrap. Nothing here changes a result for a well-conditioned multi-instrument analysis.

## Summary

In a few edge cases (fewer than 3 instruments, near-identical exposure effects) three estimators returned a value where the estimator is not defined. This PR makes each say so instead, and brings the weighted mode and the Steiger test into line with their cited papers as implemented in TwoSampleMR. Every changed term is named below with the function it follows, so it can be checked against the reference rather than against this description.

- **MR-Egger** below 3 instruments, or with exposure effects too close to identical to identify a slope, is reported as not applicable with the reason. Previously it ran regardless, and at one instrument the value it returned was numerically meaningless.
- **Weighted median and weighted mode** below 3 instruments are likewise reported as not applicable. This is the one change that goes beyond the original defects; the reason is given below.
- **Steiger** no longer depends on the units the outcome is reported in.
- **Weighted mode** uses the paper's bandwidth rule, weights, ratio SE, and bootstrap standard error.
- Four of the eight PMIDs in `## Citations` pointed at unrelated papers; corrected and verified.

The 30-instrument demo is unchanged for IVW, MR-Egger and Weighted Median to the printed precision; Weighted Mode moves from 0.5989 (se 0.0144) to 0.6031 (se 0.0705). `SKILL.md`'s example output was regenerated from the demo and matches it row for row.

### MR-Egger with fewer than 3 instruments

Egger fits a slope and an intercept, so with one or two instruments there is no residual degree of freedom. At one instrument the denominator `sum_w*sum_wbx2 - sum_wbx**2` is algebraically zero but computed as the difference of two nearly equal numbers, so it lands on a rounding remainder of arbitrary sign; the `abs(denom) < 1e-300` guard caught it only sometimes. Reproduced on the shipped code with one instrument:

```
method     estimate   se               ci_lower        ci_upper
MR-Egger   0.000000   9268190.002368   -18165652.40    18165652.40
```

The commit that introduced the guard (`bc7b91f`) measured the three outcomes over 5,000 random single-instrument inputs: the guard firing (~45%), `math.sqrt` raising and the process exiting non-zero with the IVW already computed discarded (~27%), and a finite slope with a standard error in the millions written to every output (~27%). At two instruments the run exited 0 having written `"se": Infinity`, which is not valid JSON.

**Change.** Below 3 instruments the estimator is reported as `{"method": "MR-Egger", "applicable": false, "reason": ..., "n_snps": n}` with no numeric fields, as `not_applicable` in every numeric column of `mr_results.tsv` with a `note`, as a named absence in the report, and with no line or legend entry in the scatter plot. `result.json` is written with `allow_nan=False` so a future non-finite fails loudly rather than shipping. The regression itself uses the centered form of the denominator, `sum_w * sum_i w_i (bx_i - xbar_w)^2`, which cannot go negative and is exactly zero when every exposure effect is equal; and a dimensionless conditioning test, `denom / (sum_w * sum_wbx2)` against `sqrt(machine epsilon)`, written as `not (rel > t)` so a NaN (reachable from a zero outcome SE) is refused. TwoSampleMR `mr_egger_regression` returns NA below 3 complete SNPs. Bowden, Davey Smith & Burgess 2015 (doi:10.1093/ije/dyv080; PMID 26050253).

### Weighted median and weighted mode with fewer than 3 instruments

TwoSampleMR declines all three sensitivity estimators below 3 SNPs, not only Egger:

```r
mr_egger_regression:  if (sum(!is.na(b_exp) & ...) < 3) return(nulllist)
mr_weighted_median:   if (sum(!is.na(b_exp) & ...) < 3) return(list(b = NA, se = NA, pval = NA, nsnp = NA))
mr_mode:              if (nrow(dat) < 3) { warning("Need at least 3 SNPs"); return(NULL) }
```

The median and the mode are order statistics of the ratios; with one instrument each returns the single ratio it was given. On the shipped code, one instrument produced

```
IVW               0.500000   se 0.100000
MR-Egger          0.000000   se 9268190
Weighted Median   0.500000   se 0.098872
Weighted Mode     0.498999   se 0.200000
```

and the report counted the three as agreeing with IVW: "Sensitivity analyses show consistent estimates across IVW, MR-Egger, weighted median, and weighted mode, supporting a robust causal inference." The median and mode rows restate the single Wald ratio. Now:

```
IVW               0.500000   se 0.100000
MR-Egger          not_applicable   note: requires at least 3 instruments; this analysis has 1
Weighted Median   not_applicable
Weighted Mode     not_applicable
```

and the report says: "No sensitivity estimator applies to this instrument set, so the IVW estimate stands alone and is not corroborated." IVW is the one estimator defined at n = 1, where it is the Wald ratio.

### Steiger directionality

The test compares the fraction of variance the instruments explain in the exposure against the fraction they explain in the outcome. The proper quantity is `r2 = 2p(1-p) * beta^2 / Var(trait)`. The code had the numerator without the denominator, so the two sides were compared in the traits' own units, and the verdict moved with a unit change: the same ten instruments read "correct direction" with the outcome at x1 and "REVERSED" at x3, x10 and x100, with p near zero on both sides.

The input format carries no trait variance, so `r2` is obtained from beta, se and n instead, as TwoSampleMR `get_r_from_bsen` does: `F = (b/se)^2; r2 = F / (n - 2 + F)`. That is the same quantity, with the standard error supplying the trait variance the input does not carry, and it is unit-free because beta and se rescale together. `Instrument` gains optional `n_exposure` / `n_outcome`, read from the input JSON when present.

- With sample sizes: `r = sqrt(sum r2)` per side and the test for two independent correlations, Fisher's z-transform with `sqrt(1/(n1-3) + 1/(n2-3))`, per-instrument n aggregated by the mean, as TwoSampleMR `mr_steiger` does through `psych::r.test`. The paper states the one-sample form (Steiger's Z for correlated correlations); its two-sample implementation uses the independent-samples test, and so does this.
- Without sample sizes: the direction is read from `sum(z^2)` under the stated assumption that the two studies are of comparable size, with that assumption in a `steiger_note`, and the p-value is `null`.

Two things worth stating plainly. This removes an artefact; it does not make the test powerful for a molecular exposure against a complex-disease outcome, where the outcome-side `r2` is smaller than the exposure-side almost regardless of causation (a cis-eQTL can explain a large share of a gene's expression; a disease SNP explains a fraction of a percent, before phenotyping noise). Hemani 2017 discusses the measurement-error bias in the same direction. In that setting "correct" carries little information and "REVERSED" is the informative verdict. And the conversion is the continuous-trait one; for a binary outcome it is approximate, as TwoSampleMR's own warning says, and the stopgap is to pass an effective n. Hemani, Tilling & Davey Smith 2017 (doi:10.1371/journal.pgen.1007081; PMID 29149188).

**A question for review.** The previous p-value came from `se_diff = sqrt(r2_exp + r2_out) * 0.01`. We could not find the 0.01 in the paper or in TwoSampleMR. If it has a source, we would rather keep it and would like to know.

### Weighted mode

The standard error was `2 / sum(|bx|/sy)`, a function of the instrument count and the outcome SEs and not of the ratios: 0.0138 for fifteen instruments that all agree, and 0.0138 for the same fifteen split into two camps. The bandwidth was an absolute 0.5, wider than the spread of ordinary ratios, so the kernels merged and the "mode" sat near the weighted mean (0.410 against a mean of 0.438 on a 9-to-6 split whose mode is 0.20).

Now the paper's weighted MBE as TwoSampleMR `mr_mode` implements it (Hartwig, Davey Smith & Bowden 2017, doi:10.1093/ije/dyx102; PMID 29040600):
- ratio SE by the second-order delta method, `sqrt(sy^2/bx^2 + by^2*sx^2/bx^4)`, the "not assuming NOME" column the reference uses for the weighted mode;
- standardised inverse-variance weights, eq. 5, `w_j = se_j^-2 / sum(se^-2)` (the skill had `1/se`);
- bandwidth `h = phi * s` with the modified Silverman rule, eq. 7, `s = 0.9 * min(sd, 1.4826*mad) / L^(1/5)`, floored at 1e-8, default `phi = 1`;
- parametric bootstrap, each ratio redrawn from N(ratio, se_ratio) with the bandwidth recomputed per draw; SE = `1.4826 * mad` of the draws; p from a t distribution on L - 1 degrees of freedom.

On the 9-to-6 split the mode is now 0.196 (the larger camp) and the two datasets above get SEs of 0.060 and 0.038.

### Citations

Four of the eight PMIDs in `## Citations` resolved to unrelated records: an ApoE/stroke meta-analysis for Burgess 2013, a biomimetic-catalysis paper for Bowden 2016, a circadian food-intake study for Hemani 2017, and the JAMA STROBE-MR statement where the text cites BMJ 375:n2233. The author, year and journal text beside each link were correct, which is what makes this class of error easy to miss on review and a known failure mode when citations are written from recall, LLM-assisted or not. Each replacement was resolved through Europe PMC and matched on authors, title, journal and year. The check we run before any citation lands is public and CC0: https://gist.github.com/madaraviv/97d4b98c17eccba99804108c53fc3fc7

## Not changed, for awareness

- The weighted median's weights use the first-order ratio variance `sy^2/bx^2`; `mr_weighted_median` uses the second-order form. Left as is because the PR does not otherwise touch the median.
- MR-Egger here does not orient instruments to a positive exposure effect before the regression; `mr_egger_regression` does.

Either can be added if wanted.

## Verification

- 42 tests in `skills/mendelian-randomisation/tests/`, on Python 3.11 and 3.14, 0 warnings.
- Each guard observed failing with it removed: 8 mutations for the Egger change, 9 for the weighted-mode/Steiger change (bandwidth rule, weight exponent, ratio-SE order, MAD vs sd, t vs normal, n - 2, mean vs min, and each n < 3 guard).
- `agentskills validate skills/mendelian-randomisation`: valid. `python clawbio.py list` loads.
- The demo was regenerated and diffed against `SKILL.md`'s example table.
- The `test (3.11)` job fails on `tests/test_public_claims.py::test_demo_and_cli_counts_match_catalog` (README says 91 demos, catalog has 92). That test fails identically on `main` at `1fcecb7`, and this PR does not touch README or the catalog.

## Consumers

Anything that reads `result.json` and expects `estimate` and `se` on every row must check `applicable` first; it is in the Gotchas.

🤖 Generated with [Claude Code](https://claude.com/claude-code); reviewed by Aviv Madar (@madaraviv)

### DIFF
--- a/skills/mendelian-randomisation/SKILL.md
+++ b/skills/mendelian-randomisation/SKILL.md
@@ -116,7 +116,7 @@ You are **Mendelian Randomisation**, a specialised ClawBio agent for causal infe
 
 | Format | Extension | Required Fields | Example |
 |--------|-----------|-----------------|---------|
-| Harmonised instruments JSON | `.json` | SNP, effect_allele, other_allele, eaf, beta_exposure, se_exposure, pval_exposure, beta_outcome, se_outcome, pval_outcome | `demo_instruments.json` |
+| Harmonised instruments JSON | `.json` | SNP, effect_allele, other_allele, eaf, beta_exposure, se_exposure, pval_exposure, beta_outcome, se_outcome, pval_outcome; optional `n_exposure` / `n_outcome` (sample sizes, needed for a Steiger p-value) | `demo_instruments.json` |
 
 ## Workflow
 
@@ -154,14 +154,15 @@ Expected output: A full MR report for 30 synthetic BMI → T2D instruments showi
 
 1. **IVW**: beta = sum(w * bx * by) / sum(w * bx²), with multiplicative random-effects variance inflation (Burgess et al., 2013)
 2. **MR-Egger**: Weighted linear regression of by on bx with intercept; slope = causal estimate, intercept = pleiotropy (Bowden et al., 2015). Reported as **not applicable**, with a stated reason, when it is undefined on the given instruments: fewer than 3 of them (it fits two parameters, so below 3 there is no residual degree of freedom), or exposure effects too close to identical for the slope to be identified. Never a number in those cases.
-3. **Weighted Median**: Median of Wald ratios weighted by inverse-variance; consistent when ≥50% weight from valid instruments (Bowden et al., 2016)
-4. **Weighted Mode**: Kernel density mode of weighted Wald ratios (Hartwig et al., 2017)
+3. **Weighted Median**: Median of Wald ratios weighted by inverse-variance; consistent when ≥50% weight from valid instruments (Bowden et al., 2016, doi:10.1002/gepi.21965; PMID 27061298)
+4. **Weighted Mode**: Kernel density mode of weighted Wald ratios, with a bandwidth proportional to the spread of the ratios (`phi` x their standard deviation) and a bootstrapped standard error (Hartwig et al., 2017)
 
 **Key thresholds**:
 - F-statistic > 10 for instrument strength (Staiger & Stock, 1997)
 - I²_GX > 0.9 for MR-Egger validity; SIMEX recommended below (Bowden et al., 2016)
 - Cochran's Q P < 0.05 indicates heterogeneity
 - Egger intercept P < 0.05 indicates directional pleiotropy
+- Steiger directionality is computed from z-statistics, so it does not depend on the units the traits are reported in; supply `n_exposure` and `n_outcome` per instrument for a p-value, without them only the direction is reported
 - MR-Egger needs >= 3 instruments and at least two distinct exposure effects; below either it is not applicable rather than imprecise
 
 ## Example Output
--- a/skills/mendelian-randomisation/SKILL.md
+++ b/skills/mendelian-randomisation/SKILL.md
@@ -153,7 +153,7 @@ Expected output: A full MR report for 30 synthetic BMI → T2D instruments showi
 ## Algorithm / Methodology
 
 1. **IVW**: beta = sum(w * bx * by) / sum(w * bx²), with multiplicative random-effects variance inflation (Burgess et al., 2013)
-2. **MR-Egger**: Weighted linear regression of by on bx with intercept; slope = causal estimate, intercept = pleiotropy (Bowden et al., 2015)
+2. **MR-Egger**: Weighted linear regression of by on bx with intercept; slope = causal estimate, intercept = pleiotropy (Bowden et al., 2015). Reported as **not applicable**, with a stated reason, when it is undefined on the given instruments: fewer than 3 of them (it fits two parameters, so below 3 there is no residual degree of freedom), or exposure effects too close to identical for the slope to be identified. Never a number in those cases.
 3. **Weighted Median**: Median of Wald ratios weighted by inverse-variance; consistent when ≥50% weight from valid instruments (Bowden et al., 2016)
 4. **Weighted Mode**: Kernel density mode of weighted Wald ratios (Hartwig et al., 2017)
 
@@ -162,6 +162,7 @@ Expected output: A full MR report for 30 synthetic BMI → T2D instruments showi
 - I²_GX > 0.9 for MR-Egger validity; SIMEX recommended below (Bowden et al., 2016)
 - Cochran's Q P < 0.05 indicates heterogeneity
 - Egger intercept P < 0.05 indicates directional pleiotropy
+- MR-Egger needs >= 3 instruments and at least two distinct exposure effects; below either it is not applicable rather than imprecise
 
 ## Example Output
 
@@ -255,7 +256,7 @@ The agent dispatches and explains. The skill (Python) executes. The agent must N
 
 **Chaining contract**:
 - **Input**: JSON with `instruments` array; each instrument has `SNP`, `beta_exposure`, `se_exposure`, `pval_exposure`, `beta_outcome`, `se_outcome`, `pval_outcome`, `effect_allele`, `other_allele`, `eaf`, `f_statistic`
-- **Output**: `result.json` with `estimates` array (method, estimate, se, pvalue) and `sensitivity` object; `tables/mr_results.tsv` for downstream consumption
+- **Output**: `result.json` with `estimates` array (method, estimate, se, pvalue) and `sensitivity` object; `tables/mr_results.tsv` for downstream consumption. An estimator that does not apply appears as `{"method", "applicable": false, "reason", "n_snps"}` with no numeric fields, and as `not_applicable` in every numeric column of the TSV plus a `note` column. `result.json` is written with `allow_nan=False`, so it is always valid JSON per RFC 8259 or it is not written at all.
 
 ## Maintenance
 
--- a/skills/mendelian-randomisation/SKILL.md
+++ b/skills/mendelian-randomisation/SKILL.md
@@ -155,7 +155,7 @@ Expected output: A full MR report for 30 synthetic BMI → T2D instruments showi
 1. **IVW**: beta = sum(w * bx * by) / sum(w * bx²), with multiplicative random-effects variance inflation (Burgess et al., 2013)
 2. **MR-Egger**: Weighted linear regression of by on bx with intercept; slope = causal estimate, intercept = pleiotropy (Bowden et al., 2015). Reported as **not applicable**, with a stated reason, when it is undefined on the given instruments: fewer than 3 of them (it fits two parameters, so below 3 there is no residual degree of freedom), or exposure effects too close to identical for the slope to be identified. Never a number in those cases.
 3. **Weighted Median**: Median of Wald ratios weighted by inverse-variance; consistent when ≥50% weight from valid instruments (Bowden et al., 2016, doi:10.1002/gepi.21965; PMID 27061298)
-4. **Weighted Mode**: Kernel density mode of weighted Wald ratios, with a bandwidth proportional to the spread of the ratios (`phi` x their standard deviation) and a bootstrapped standard error (Hartwig et al., 2017)
+4. **Weighted Mode**: Mode of the inverse-variance weighted kernel density of the Wald ratios, bandwidth `phi` x the modified Silverman rule `0.9 min(sd, 1.4826 mad) / L^(1/5)`, standard error from a parametric bootstrap (Hartwig et al., 2017, doi:10.1093/ije/dyx102; PMID 29040600; as implemented in TwoSampleMR `mr_weighted_mode`)
 
 **Key thresholds**:
 - F-statistic > 10 for instrument strength (Staiger & Stock, 1997)
@@ -163,7 +163,7 @@ Expected output: A full MR report for 30 synthetic BMI → T2D instruments showi
 - Cochran's Q P < 0.05 indicates heterogeneity
 - Egger intercept P < 0.05 indicates directional pleiotropy
 - Steiger directionality is computed from z-statistics, so it does not depend on the units the traits are reported in; supply `n_exposure` and `n_outcome` per instrument for a p-value, without them only the direction is reported
-- MR-Egger needs >= 3 instruments and at least two distinct exposure effects; below either it is not applicable rather than imprecise
+- MR-Egger, weighted median and weighted mode each need >= 3 instruments (MR-Egger also needs at least two distinct exposure effects); below that each is reported as not applicable rather than as a number. IVW is defined at n = 1, where it is the single Wald ratio
 
 ## Example Output
 
@@ -181,7 +181,7 @@ Expected output: A full MR report for 30 synthetic BMI → T2D instruments showi
 | IVW | 0.5979 | 0.0369 | [0.5255, 0.6702] | 5.17e-59 |
 | MR-Egger | 0.5989 | 0.0391 | [0.5223, 0.6756] | 6.62e-53 |
 | Weighted Median | 0.6001 | 0.0469 | [0.5081, 0.6921] | 2.07e-37 |
-| Weighted Mode | 0.5989 | 0.0144 | [0.5708, 0.6271] | 0.00e+00 |
+| Weighted Mode | 0.6031 | 0.0705 | [0.4648, 0.7413] | 2.03e-09 |
 
 ## Sensitivity Analysis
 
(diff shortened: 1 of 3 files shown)


madaraviv:
@RezaJF if you can take a look here and approve/edit i'll appreciate it. this is based on my usage of the skill. 

Thanks!
Aviv

madaraviv:
@manuelcorpas This is a needed fix as the ebi catalog took off the api path. This workaround is based on a discussion i had with them, see  https://github.com/eQTL-Catalogue/eQTL-Catalogue-resources/issues/59. Let me know if anything else is needed. Thanks! Aviv

manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Thank you, this is careful work. I checked the weighted mode against TwoSampleMR `mr_mode`. The second-order ratio SE, standardised inverse-variance weights, the modified Silverman bandwidth at phi = 1, the scaled MAD of the parametric bootstrap and the t on L - 1 all match. The Steiger r² conversion and the Fisher z for independent correlations match `get_r_from_bsen` and `mr_steiger`. The centred Egger algebra is correct. Returning not applicable below 3 SNPs matches the reference. The corrected PMIDs (24114802, 27061298, 29149188, 34702754) are right.

1. When no sample sizes are given, `_write_report_md` still prints "Exposure → Outcome confirmed" for Steiger. Summed z² favours the larger GWAS, and the exposure GWAS usually is larger. When the p-value is None, "direction consistent, not tested" would be more accurate.
2. The estimates table loop in `_write_report_md` (mendelian_randomisation.py:506 on main) is unchanged, so rows that do not apply show as `nan` in report.md. A `not computed` row would match the TSV and the JSON.
3. This one predates your change: `mr_egger` (mendelian_randomisation.py:134 on main) does not flip alleles so that `beta_exposure` is positive before fitting, as TwoSampleMR does. Without that, the intercept depends on allele coding.

Scanner: its only warning is a CI signal with no code location. `test (3.13)` failed, but the test jobs also fail on several unrelated open PRs, so I cannot tell from here whether this change caused it.

---

madaraviv:
Thanks for the careful check. All three suggestions are now on the branch as new commits, plus a merge of main:

1. Steiger with no sample sizes: the report and the sensitivity TSV now say "Direction consistent with exposure → outcome; significance not assessable without sample sizes" (the reverse-causation warning is worded the same way): the comparison was made and gave a direction, and what is missing is its p-value. "Confirmed" is reserved for a computed p-value. One note on the reasoning: the z-squared comparison stands in for variance explained only if the studies are of comparable size, and z-squared grows with n, so when the outcome GWAS is the larger study, the usual case for a disease outcome, the comparison leans toward "reversed" rather than toward the exposure from a statistical standpoint; however a cis-QTL exposure typically wins regardless because a few variants explain a large share of expression.
2. report.md prints "not computed" in every cell of an estimator row that did not apply, matching the TSV and the JSON; the reason is stated below the table as before.
3. MR-Egger now orients instruments to a positive exposure effect before the fit, flipping the outcome effect with it, as mr_egger_regression does. To check it against the reference rather than against itself, I loaded mr_egger_regression verbatim from TwoSampleMR's R/mr.R (commit c14776b8) into R 4.6.0 and ran it on the 30 demo instruments: slope 0.602231, SE 0.081583, intercept -0.000187, intercept SE 0.003121. The branch now returns the same four values to six decimals, pinned as a parity test; the previous code returned slope 0.598932 and SE 0.039122, which match nothing in that implementation. A second test re-codes every other demo instrument to its other allele and checks the fit is unchanged; it fails on the previous code. So the "demo byte-identical" sentence in the description no longer holds for MR-Egger; IVW, weighted median and weighted mode are unchanged. That comparison also showed the p-values on a different reference: TwoSampleMR uses a t on n - 2 df for the Egger slope and intercept, this skill used a normal. A further commit switches Egger to the t (IVW and the weighted median stay normal, the weighted mode stays t on n - 1, matching TwoSampleMR on all four). The demo's slope p goes from 1.6e-13 to 4.9e-08, TwoSampleMR's value, now pinned in the parity test; on three instruments, where Egger has one residual degree of freedom, the normal called the slope significant (p 0.006) and the t gives 0.22.

The red test job was tests/test_public_claims.py (README says 91 skills with demo data, catalog has 92); the branch was five commits behind main, and merging main clears it. The CodeQL note about implicit string concatenation was on an earlier commit and is already a single string at the current head.

RezaJF:
@madaraviv Thank you — this is excellent work, and I say that as the person who wrote the code it corrects. Four of these five defects are mine. The two that matter most are the ones I would never have found by reading my own code: a Steiger verdict that moves with the outcome's units, and a weighted-mode standard error that is identical for a tight cluster and a two-camp split. Checking against TwoSampleMR term by term rather than against the papers' prose is the right standard, and I'll hold to it in this skill from here.

**On your question about the `0.01`.** It has no source. `se_diff = sqrt(r2_exp + r2_out) * 0.01` was a placeholder I wrote to get the p-value onto a plausible scale, and it should never have shipped looking like a derived quantity. Please drop it — returning `None` with the assumption stated in `steiger_note` is strictly better than a number with nothing behind it. Thank you for asking rather than silently keeping it; that is the more expensive thing to do and it was the right call.

I reviewed the substance and agree with all of it. I checked each of these against the reference rather than against your description of it: the centred Egger denominator and the dimensionless conditioning ratio, `r2` from the z-statistic per `get_r_from_bsen`, the independent-samples Fisher z per `mr_steiger`, the modified Silverman bandwidth, standardised inverse-variance weights, the second-order delta-method ratio SE, the scaled MAD of the parametric bootstrap, the t on L − 1, and declining all three sensitivity estimators below 3 SNPs. The allele orientation and the t on n − 2 for Egger are right too, and pinning them against `mr_egger_regression` loaded verbatim into R is the correct way to have done it. I also re-resolved the four PMIDs through Europe PMC independently — 24114802, 27061298, 29149188, 34702754 all match on authors, title, journal and year, and the BMJ record is the right STROBE-MR for the text beside it.

I reproduced your measurements on the shipped code before reviewing, and they land where you said, including the `~45% / ~27% / ~27%` split at one instrument and the median SE of order `10^6` on the third branch.

**One thing your branch leaves open**, and it is the same class of defect at the one instrument count the guards do not cover: `n = 0`. `MIN_SENSITIVITY_INSTRUMENTS` gates 1 and 2, IVW is genuinely correct at 1, and nothing gates 0 — which is reachable from the CLI with a well-formed file whose `instruments` array is empty, and more realistically from any upstream p-value threshold, LD clumping or harmonisation step that removes every SNP. On `8326dda`:

```
ivw([])            MREstimate(estimate=nan, se=inf, applicable=True)
steiger_test([])   (False, nan, '')
CLI, empty array   "0 instruments loaded", four "Running ..." lines, then an
                   unhandled ZeroDivisionError from compute_i_squared_gx
```

The `applicable=True` is the shape `MREstimate` exists to prevent, and that NaN p-value would reach your own `allow_nan=False` and turn a bad input into a crash at the last step. Three causes: `ivw` does not gate `n = 0`; `compute_i_squared_gx` tests `q <= df or df == 0`, but `df` is `-1` at `n = 0`, so both routes past the guard are open; and `steiger_test`'s `all(...)` over an empty list is vacuously True, so it takes the branch that needs sample sizes and exits through `np.mean([])`.

I also found two boundaries your suite does not pin. Widening `MIN_SENSITIVITY_INSTRUMENTS` to 4 fails nothing, because every test expecting a number from the median and mode uses 12, 15 or 30 instruments. And the mixed case — one estimator declining while the others stand — is never reached: the demo has all four applying and the two-instrument run has only IVW, so the branch of your report comparison where `comparable` and `skipped` are both non-empty never executes.

All of that is now in #435, stacked on this branch so it does not hold you up: 12 tests, 57 total, 0 warnings under `-W error::RuntimeWarning`, seven mutations each observed failing, demo unchanged and still matching your four parity values to six decimals. Merge this one first and #435 reduces to a single commit. One note you will appreciate: my "threshold narrowed to 2" mutation first appeared to pass, because `= 3` and `= 2` are the same byte length and Python reused cached bytecode — the mutant never ran. Exactly the failure your own test comments warn about.

Two judgement calls in #435 I would value your view on. `run_pipeline` raises rather than writing a report, on the reasoning that zero instruments is the absence of the analysis rather than an analysis with unavailable estimators, so there is nothing to report — and it refuses before creating the output directory so a failed run leaves no artefacts. And on your two "not changed, for awareness" items: I would take the weighted median's second-order ratio variance too, since `mr_weighted_median` uses it, but in its own PR rather than growing either of these.

Approving on substance. I do not have write access here, so this needs a maintainer to merge — @manuelcorpas, this one is ready as far as I can tell, and the audit's three suggestions plus the allele-orientation item it raised have all been addressed on the branch.

manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Thank you, Aviv. I checked the MR-Egger changes against TwoSampleMR `mr_egger_regression`. Flipping instruments to positive exposure effects, `se = sqrt(phi * unscaled)` with `phi = max(1, sigma^2)`, and a t reference on n - 2 df all match. The parity test against R output locks that in. The weighted mode now follows Hartwig et al. 2017: inverse-variance weights, a Silverman bandwidth recomputed on each draw, a scaled MAD of the bootstrap and a t on L - 1. Below 3 SNPs these methods now report not applicable instead of a number, and `allow_nan=False` stops invalid JSON being written silently. The corrected PMIDs are right; the old STROBE-MR one was the JAMA version.

The static scan reported no findings, so there were none to re-check.

1. `mendelian_randomisation.py:552` turns z into r2 with the formula for continuous traits. For a binary outcome in log odds, like the T2D demo, TwoSampleMR uses `get_r_from_lor` with prevalence. The formula used here understates the outcome's r2, which pushes the test towards "confirmed". A SKILL.md note or a units flag would help once people supply `n_outcome`.
2. `SKILL.md:193` still shows "Confirmed" for the demo Steiger row. Without sample sizes the code now writes "Direction consistent with exposure → outcome; significance not assessable without sample sizes".
3. The comment at `mendelian_randomisation.py:230` says a zero exposure effect keeps sign +1 in TwoSampleMR. In R, `sign(0)` is 0. This does no harm here, but the comment is wrong.

---
Approved and merged.

PR #434, opened 2026-09-13
fix(eqtl-catalogue-region-fetch): dataset metadata from a bundled index; the catalogue's metadata API is retired

madaraviv:
## TL;DR

The eQTL Catalogue retired its metadata REST API in September 2026 (HTTP 410; the maintainers confirmed it on our issue, [eQTL-Catalogue-resources#59](https://github.com/eQTL-Catalogue/eQTL-Catalogue-resources/issues/59)), and this skill called it on every fetch, so every fetch now fails. This PR moves dataset metadata to a small table bundled with the skill, derived from the catalogue's own published dataset table, and reads the per-variant file class from it instead of inferring it from the quantification method. The FTP fetch path is unchanged: a fresh fetch of the SORT1 demo is byte-identical to a pre-September result.

## What this fixes

`fetch_region` resolved the study directory and the quantification method through `https://www.ebi.ac.uk/eqtl/api/v2/datasets/{id}` before opening the FTP file. The catalogue disabled that API permanently; we asked on [eQTL-Catalogue-resources#59](https://github.com/eQTL-Catalogue/eQTL-Catalogue-resources/issues/59) when it started failing in August, and the maintainers confirmed on 2026-09-09 that it will not return:

```
$ python eqtl_catalogue_region_fetch.py --demo --no-cache --output /tmp/x
requests.exceptions.HTTPError: 410 Client Error: Gone for url: https://www.ebi.ac.uk/eqtl/api/v2/datasets/QTD000276
```

A result cache from before September (`~/.clawbio/eqtl_catalogue_region_fetch_cache`) masks this for regions fetched earlier, which is why the bundled demo can still appear to work on a machine that ran it in the spring.

## What changed

**Metadata comes from a table bundled with the skill, `data/dataset_index_r7.tsv`, derived from the catalogue's published [`tabix/tabix_ftp_paths.tsv`](https://github.com/eQTL-Catalogue/eQTL-Catalogue-resources/blob/master/tabix/tabix_ftp_paths.tsv).** 758 datasets, one row each: `study_id`, labels, sample size, `quant_method`, and `file_class`, the per-variant file the catalogue's table lists for that dataset (`all` or `cc`). It is derived from the catalogue's published dataset table, [`tabix/tabix_ftp_paths.tsv`](https://github.com/eQTL-Catalogue/eQTL-Catalogue-resources/blob/master/tabix/tabix_ftp_paths.tsv) in the [eQTL-Catalogue-resources](https://github.com/eQTL-Catalogue/eQTL-Catalogue-resources) repository (raw: https://raw.githubusercontent.com/eQTL-Catalogue/eQTL-Catalogue-resources/master/tabix/tabix_ftp_paths.tsv; r7, retrieved 2026-08-19, byte-identical to the file served on 2026-09-13): the three FTP URL columns are dropped, since they are 77.6% of the bytes and reconstruct from `study_id`, `dataset_id` and `file_class` (asserted equal to the source on 758 of 758 rows for all three URL families at derivation), and `file_class` is added, read off the published path. 66,786 bytes, under the 256 KB the wheel build hook allows for skill data. Provenance, licence (CC-BY-4.0) and attribution are in `data/dataset_index_r7.provenance.json`.

**The file class is read, not inferred.** `ftp_url_for` takes `file_class` instead of `quant_method`. The old rule (`.all` for `ge`/`microarray`, `.cc` otherwise) disagrees with the catalogue's table for one r7 dataset, QTD000584 (aptamer, Sun 2018), listed as `.all`. Probed 2026-09-13 on 33 of the 758 datasets (30 drawn at random, stratified 15 per listed class, plus the three above): every dataset listed `.all` also serves a `.cc` file, and no dataset listed `.cc` serves an `.all` file. So for every `.all` dataset the wrong guess opens a file that exists and holds different rows, so opening the wrong one substitutes the credible-set-filtered rows for the full ones without an error. Measured on QTD000584 over the 1 Mb SORT1 locus (chr1:108.77-109.77 Mb, GRCh38): `.all` holds 33,240 rows across 11 proteins, `.cc` holds 3,892 rows for 1 protein, 11.7% of the rows and 1 of the 11 traits. The 3,892 `.cc` rows all belong to the GSTM1 aptamer (the one trait with a credible set there); SORT1's own aptamer is not in `.cc` for this region, so under the old rule a query for the SORT1 protein in this dataset returned zero rows, with no error. The index carries the file the catalogue lists.

**An id outside the bundled release is stated, not guessed.** `fetch_dataset_metadata` raises `EQTLCatalogueDatasetNotFound` (a subclass of `EQTLCatalogueAPIError`, so existing handlers keep working) with the instruction to pass `study_id` and `file_class`. `fetch_region` gains `file_class`; passing both bypasses the index entirely, which is the route for a dataset added after r7. That route is a bridge, not the plan: when the catalogue publishes r8 we will re-derive the bundled table for it in a follow-up PR, so that datasets added in the new release resolve on their own; the explicit override is there so nothing is unreachable in between.

**Smaller things.** `requests` is no longer needed and is dropped from `environment.yml`; `api_base` and `session` are still accepted by the client constructor so existing callers keep working, and ignored. The manifest's `dataset_release` is the index release, `r7`. `.gitignore` gains the two data files, following the repo's existing per-skill allow-list entries. `SKILL.md`: the resolution step and endpoints now describe the bundled table, and a Gotcha covers the retired API, the out-of-release route and the cache.

## Verification

- Fresh fetch (`--no-cache`) on the new code: SORT1 demo 2,833 variants, `variants.tsv` byte-identical to the pre-September cached result; IRF5 demo 3,276 variants, the count the original submission reported.
- 26 offline tests; the dataset ids used are real rows of the bundled index, so no metadata mocking. Plus a `test_live_eqtl_catalogue_region_fetch.py` in the same shape as the sibling region-fetch skills already carry (`live` marker, opt-in via `RUN_LIVE_TESTS=1`, skipped by default): a real fetch of the SORT1 demo region, and a check that the FTP serves the index the table lists for QTD000584. This skill had no live test, which is how it could sit broken against its source with a green suite. Five mutations each observed failing at least one test: file class inferred instead of read, the explicit-bypass branch removed, not-found collapsed to the generic error, the column assertion removed, the missing-file check removed.
- `agentskills validate skills/eqtl-catalogue-region-fetch`: valid. `python clawbio.py list` loads. Python 3.11 and 3.14.

## Not in this PR

The skill and the bundled table are on r7, the catalogue's current release; the FTP still serves the r7 tabix-indexed files this skill reads. The catalogue has announced that a future r8 will move the bulk files to parquet without tabix indexes. That has not landed. If and when it does, the fetch path will need its own change, and we will bring that as a separate PR at that time.

🤖 Generated with [Claude Code](https://claude.com/claude-code); reviewed by Aviv Madar (@madaraviv)


manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  
Static scan: 0 critical, 2 warning. Domain review: complete.

Thank you, this is a careful fix. `fetch_region` now reads `file_class` from the catalogue's own table instead of inferring it from `quant_method` (`eqtl_catalogue_region_fetch.py:373-380`). That removes a silent wrong-file substitution for QTD000584, and the new test pins it to a real row rather than a mock. I checked all 758 rows: every `ge` and `microarray` dataset is `all`, the splicing and transcript methods are `cc`, and the aptamer dataset is the only exception. The loader fails loudly on a missing table or changed columns. The only network call left is the tabix read from the public FTP.

Scanner findings: the unpinned `pip install` in `SKILL.md` was already there (you only removed `requests`). The `urllib` import is a HEAD request to a public URL in a live test that is skipped by default (`conftest.py:13-21`). Neither is a defect. The `test (3.13)` job failed on this commit; I could not see its log, so please check it.

1. `dataset_index_r7.provenance.json:43` says "MODIFICATIONS: none" and calls the file byte-identical to the source. Line 27 says columns were dropped and added, and line 44 lists 12 columns where the file has 10. CC-BY 4.0 asks that changes be stated, so this is worth correcting.
2. `locuscompare_region_render.py:527` still bases the `.cc` caveat on `quant_method`, so QTD000584 would be labelled credible-set-filtered even though `.all` was opened. Carrying `file_class` on the release would fix this.

---

madaraviv:
Thanks. Both suggestions are on the branch as a new commit, plus a merge of main:

1. The provenance record now states the derivation in one voice: the attribution text says the file is derived (three FTP URL columns dropped, file_class added from the published file name, everything else byte-identical) and points at the modifications field, and the column list is the file's ten. The sha256 is of the derived file, source_sha256 of the retrieved source.
2. EQTLCatalogueRelease now carries file_class, set from the file fetch_region actually opened and restored from cache, with a test. The consumer side you pointed at (locuscompare_region_render.py basing the .cc caveat on quant_method) is a change to that skill and will come as its own PR against it, reading release.file_class.

The red test job was tests/test_public_claims.py (README 91 vs catalog 92); the branch was five commits behind main and merging main clears it.

madaraviv:
One more commit: file_class now rides on the RegionResult rather than on the release object. It describes this fetch (which file was opened for this region), not the dataset's release metadata, and the result is where a consumer, the locuscompare-region-render skill, reads it. Set at construction, restored from cache, serialised in to_dict; the test follows it. 26 passed, 2 skipped (live).

This is a dependency for locuscompare-region-render: the copy on main today infers its credible-set-filtered caveat from quant_method (the mislabel you pointed at for QTD000584), and the sync PR I will open for that skill after this one lands reads result.file_class instead, so it needs this PR merged first.

madaraviv:
Small follow-up while I was poking at the two file classes: the CLI now accepts `study_id` and `file_class` in the config (the Python API already did; the config reader just wasn't passing them along), and it tells you if a config carries a key it doesn't know, so a typo can't slip by. The manifest records which file was opened. Two tests added. 28 passed, 2 skipped.

madaraviv:
Docstring only: `.cc` is now described from the catalogue's own definition (docs/Columns_parquet.md) instead of our earlier paraphrase. No behaviour change, no test moves. This just adds more info and references to the source files.

manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Thanks, this is a careful fix. The skill now reads the per-variant file class from the catalogue's own published table instead of guessing it from `quant_method`. That closes a real silent substitution. I checked the bundled index: QTD000584 (Sun_2018, aptamer) is listed `.all`, and the old rule would have opened the credible-set-filtered `.cc` file. Dropping `requests` removes a dead dependency. `load_dataset_index` stops with an error on a changed column layout instead of returning empty labels. `_validate_config` refusing unknown keys is a good guard. The QTD000584 test checks the URL that is actually opened, which is the right thing to test.

Scanner notes: the unpinned `pip install` warning on `SKILL.md` was already there before this PR, which only removes `requests` from that line. The `urllib` import is a HEAD request in a test that only runs when live tests are switched on, and it sends no user data. Neither is a defect.

1. `eqtl_catalogue_region_fetch.py:720` only adds the file class to the cache key when a config sets it. A QTD000584 window cached under the old rule would still be served. Versioning the key, for example with the index release, would retire those entries.
2. Some upstream labels look wrong and will show up in panel titles. At `dataset_index_r7.tsv:465` and `:470` the Schmiedel_2018 Treg memory and naive labels look swapped, and at `:665` a Nathan_2022 CD8+ dataset is labelled `CD4+ T cell`. They match the source table, so a short note in SKILL.md would be enough.

---

madaraviv:
Thanks, both made sense:

1. The cache key now names the file class that will be opened (from the config, else from the bundled table) and the table's release, so a window cached before the table existed, under the old quant_method rule, is never served again, and a future table release retires the cache the same way. Two tests: the key for a dataset with no explicit file class ends in the table's class and release, and a pre-table entry on disk is ignored, the fetch runs, and the result lands under the new name. Both fail on the previous key.
2. The labels match the catalogue's own tabix_ftp_paths.tsv (and its dataset_metadata tables) on master today, so they are the source's values rather than a derivation error. I have raised them with the catalogue (eQTL-Catalogue-resources#61, with the per-row fix); if the maintainers take it, the bundled table gets updated in a follow-up PR here. No note in the skill until then.

30 passed, 2 skipped (live).

manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Thank you for this. It fixes a real silent substitution. The skill now takes each dataset's file class from the catalogue's own table instead of guessing it from `quant_method`, so QTD000584 (aptamer) no longer opens the credible-set-filtered `.cc` file in place of the full `.all` one.

I checked the bundled table against its provenance file. It has 306 `all` rows and 452 `cc` rows. No `ge` or `microarray` dataset is listed as `cc`, and QTD000584 is the only other method listed as `all`, as the PR says. The fetch stops with an error if the table is missing, if its columns change or if a `dataset_id` is not in it. The cache key now includes the file class and the table release, so old cached windows are not served. Only public FTP byte ranges are requested, and no user data leaves the machine.

Static scanner findings: the "unpinned pip install" warning in `SKILL.md` is not new here. That line already existed, and this PR only removes `requests` from it. The `urllib` note is a HEAD request in a live test that only runs under the `live` marker. Neither is a defect.

1. `skills/locuscompare-region-render/locuscompare_region_render.py:527` still decides the `.cc` caveat from `quant_method`. A QTD000584 plot will now be captioned as credible-set-filtered while it shows the full `.all` rows. Basing the caveat on the new `file_class` field of the result would fix this.

---

manuelcorpas (approved):
Maintainer read on head 314fe48fedef16a9eb2f2c8dadd2863d6d379841. Verified against the live sources: the metadata API returns 410, eQTL-Catalogue-resources#59 confirms the retirement, the bundled r7 table re-derives byte-identically from upstream tabix_ftp_paths.tsv (758 rows, three URL families reconstruct 758/758, 306 all / 452 cc, QTD000584 the only non-ge/microarray 'all'), the file sits under the 256 KB wheel hook, and no inference fallback survives at head. All three earlier suggestions are addressed. Nothing new leaves the machine. Two follow-ups, neither blocking: `_resolve_study_id` at eqtl_catalogue_region_fetch.py:323 is now unreferenced, and locuscompare_region_render.py:527 still derives the .cc caveat from quant_method. Automated audit by ClawBio pr-audit plus a maintainer read.

PR #435, opened 2026-09-14
fix(mendelian-randomisation): no estimator, and no report, from an empty instrument set

RezaJF:
> **Stacked on #433.** This branch is based on that PR's head, so until #433 merges the diff above also shows its nine commits. Only the last commit is mine. Merge #433 first and this reduces to a single commit.

## TL;DR

#433 makes three estimators say when they do not apply below 3 instruments. It leaves `n = 0` ungated, where the same class of defect remains: `ivw([])` returns `applicable=True` with `estimate=nan, se=inf`, `steiger_test([])` returns a NaN p-value, and the CLI dies on an unhandled `ZeroDivisionError` after reporting four estimators "Running". This gates it, and pins two boundaries in #433 that no test currently holds. The demo is unchanged.

## What this fixes

`MIN_SENSITIVITY_INSTRUMENTS` gates 1 and 2. IVW is genuinely correct at 1, where it is the Wald ratio. Nothing gates 0 — which is reachable from the CLI with a well-formed file whose `instruments` array is empty, and, more realistically, from any upstream p-value threshold, LD clumping or harmonisation step that filters every SNP out.

Measured on #433 at `8326dda`:

```
ivw([])            MREstimate(estimate=nan, se=inf, applicable=True)
steiger_test([])   (False, nan, '')
CLI, empty array   "0 instruments loaded", four "Running ..." lines, then an
                   unhandled ZeroDivisionError from compute_i_squared_gx,
                   a stack trace, and no report
```

`applicable=True` with a NaN estimate is the one shape `MREstimate` exists to make impossible, and the NaN p-value would reach #433's own `allow_nan=False` backstop and convert a bad input into a crash at the final step.

Three independent causes, each fixed where it lives:

- **`ivw`** returns `not_applicable` below `MIN_IVW_INSTRUMENTS = 1`. It is public and directly callable, so it answers for itself rather than trusting the caller — the same reasoning #433 gives for guarding `phi`.
- **`compute_i_squared_gx`** tested `q <= df or df == 0`. At `n = 0`, `df` is `-1`, so `0.0 <= -1` is False and `-1 == 0` is False: both routes past the guard were open and `(q - df) / q` divided by a zero `q`. Now `df <= 0`, tested before the weighted mean so `bx_bar` never evaluates `0/0` (which also removes a `RuntimeWarning`).
- **`steiger_test`**'s `all(v is not None and v > 3 for v in n_exp + n_out)` is vacuously True over an empty list, so the no-instrument case took the branch that requires sample sizes, reached `np.mean([])`, and returned a NaN. Now predicated on a non-empty set.

And the root cause at the pipeline boundary: nothing validated that there was anything to analyse. Zero instruments is not an analysis whose estimators are unavailable — that is what `applicable=False` is for, and it still leaves a report worth writing — it is the absence of the analysis. `run_pipeline` raises `NoInstrumentsError` before creating the output directory, so a refused run leaves no artefacts to be mistaken for a completed one, and `main` surfaces it through `parser.error`: exit 2, no traceback, and a message naming the likely cause. `NoInstrumentsError` subclasses `ValueError`, so callers already catching that keep working.

**Judgement call for review:** raising rather than writing a report that says nothing applied. I took the view that a report describes an analysis, and with no instruments there is none. Happy to invert it if the maintainers prefer an artefact in every case.

## Two boundaries #433 leaves unpinned

Both currently behave correctly. Neither is held by a test, so a future edit moves them silently.

- **`n = 3` must produce an estimate** for the weighted median and mode. Every test that expects a number from those two uses 12, 15 or 30 instruments, so widening `MIN_SENSITIVITY_INSTRUMENTS` to 4 fails nothing today. It now fails two tests.
- **The mixed case**, one estimator declining while the others stand. Every end-to-end case in #433 is all-or-nothing: the demo, where all four apply, and the two-instrument run, where only IVW does. The branch of the report comparison where `comparable` and `skipped` are both non-empty is never executed. Five instruments with near-identical exposure effects reach it — Egger's slope is unidentified while the median and mode are ordinary order statistics of five well-defined ratios — and the report must name the three that ran without falling through to "the IVW estimate stands alone".

`n = 1` for the median and mode also gains a direct test; previously their guards were reachable only through the single two-instrument end-to-end run.

## Verification

- **12 new tests, 57 total**, 0 warnings under `-W error::RuntimeWarning`.
- **Seven mutations, each observed failing**: the `ivw` gate removed; `df <= 0` reverted to `df == 0`; the `steiger_test` non-empty predicate removed; the pipeline refusal removed; the threshold widened to 4 and narrowed to 2; and the report re-admitting inapplicable rows to the comparison.
- One note on method, because it is the trap #433's own comments warn about: the "threshold narrowed to 2" mutation first appeared to pass. `= 3` and `= 2` are the same byte length, so Python reused cached bytecode and the sweep never ran the mutant. With `PYTHONDONTWRITEBYTECODE=1` it fails correctly. A mutation never observed failing is not known to be observed.
- **Demo unchanged**: IVW `0.597863` (se `0.036909`), MR-Egger `0.602231` (se `0.081583`), Weighted Median `0.600088` (se `0.046948`), Weighted Mode `0.603087` (se `0.070531`) — the same four values #433 pins against TwoSampleMR `mr_egger_regression`.
- `SKILL.md` gains a Gotcha for the refusal, since a caller that catches it should report which step emptied the instrument set. 281 lines, within the 500 limit.

## Consumers

`run_pipeline` now raises on an empty instrument set where it previously crashed further in. Anything calling it in a loop over cohorts should catch `NoInstrumentsError` (or `ValueError`) and record which cohort had no usable instruments.

Credit to @madaraviv — this is a follow-through on #433, which found and fixed the substantive statistical defects in this skill.

### DIFF
--- a/skills/mendelian-randomisation/SKILL.md
+++ b/skills/mendelian-randomisation/SKILL.md
@@ -116,7 +116,7 @@ You are **Mendelian Randomisation**, a specialised ClawBio agent for causal infe
 
 | Format | Extension | Required Fields | Example |
 |--------|-----------|-----------------|---------|
-| Harmonised instruments JSON | `.json` | SNP, effect_allele, other_allele, eaf, beta_exposure, se_exposure, pval_exposure, beta_outcome, se_outcome, pval_outcome | `demo_instruments.json` |
+| Harmonised instruments JSON | `.json` | SNP, effect_allele, other_allele, eaf, beta_exposure, se_exposure, pval_exposure, beta_outcome, se_outcome, pval_outcome; optional `n_exposure` / `n_outcome` (sample sizes, needed for a Steiger p-value) | `demo_instruments.json` |
 
 ## Workflow
 
@@ -154,14 +154,15 @@ Expected output: A full MR report for 30 synthetic BMI → T2D instruments showi
 
 1. **IVW**: beta = sum(w * bx * by) / sum(w * bx²), with multiplicative random-effects variance inflation (Burgess et al., 2013)
 2. **MR-Egger**: Weighted linear regression of by on bx with intercept; slope = causal estimate, intercept = pleiotropy (Bowden et al., 2015). Reported as **not applicable**, with a stated reason, when it is undefined on the given instruments: fewer than 3 of them (it fits two parameters, so below 3 there is no residual degree of freedom), or exposure effects too close to identical for the slope to be identified. Never a number in those cases.
-3. **Weighted Median**: Median of Wald ratios weighted by inverse-variance; consistent when ≥50% weight from valid instruments (Bowden et al., 2016)
-4. **Weighted Mode**: Kernel density mode of weighted Wald ratios (Hartwig et al., 2017)
+3. **Weighted Median**: Median of Wald ratios weighted by inverse-variance; consistent when ≥50% weight from valid instruments (Bowden et al., 2016, doi:10.1002/gepi.21965; PMID 27061298)
+4. **Weighted Mode**: Kernel density mode of weighted Wald ratios, with a bandwidth proportional to the spread of the ratios (`phi` x their standard deviation) and a bootstrapped standard error (Hartwig et al., 2017)
 
 **Key thresholds**:
 - F-statistic > 10 for instrument strength (Staiger & Stock, 1997)
 - I²_GX > 0.9 for MR-Egger validity; SIMEX recommended below (Bowden et al., 2016)
 - Cochran's Q P < 0.05 indicates heterogeneity
 - Egger intercept P < 0.05 indicates directional pleiotropy
+- Steiger directionality is computed from z-statistics, so it does not depend on the units the traits are reported in; supply `n_exposure` and `n_outcome` per instrument for a p-value, without them only the direction is reported
 - MR-Egger needs >= 3 instruments and at least two distinct exposure effects; below either it is not applicable rather than imprecise
 
 ## Example Output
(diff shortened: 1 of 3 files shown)


manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Thank you, Reza. This is careful work. I checked the numerical changes against TwoSampleMR and the cited papers, and they match. That covers orienting instruments to a positive exposure effect before MR-Egger, the t reference on n-2 df, and the weighted mode's delta-method ratio SE, inverse-variance weights, Silverman bandwidth and scaled-MAD bootstrap SE. It also covers the Steiger r² from z with the independent-correlations Fisher test, and the n < 3 cut-offs. The corrected PMIDs (24114802, 27061298, 29149188, 34702754) are right. Refusing an empty instrument set before anything is written, together with `allow_nan=False`, closes a real silent-failure route. The scanner's one INFO note is the `subprocess.run` in the CLI test. It calls `sys.executable` with a list of arguments and no shell, so it is harmless.

1. The example output in `SKILL.md:182` still shows the old MR-Egger row (0.5989, SE 0.0391). Your parity test gives 0.6022, SE 0.0816. `SKILL.md:193` still says "Confirmed", but without sample sizes the code now calls the direction consistent, not confirmed.
2. The docstring at `mendelian_randomisation.py:819` only covers the no-sample-size bias when the outcome GWAS is the larger study. The opposite case is common too, for example a large BMI exposure GWAS against a smaller disease cohort, and that one leans towards "correct". Saying both in the note would push users to supply `n_exposure` and `n_outcome`.
3. This diff also contains the changes from #433. Please agree a merge order with that author, so neither of you has to rebase through the other's conflicts.

---
Approved and merged.

PR #436, opened 2026-09-14
fix(mendelian-randomisation): the Steiger note names its variance-explained conversion; the SKILL.md example output matches the report

madaraviv:
## TL;DR

The three notes from the audit of #433. The Steiger note now states its assumption: the variance explained is computed as for continuous traits on both sides, and a binary trait in log odds is not supported. The SKILL.md example output is now the report the demo writes. The MR-Egger comment about a zero exposure effect turned out to be correct, so the code is unchanged from main and a new test pins that behaviour to TwoSampleMR's own output. Demo estimates unchanged. 58 tests pass.

## What changed

**Steiger assumption, stated.** `steiger_test` converts each side's variance explained from z with `z^2 / (z^2 + n - 2)`, the continuous-trait formula (TwoSampleMR `get_r_from_bsen`), so it assumes the exposure and the outcome are both continuous. A binary trait in log odds needs a different conversion, from the case and control counts and the prevalence, which the instrument JSON does not carry, so it is not supported. The note on the Steiger row now states that assumption whenever a p-value is reported, and the docstring and SKILL.md's Steiger bullet say the same. Without sample sizes, the note no longer repeats the interpretation's "no p-value" clause, so the row reads as one sentence.

**MR-Egger at a zero exposure effect, pinned.** `mr_egger_regression` orients with `sign0` ([R/mr.R lines 591-598](https://github.com/MRCIEU/TwoSampleMR/blob/c14776b89056e88422d9dff92b517b870a1339a4/R/mr.R#L591-L598)), which counts an exposure effect of exactly zero as positive, so that instrument keeps its outcome effect. `np.where(bx < 0, -1.0, 1.0)` on main already does this. A new test pins it to the reference: that function, loaded verbatim into R 4.6.0, on the 30 demo instruments plus one at beta_exposure 0 and beta_outcome 0.3 returns slope 0.3278598234 (se 0.2842734543) and intercept 0.01158227036 (se 0.01082998364). The comment above the line now quotes `sign0` with its line reference.

**SKILL.md example output.** The example is now the report.md the demo writes, with the timestamp elided. The sensitivity table has four columns, so the Steiger row renders inside it; the MR-Egger row reads 0.6022 (se 0.0816), the values since #433; and the interpretation and disclaimer are the ones the code prints.

## Verification

- 58 passed. With the orientation changed to `np.sign`, the zero-exposure test fails.
- Demo unchanged: IVW 0.5979 (se 0.0369), MR-Egger 0.6022 (se 0.0816), weighted median 0.6001 (se 0.0469), weighted mode 0.6031 (se 0.0705). report.md and result.json are identical to the previous head apart from the timestamp.

Not in this change: the weighted median's second-order ratio variance, which @RezaJF suggested taking in its own PR.


manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Thank you for this. The Steiger change is right. TwoSampleMR `mr_steiger` does switch to `get_r_from_lor` for log odds traits, and this input has no case, control or prevalence fields. Naming the continuous-trait conversion in the note (mendelian_randomisation.py:604) is honest, and the tests pin it.

**Needs confirming**

1. MR-Egger orientation, mendelian_randomisation.py:255 and :260. As far as I know, `mr_egger_regression` in TwoSampleMR does not use base `sign()`. It uses a `sign0` helper that sets `x[x == 0] <- 1` first. So a zero exposure effect keeps its outcome effect with sign +1, which is what the old comment said. If so, `np.sign` moves away from the reference. It also changes the result: an instrument with no exposure effect but a non-zero outcome effect is pure pleiotropy, and zeroing it pulls the Egger intercept towards the null. Could you check R/mr.R in TwoSampleMR and link the line? The test at test_mendelian_randomisation.py:157 asserts the new rule, not output from the reference, so it cannot settle this. A fixture produced by a real TwoSampleMR run would.

**Non-blocking suggestions**

- SKILL.md:193 now has four cells, but the header at SKILL.md:186 has three, so the note falls outside the rendered table.
- `get_r_from_lor` applies to a binary exposure as well as a binary outcome; the note could cover both sides.

The static scan reported no findings, so there are no false positives to record.

---

manuelcorpas (changes requested):

**Verdict: changes requested.** The Steiger work is right and should land; the MR-Egger orientation change goes the wrong way and needs to come out.

**The reference does not zero a zero exposure effect.** The automated review this morning suspected a `sign0` helper; I read TwoSampleMR `R/mr.R` on master, lines 591 to 599 (the same code is at v0.5.6 lines 497 to 504, so it is not recent):

```r
sign0 <- function(x) {
  x[x == 0] <- 1
  return(sign(x))
}
to_flip <- sign0(b_exp) == -1
b_out <- b_out * sign0(b_exp)
b_exp <- abs(b_exp)
```

So the old `np.where(bx < 0, -1.0, 1.0)` was already byte-for-byte `sign0`, and the old comment ("a zero effect keeps its sign as +1 there too") was correct. The new `orient = np.sign(bx)` at `mendelian_randomisation.py:255` introduces the divergence the PR describes removing, and the new comment quotes a `sign(0)` line that does not exist in the reference. On your own example the reference behaviour is the 0.327860 slope, not 0.602060. Neither implementation drops the instrument: R keeps it at `(0, b_out)` with weight `1/se_out^2` in the weighted `lm`; the new code keeps the weight but rewrites the point to `(0, 0)`, which pulls the intercept towards zero for no reason the reference has.

Please revert the orientation to the previous expression and invert the new test at `tests/test_mendelian_randomisation.py:157-176`: `mr_egger(demo + zero)` should equal `mr_egger(demo + beta_exposure 1e-12)`, not the zeroed fit. The test structure itself is good, it observes the rule and fails on revert; it just pins the wrong rule.

**Non-blocking, from the same read.**

- The docstring attributes the `z^2/(z^2+n-2)` conversion to `mr_steiger`; in TwoSampleMR that function fills a missing r via `get_r_from_pn`, and it is `directionality_test` that messages the user to pre-compute `r.outcome` with `get_r_from_lor` for a binary outcome. The formula itself matches `add_rsq.r` 217-221 exactly. Worth tightening since the PR's purpose is comment accuracy.
- SKILL.md's Sensitivity table header (line 186) has three columns and the rebuilt Steiger row has four, so the note still renders outside the table.
- SKILL.md line 182 lists MR-Egger as 0.5989 (se 0.0391) while the PR body and the code report 0.6022 (se 0.0816). Pre-existing, but you are editing that block.

madaraviv:
You are right, thank you. `mr_egger_regression` orients with `sign0`, not base `sign` ([R/mr.R lines 591-598](https://github.com/MRCIEU/TwoSampleMR/blob/c14776b89056e88422d9dff92b517b870a1339a4/R/mr.R#L591-L598)), so the original `np.where(bx < 0, -1.0, 1.0)` and its comment were correct. My change introduced the mismatch it claimed to remove, and the test passed because it checked the code against my restatement of the rule rather than against R.

Pushed 411ed88:

- The orientation is reverted, and the comment now quotes `sign0` with the line reference.
- The test now pins TwoSampleMR's own output, as you suggested. `mr_egger_regression`, loaded verbatim into R 4.6.0, on the 30 demo instruments plus one at beta_exposure 0 and beta_outcome 0.3 returns slope 0.3278598234 (se 0.2842734543) and intercept 0.01158227036 (se 0.01082998364). The test asserts those numbers and the two p-values, and fails with `np.sign` in place.
- The Steiger docstring no longer attributes the log-odds conversion to `mr_steiger`. Since the skill does not implement that conversion, the docstring, the note and SKILL.md now state the assumption instead: the exposure and the outcome are both treated as continuous, and a binary trait in log odds, on either side, is not supported.
- SKILL.md's example output is now the report the demo writes, timestamp aside, so the sensitivity table has four columns with the Steiger row inside it, and MR-Egger reads 0.6022 (se 0.0816).

Demo estimates and result.json are unchanged; 58 tests pass. I have retitled the PR and rewritten the description to cover only what remains.

manuelcorpas (approved):

**Verdict: approved.** Every point from the 15 September review is resolved at 411ed88.

- `mendelian_randomisation.py:260` is back to `np.where(bx < 0, -1.0, 1.0)`, and the comment's citation of TwoSampleMR `R/mr.R` 591-598 at c14776b8 is exact.
- The new test at `tests/test_mendelian_randomisation.py:156-178` pins the reference output. I recomputed the weighted least squares fit independently from `example_data/demo_instruments.json` plus the (0, 0.3) instrument: slope 0.3278598234 (se 0.2842734543), intercept 0.0115822704 (se 0.0108299836), p 0.2581943402 and 0.2936821788, all six matching to the stated precision. With `np.sign` the slope moves to 0.602060 and the test fails.
- The Steiger continuous-trait assumption is stated in the docstring (:549-553), the report note (:603-606) and SKILL.md:165, and nothing attributes the log-odds conversion to `mr_steiger` any more.
- SKILL.md's example is now the demo report: four-column sensitivity table, MR-Egger 0.6022 (se 0.0816), no stale 0.5989 anywhere.
- The only non-comment code changes are the two note strings, so no estimate changes. CI ran the skill's tests on 3.11, 3.12 and 3.13 (58 passed).

Thank you for turning the correction into a reference-pinned test; that is exactly the shape that stops the next drift.

PR #438, opened 2026-09-15
bio-orchestrator: route population-genetics statistics and alignment NEXUS files to dnasp

drdaviddelorenzo:
## Summary

- Requests that name a population-genetics statistic computed by the `dnasp` skill never reached it. Shorter generic keywords won the longest-match loop: "alignment" sent "Tajima's D on this alignment" to `seq-wrangler`, "diversity" sent "nucleotide diversity per population" to `equity-scorer`, and "fold" sent "unfolded site frequency spectrum" to `struct-predictor`.
- These requests now route to `dnasp` when no other explicit intent is named, with a hint about its inputs. Keywords that only describe its input ("alignment", "diversity", "variant", "compare") and population-genetics phrases ("unfolded", "population structure", "gene flow") are ignored for that decision. An explicit intent still wins: "find papers about Tajima's D" reaches `lit-synthesizer`, and "predict this protein fold and compute Tajima's D" keeps `struct-predictor` in multi-skill detection.
- NEXUS files (`.nex`, `.nexus`, `.nxs`) route to `dnasp` only when they contain a DATA or CHARACTERS block; tree-only NEXUS files are not routed.

Existing routes are unchanged: "compute diversity metrics" still reaches `equity-scorer`, read alignment still reaches `seq-wrangler`, and all existing orchestrator tests pass. The orchestrator `SKILL.md` routing table, NEXUS note and keyword lists are updated to match.

## Skill affected

bio-orchestrator (routing to dnasp)

## Checklist

## Test output

Python 3.12.9, macOS, on current `main`:

```
python -m pytest -q skills/bio-orchestrator/tests
101 passed

python -m pytest -q clawbio/tests
397 passed
```


manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Thanks, this is a careful piece of routing work. The gating is the right shape: `dnasp` only wins when no independent intent is named, so `find papers about Tajima's D` still reaches `lit-synthesizer` and `annotate variants then compute Tajima's D` still reaches `vcf-annotator`. Stripping `unfolded` and `population structure` before keyword matching (orchestrator.py:279 onwards) is a neat fix for the `fold` and `structure` collisions, and the tests cover both directions rather than only the happy path. The NEXUS suffixes match what `dnasp.py:565` actually parses, `.nex` was not previously in `EXTENSION_MAP`, so nothing is re-routed away from another skill, and tree-only NEXUS correctly returns no route.

The static scan reported nothing on this PR and I found nothing to add to it.

**Non-blocking suggestions** (for later, not required here)

1. The hint at orchestrator.py:439 and the table row at SKILL.md:40 both offer "multi-sample VCF" as an input, but `load_alignment` in `skills/dnasp/dnasp.py:565` handles FASTA and NEXUS only, and the dnasp SKILL.md currently sends VCFs to variant annotation. A VCF fails loudly rather than returning a bad number, so this is cosmetic, but the line only becomes true once the dnasp VCF work lands.
2. `_nexus_has_alignment` (orchestrator.py:279) inspects the first 1 MB. A file with a large TAXA or TREES block ahead of the DATA block would simply not be routed. That is the safe direction, but a suffix scan would be more robust.
3. Aligned FASTA still has no extension route, so `.fas` files only reach `dnasp` via the query path.

---
Approved and merged.

PR #439, opened 2026-09-15
phylogenetics-builder: keep the alignments produced by the pipeline

drdaviddelorenzo:
## Summary

- The MSA and trimAl stages wrote `aligned.fasta` and `trimmed.fasta` inside a temporary directory that is deleted when the run ends. The alignment could not be inspected, cited or reused by another skill.
- Those files are now copied to `alignment/aligned.fasta` and `alignment/trimmed.fasta` when the corresponding stage ran. They are listed in `result.json` (`alignment_files` and `preferred_artifacts`), in the report's outputs table and in `reproducibility/checksums.sha256`, and declared as an output in `SKILL.md`.
- A suggested action points to the untrimmed alignment for population-genetics statistics, for example with the `dnasp` skill, because trimming removes alignment columns. Managed alignment files are cleared at the start of each run and whenever their stage produced nothing, so a rerun into the same output directory never reports an alignment from an earlier run. A pre-aligned, untrimmed run exports nothing.

## Skill affected

phylogenetics-builder

## Checklist

## Test output

Python 3.12.9, macOS, on current `main`. The new tests mock the external aligner, trimAl and IQ-TREE, so they run without those binaries:

```
python -m pytest -q -m "not integration" skills/phylogenetics-builder/tests
74 passed, 3 deselected
```


manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Good fix, and the underlying problem was real: the pipeline worked inside a `TemporaryDirectory`, so both the MSA and the trimAl output were discarded at the end of every run. The export records them in `result.json`, the report table and the checksum manifest, and the stale-file handling is careful; `export_alignments` (`phylogenetics_builder.py:747`) removes a managed file whose stage did not run, so a rerun into the same directory cannot report an alignment left by an earlier one. The tests check real filesystem state rather than echoing the implementation's own strings, and `test_rerun_with_failed_trimming_keeps_only_the_new_alignment` covers precisely the case that would otherwise mislead. Pointing users at the untrimmed file for population-genetics statistics is the right call; trimAl drops columns and would bias diversity and segregating-site counts.

The static scan reported no findings on this PR, and I found nothing to add against it.

1. The pre-clean at `phylogenetics_builder.py:795` runs before the input is resolved. A rerun such as `--input out/alignment/aligned.fasta --aligned --output out` deletes the input before it is opened, then aborts with "input file not found". It fails loudly and the alignment is regenerable, so no wrong tree results, but moving the pre-clean below input validation, or skipping deletion when the destination resolves to the input path, would close the trap.
2. `{"type": "alignment"}` is a new artifact type; worth checking the UI handles types it does not know.

---

---

drdaviddelorenzo:
Suggestion 1 from the audit is fixed in 92a7459, since the trap was easy to hit: the pre-clean of managed alignment files now runs after the input is validated, and it never deletes the file the run is reading (`export_alignments` takes a `protect` path). `test_rerun_on_an_exported_alignment_does_not_delete_its_own_input` reruns with `--input out/alignment/aligned.fasta --aligned --output out` and asserts the input survives with its contents intact. Suite: 75 passed, 3 integration tests deselected.

Suggestion 2 (the new `alignment` artifact type) is left for whoever owns the interface.

drdaviddelorenzo:
Suggestion 2 from the audit is raised as #440 rather than changed here: `preferred_artifacts` carries a `type` that nothing validates, and whether the vocabulary is open (with a UI fallback) or fixed and documented is a contract decision beyond this PR. `alignment` is the accurate label for these files, so it stays.

manuelcorpas (approved):

**Verdict: approved.** The change does what the description says and is wired everywhere it should be: `export_alignments()` (`phylogenetics_builder.py:747-777`) copies the aligner and trimAl outputs into `alignment/aligned.fasta` and `alignment/trimmed.fasta`, and the paths reach `result.json`, `preferred_artifacts`, the report's Outputs table, `checksums.sha256` and SKILL.md. SKILL.md, tests and code agree on the paths. `trimmed.fasta` is the file actually handed to ModelFinder and IQ-TREE, and the suggested action to use the untrimmed copy for population statistics is right. No new network or subprocess, demo path unchanged. CI ran this skill's tests on all three Pythons (75 passed).

Two things for a follow-up PR, neither blocking:

1. **`protect` stops deletion but not overwriting.** At `:769-771`, `shutil.copyfile(source, dest)` runs whenever the stage produced output, so a rerun with `--input out/alignment/trimmed.fasta` into the same `out` (the reuse pattern this PR advertises) rewrites that file with the new trimmed output. `result.json["input_file"]` and `reproducibility/commands.sh` then name a file that no longer holds the bytes that were fed in. Refusing to export over the protected path, or writing to a temp name and renaming after the run's input has been read, would close it.
2. **The trimmed fixture cannot tell the two exports apart.** The trimAl stub at `tests/test_phylogenetics_builder.py:831` strips `-` from `demo_alignment.fasta`, which contains no gap characters, so the aligned and trimmed fixtures are byte-identical and the only assertion on `trimmed.fasta` is that it exists. Writing a distinguishable stub output and asserting its content would make a swapped copy fail.

Also worth a `version` bump in SKILL.md, since the skill declares a new output and a new `result.json` field.

drdaviddelorenzo:
Follow-up opened as #446: the export now refuses a destination the run is reading (reported in contract_alerts, not just stderr), the trimAl stub derives its output from the stage's own input so a swapped copy fails, and the version goes to 0.3.0 in SKILL.md, __version__ and the catalog.

PR #441, opened 2026-09-16
fix: parse BOM-prefixed genetic exports

krudo-taco:
## Summary

Use `utf-8-sig` in the shared genetic-file opener so AncestryDNA and MyHeritage exports that start with a UTF-8 BOM keep their CSV/TSV header names. Format detection still succeeded on those files, but `csv.DictReader` saw `\ufeffrsid` and returned no records.

## Skill affected

Shared parser (`clawbio.common.parsers`), used by PharmGx, NutriGx, ancestry-risk-profiler, and other genotype consumers.

## Checklist

- [x] Existing parser contract is preserved for ordinary UTF-8 files.
- [x] Tests included and passing (`pytest` on the new BOM regressions).

## Test output

- `python3 -m pytest clawbio/common/tests/test_parser_bom.py -q --disable-warnings -o addopts=`: 3 passed.
- `python3 -m py_compile clawbio/common/parsers.py`: passed.
- `git diff --check`: passed.


manuelcorpas (approved):

**Verdict: approved.** `utf-8-sig` is a strict superset of `utf-8` for the ordinary case, works in `gzip.open` text mode, and the two nf-core samplesheet builders in this repo already read with it, so the shared opener is now consistent with them. I applied the patch to a clean clone: all four new tests fail on main and pass with the change.

One thing on our side, not yours: CI runs `pytest tests/` and two files under `clawbio/tests/`, so `clawbio/common/tests/` is never collected and the green tick here does not cover these tests. I have verified them locally instead and will fix the collection separately.

PR #442, opened 2026-09-16
fix: treat partial VCF GT calls as no-calls

krudo-taco:
## Summary

The shared VCF parser dropped only `.` alleles from a separated GT, so `0/.` and `./1` became single-base callable genotypes. Skip any slash- or pipe-separated GT that contains a missing allele. Complete haploid (`1`) and complete polyploid (`0/1/1`) calls remain supported.

## Skill affected

Shared parser (`clawbio.common.parsers`), used by PharmGx and other VCF consumers.

## Checklist

- [x] Existing complete-call contract is preserved.

## Test output

- `python3 -m pytest clawbio/common/tests/test_parser_partial_gt.py -q --disable-warnings -o addopts=`: 1 passed.
- `python3 -m py_compile clawbio/common/parsers.py`: passed.
- `git diff --check`: passed.


manuelcorpas (approved):

**Verdict: approved.** A GT of `0/.` or `./1` was becoming a one-allele callable genotype through `parse_vcf`, while `parse_vcf_matrix` in the same file already treats any `.` in the GT as missing, so this brings the two paths into agreement. The direction is the safe one: a partial no-call must never surface as a call downstream in PharmGx or the other genotype consumers. Complete haploid and polyploid calls are pinned by the same test. Applied to a clean clone: the test fails on main and passes with the change.

As on #441, `clawbio/common/tests/` is not collected by CI, so I ran it locally; that is a maintainer-side gap.

PR #443, opened 2026-09-16
fix: honor GWAS Catalog association max_hits

krudo-taco:
## Summary

`get_associations(rsid, max_hits=100)` called the GWAS Catalog association endpoint without `size`. The Catalog pages at 20 records by default, so a valid 25-association result was silently truncated. Forward `size=max_hits`. The public `max_hits` argument is unchanged.

## Skill affected

`gwas-lookup`.

## Checklist

- [x] Existing SKILL.md limit contract is now forwarded to the API.

## Test output

- `python3 -m pytest skills/gwas-lookup/tests/test_gwas_lookup.py -q --disable-warnings -o addopts=`: 17 passed (includes the new size-forwarding regression).
- `python3 -m py_compile skills/gwas-lookup/gwas_lookup_api/gwas_catalog.py`: passed.
- `git diff --check`: passed.


manuelcorpas (changes requested):

**Verdict: changes requested.** The change is harmless, but its premise does not hold for this endpoint, so as written it fixes nothing and records a wrong fact about the API.

I called the Catalog live today (2026-09-16):

- `singleNucleotidePolymorphisms/rs7903146/associations` returns 352 associations, no `page` object, and `_links` containing only `self`.
- The same URL with `?size=25`, `?size=5` and `?size=5&page=0` returns 352 every time.

So the per-SNP association sub-resource is not paged at 20 and ignores `size`; the existing `raw_assocs[:max_hits]` at `gwas_catalog.py:42` is already what enforces `max_hits`. Forwarding `size` only changes the cache key. The new test's `FakeClient` builds in a 20-record default (`captured["params"].get("size", 20)`) that the real API does not have, so it pins the belief rather than the service.

If you have an rsID where the response really is truncated, please put the URL and the observed count in the description and I will re-read; otherwise I suggest closing this one. The other four PRs in the batch read well.

manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Thank you for this fix. It is small and aimed at the right problem. Before this change, `get_associations` in `gwas_catalog.py:32` sent no `size` parameter. If the Catalog returned its default page of 20 associations, a `--max-hits 100` run would stop at 20 without any warning. Sending `params={"size": max_hits}` asks the API for the full number. The local `raw_assocs[:max_hits]` slice still enforces the limit. `BaseClient.get` already accepts `params`, and its cache key includes them (`base_client.py:62`). Entries cached under the old key will not be reused, so there is no stale-cache risk. Parsing of associations, effect sizes and traits is unchanged, so this does not affect scientific output beyond returning more rows. The static scan reported no findings, and I found nothing to dispute.

The new test checks that the endpoint and `size` are forwarded, and that `max_hits=25` returns 25 rows from a 30-row mock page. That covers the claimed behaviour.

1. `gwas_catalog.py:74` sets `total_associations` to `len(raw_assocs)`, which is the size of the returned page, not the Catalog's true total. If an rsID has more associations than `max_hits`, the report will under-count. Reading `page.totalElements` when it is present would make that field accurate.
2. The Catalog may have its own maximum page size. A short comment, or a check that warns when the returned page is smaller than the true total, would make any remaining truncation visible.

---
Approved and merged.

PR #445, opened 2026-09-16
fix: distinguish unavailable evidence from empty results

krudo-taco:
## Summary

Omics target evidence mapping treated UniProt/PubMed/ClinicalTrials.gov request failures as empty successful searches (`no_result` / `[]` / "No literature hits found."). Distinguish `unavailable` from `no_result` in the evidence object and report wording so a timeout or outage is not presented as "nothing found".

Open Targets already returned `unavailable` on request failure. This change aligns UniProt, PubMed, and ClinicalTrials.gov with that contract.

## Skill affected

`omics-target-evidence-mapper`.

## Checklist

- [x] Existing SKILL.md research-triage contract is preserved.

## Test output

- `python3 -m pytest skills/omics-target-evidence-mapper/tests/test_omics_unavailable.py skills/omics-target-evidence-mapper/tests/test_omics_target_evidence_mapper.py -q --disable-warnings -o addopts=`: 6 passed.
- `python3 -m py_compile skills/omics-target-evidence-mapper/omics_target_evidence_mapper.py`: passed.
- `git diff --check`: passed.


manuelcorpas (changes requested):

**Verdict: changes requested, one small fix and then this merges.** Separating a request failure from an empty search is right, and it brings UniProt, PubMed and ClinicalTrials.gov onto the contract Open Targets already used in this file. The evidence.json shape for `literature` and `trials` changes from a list to `{status, items}`; SKILL.md documents no schema, nothing outside the skill reads the file, and `build_report` keeps a shim for the old shape, so that is fine.

**Blocking:** CodeQL is red on this head. `tests/test_omics_unavailable.py:94` (`if "clinicaltrials.gov" in url`) trips `py/incomplete-url-substring-sanitization`. In a fake router it is not a security issue, but it leaves a failing check and a new open alert on main. `url.startswith("https://clinicaltrials.gov")` clears it, and the same file's unused `import pytest` at line 6 is the second alert.

**Non-blocking:** `limitations` in `build_evidence` is still the static four-line list, so a report whose literature was unavailable says "not assessed" in the body while the limitations section stays silent about it. Appending one line per unavailable source would close that.

manuelcorpas (changes requested):
### ClawBio automated audit

**Verdict: REJECT**  

Thank you, this fixes a real problem. On main, a UniProt, PubMed or ClinicalTrials.gov request that fails is reported as "No literature hits found", which is wrong. Checking `is None` separately from an empty payload is right. The `unavailable` wording in the report is clear, and the mixed success test shows that sources which did answer are still rendered.

**Blocking**

1. `omics_target_evidence_mapper.py:355-356` in `main()` still does `len(evidence["literature"])` and `len(evidence["trials"])`. Both are now dicts with two keys, so `metadata.json` will always record `literature: 2` and `trials: 2`. That happens whether there were five hits, none, or the source was unavailable. The report is correct but the metadata is wrong. Please count `len(evidence["literature"]["items"])` and write the `status` too, so an unavailable source is not recorded as a count. A test that runs `main()` against a temporary output directory would catch this.

**Non-blocking suggestions**, for later

2. `omics_target_evidence_mapper.py:138`: if esearch returns PMIDs but esummary has no entry for any of them, the result is `no_result`. PubMed did find papers in that case, so `unavailable` or a separate status would be more accurate.
3. `evidence.json` now stores `literature` and `trials` as objects, not lists. No other code in the repo reads them, but a short note in the SKILL.md or changelog would help anyone parsing the old format.

---
Changes requested.

manuelcorpas:
Correction to the automated audit above: the metadata count bug is at `skills/omics-target-evidence-mapper/omics_target_evidence_mapper.py:373-374`, not `:355-356`. The finding itself stands at head 13a35b33: `len(evidence["literature"])` and `len(evidence["trials"])` now count the two keys of `{status, items}`, so `metadata.json` records 2 for each whatever the result. Counting `["items"]` and writing `status` alongside fixes it. The CodeQL alerts at `tests/test_omics_unavailable.py:94` and `:6` from the earlier review are also still open.

krudo-taco:
Thanks for the review, and for the line-number correction (373-374).

Pushed on this branch:

- `metadata.json` now counts `literature`/`trials` `items` and records each source `status`, so an unavailable source is no longer stored as `2`
- `main()` is covered against a temporary output directory
- CodeQL: the test router uses `url.startswith("https://...")`; the unused `pytest` import is gone
- unavailable sources are named in `limitations`

Left for a later change: esearch-with-PMIDs / empty-esummary status, and a SKILL.md schema note.

manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Good fix, and the right one for this skill. Previously a dropped PubMed or ClinicalTrials.gov call rendered as "No literature hits found", which reads as evidence of absence; now it does not. The back-compat shim in `omics_target_evidence_mapper.py:269-282` keeps old list-shaped `evidence.json` working, and `_source_items`/`_source_status` at `omics_target_evidence_mapper.py:363-376` stop `metadata.json` counts from silently absorbing a failure. The tests assert that the two wordings actually differ rather than just that a key exists, which is the thing worth asserting.

I re-checked the static scan for this PR. It reported nothing, and I found nothing it missed.

**Non-blocking suggestions** (for later, not needed here)

1. `omics_target_evidence_mapper.py:174-181`: the Open Targets enrich step is still conflated. If the second GraphQL POST fails, `enrich_data` is `None`, `enrich` becomes `{}`, and the block is reported as `no_result`. That is the same trap you have just closed for the other three sources.
2. Same file: GraphQL errors arrive as HTTP 200 with an `errors` key, so `safe_request_json` will not return `None` for them. Checking for `errors` before trusting the payload would close that gap.
3. `omics_target_evidence_mapper.py:137`: if esearch returns PMIDs but every esummary record is empty, status is `no_result` despite real hits; a `partial` status would be more honest. `SKILL.md` and `examples/expected_output.md` also still describe the old list shape.

Thanks for tightening this up.

---

---

krudo-taco:
The blocking items from the 16 and 17 September reviews are fixed on this branch, and the remaining CodeQL alerts are fixed on 09be697.

- `metadata.json` counts `literature`/`trials` items and records each source status (46275258). An unavailable source is no longer stored as `2`.
- `main()` is covered against a temporary output directory.
- The unused `pytest` import is gone.
- Unavailable sources are named in `limitations`.
- The test router no longer uses `in` or `startswith` on URL strings. It compares `urlsplit(url).hostname` exactly, and the PubMed failure stays limited to `/entrez/eutils/esearch.fcgi` so esummary is not treated as a miss. That is the pattern CodeQL's `py/incomplete-url-substring-sanitization` query does not flag. Local run of the skill tests: 8 passed.

Please re-review this head. The existing changes-requested reviews are on 13a35b33, and there is no approving review on the current head, so the pull request stays blocked. The Open Targets enrich failure, GraphQL `errors`, and empty-esummary cases called out on 18 September are left for a later change, as requested.

@manuelcorpas @camlloyd

manuelcorpas (approved):
**Verdict: SAFE, with one follow-up** (audited at 09be6978)

This is the right fix and the failure paths were traced individually. Every `safe_request_json` returning `None` now maps to an explicit `unavailable` rather than to absence: UniProt, PubMed esearch and esummary, and ClinicalTrials.gov. The report prints "Literature unavailable; not assessed." before the item check, so the reassuring "No literature hits found." branch is unreachable for a failure, and `metadata.json` gains a `status` block so `counts.literature: 0` is no longer ambiguous.

The tests are the good kind. They fail with the fix reverted, they pin the *empty* side too so loosening "unavailable" to fire on empty goes red, and `test_report_wording_differs_for_unavailable_and_empty_sources` asserts the reassuring wording is **absent** from the failed render. `_fake_json_mixed_success` fails only esearch on the eutils host and explains in a comment why esummary must not be caught by the same match.

**WARNING, one instance of the same defect remains, in Open Targets.** That source makes two calls and only the first distinguishes failure from emptiness:

```python
enrich = (enrich_data or {}).get("data", {})
...
"status": "ok" if target_hit or disease_hit else "no_result",
```

If step 1 resolves IDs and the step-2 enrichment request fails, `enrich_data` is `None`, both hits are `None`, and the function returns `no_result`. An unreachable enrichment step is then reported as no disease association found, and the report prints `Status: no_result`. That is the exact collapse this PR exists to prevent, one layer down. A `if enrich_data is None: return {"status": "unavailable", ...}` before the fallback closes it.

Two smaller points:

- The Target Summary and Disease Association sections print the raw status word rather than the plain-English "not assessed" wording used for literature and trials.
- The list-shaped compatibility shims treat an empty list as `no_result`, which reintroduces the ambiguity for any caller still passing the old shape. They exist only to keep one old fixture passing; updating that fixture and deleting the shims would be cleaner.

Pre-existing and out of scope, but worth a limitations line: in `fetch_pubmed_hits`, PMIDs missing from the esummary `result` are skipped, so if all are skipped the function returns `no_result` although esearch found hits.

No conflict with #456; different skill, different files.

PR #446, opened 2026-09-16
phylogenetics-builder: protect a rerun's input from being overwritten

drdaviddelorenzo:
## Summary

Follow-up to #439, addressing the three points from its review.

- **Protecting the input covers overwriting, not just deletion.** `protect` stopped the pre-clean from deleting a rerun's own input, but the export itself still copied over it, so a rerun with `--input out/alignment/trimmed.fasta --output out` rewrote that file with the new trimAl output while `result.json` and `reproducibility/commands.sh` still named it. `export_alignments` now refuses to export over the protected path, says so on stderr, in `chat_summary_lines` and in `contract_alerts`, and leaves that key out of `alignment_files`. The message names the remedy: rerun with a different `--output`. `SKILL.md` documents the exception, so the output contract no longer promises an export that can be refused.
- **The trimAl test stub is distinguishable from the aligned export.** It stripped gap characters from a demo alignment that contains none, so `aligned.fasta` and `trimmed.fasta` were byte-identical and a swapped copy would have passed. The stub now derives its output from the stage's own input, asserting that trimAl was handed the alignment rather than the original file, and cuts every sequence to 40 columns. The result is both derived from the input and distinguishable from it, and still usable as the input of a rerun.
- **Version bump.** 0.2.0 to 0.3.0 in `SKILL.md`, in `__version__` and in the regenerated `skills/catalog.json`, since the skill declares a new output and a new `result.json` field.

New test: `test_rerun_on_an_exported_alignment_does_not_overwrite_its_own_input` reruns with the exported trimmed alignment as input, then asserts the file's bytes are unchanged, that `trimmed` is absent from `alignment_files`, and that `result.json` and the replay command still name a file whose contents are what was read.

## Skill affected

phylogenetics-builder

## Checklist

## Test output

Python 3.12.9, macOS, on current `main`:

```
python -m pytest -q -m "not integration" skills/phylogenetics-builder/tests
76 passed, 3 deselected

python -m pytest -q clawbio/tests
397 passed
```


manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Thanks for this, it is a careful fix. `export_alignments` already refused to delete a rerun's own input, and now it also refuses to overwrite it (`phylogenetics_builder.py:773`). That matters because `result.json` and the replay command in `reproducibility/commands.sh` both name that path. If the file were overwritten, they would point at bytes the run never read. The refusal goes to stderr, the chat summary and `contract_alerts`, so nothing fails silently, and the tree is still built from the trimmed file in the temporary directory. The comparison uses resolved paths on both sides, so a relative `--input` is caught too.

The test changes make the suite stronger. The trimAl stub now checks that it is handed the alignment and writes output that differs from its input, so a test can no longer pass by exporting the aligned file twice. The new rerun test checks the file's bytes, `alignment_files`, `input_file`, the replay command and the alert.

This change does not touch any tree-inference, model-selection or support-value logic, so I found no scientific risk. The static scan reported no findings for this PR.

### Non-blocking suggestions
These are for later.

1. In `test_rerun_on_an_exported_alignment_does_not_overwrite_its_own_input`, `.replace("ACGT", "TGCA")` does nothing if no sequence contains `ACGT` in its first 40 columns. The byte-equality check would then pass even without the guard. The other two assertions still cover the refusal, but writing fixed, clearly different content would make that check meaningful on its own.
2. A minor version bump (0.2.0 to 0.3.0) is more than a bug fix like this usually needs; a patch bump would be enough, unless you are also counting the new `contract_alerts` entry.

---
Approved and merged.

PR #448, opened 2026-09-16
fix(dnasp): name the CLI flag each metadata input maps to

drdaviddelorenzo:
## The defect

The DnaSP skill's metadata names its inputs logically — `window_size`, `step_size`, `pop_file`, `genetic_code` — and never said which command-line flag each maps to. Those are the Python API parameter names, so the metadata was internally coherent; it just left an agent reading only the registered contract to guess the CLI spelling.

Two independent models guessed the same wrong way. Given a task needing a sliding window, both wrote `--window-size 100 --step-size 25`. Neither flag exists; the real ones are `--window` and `--step`, and argparse rejects the command outright.

## Evidence

Measured over 300 runs, two operator models (Claude Code 2.1.273 and codex-cli 0.154.0) against ten population-genetics tasks phrased as a biologist would ask them, each reply scored by executing it:

| What the operator was given | Correct commands |
|---|---|
| `--help` output alone | 60/60 |
| registered metadata alone, before this change | 51/60 |
| registered metadata alone, after this change | 59/60 |

The sliding-window task failed in **every** run given the metadata alone, and in no run given the help text. Adding `cli_flag` takes it from 0/6 to 6/6. That result holds under both of the scorings reported in the benchmark write-up, so it does not depend on any contested scoring choice.

## The change

- `cli_flag` on all ten declared inputs in `SKILL.md`.
- Version 0.5.3 across the front matter, `__version__`, `docs/index.md` and the regenerated `skills/catalog.json`. The version-history entry is marked not result-changing.
- `tests/test_skill_contract.py`, seven tests.

## Behaviour

Unchanged. Nothing parses this front matter to drive execution — `dnasp.py` never reads its own `SKILL.md` — so no statistic moves. The full suite passes at 864, up from 857 by the seven new tests. The validation record against DnaSP 6.12.03 is untouched.

One caveat worth flagging: `__version__` is stamped into `result.json` and the reproducibility bundle, so byte-for-byte artefact comparison changes at 0.5.3 even though every computed value is identical.

## Tests

The mapping is asserted against an explicit expected table rather than merely checked for existence, so swapping two valid flags fails the suite — verified by mutation. The suite also pins `--window-size` and `--step-size` as invalid while `--window` and `--step` exist, refuses two inputs claiming the same flag, and requires the metadata version, the front-matter version and `__version__` to agree, which a partial bump had already broken in one copy.

## Noted, not fixed here

`scripts/generate_catalog.py` reads `inputs`, `outputs`, `author`, `domain`, `demo_data` and `endpoints` from every skill's front matter and the emitted catalog entry uses none of them. Nothing in the repository reads those keys off the catalog, so this changes no behaviour, but it does mean catalog-only consumers never see the new mappings. Happy to raise it separately if you'd like it either wired up or removed.


manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Thank you for this. It fixes a real problem for agents. Each input in `skills/dnasp/SKILL.md` now names its flag in `cli_flag`, which stops the made-up `--window-size`. Only metadata, docs and version strings change. The numbers the tool produces stay the same, and the 0.5.3 changelog entry says so accurately.

The new `tests/test_skill_contract.py` is well built. It checks the exact mapping in `EXPECTED_FLAGS`, not just that each flag exists, so swapping `window_size` and `step_size` would fail. `test_declared_flags_exist_in_the_parser` also ties the metadata to `build_parser()`, so the two cannot drift apart. The static scan reported no findings, and nothing here contradicts that.

### Non-blocking suggestions

These are for later.

1. `tests/test_skill_contract.py:258` only checks the argument type for `window`, `step` and `genetic_code`. The explicit table already guards against swaps, so this is fine. The docstring could say that the table does the real work.
2. `test_version_is_consistent_across_the_skill` (`tests/test_skill_contract.py:220`) compares the SKILL.md metadata with `dnasp.__version__`. It does not check `skills/catalog.json` or `docs/index.md`, and both were bumped by hand in this PR. Adding them would catch a partial bump in the files where it is most likely.
3. `_parser_flags` reads argparse's private `_actions`. That works on the Python versions in CI, but a comment noting it would help whoever maintains it next.

---
Approved and merged.

PR #451, opened 2026-09-17
refactor: migrate six skills to shared reproducibility helpers

camlloyd:
## Summary

- Migrate six skills from hand-rolled reproducibility bundles to the shared `clawbio.common.reproducibility` helpers (`write_portable_commands_sh`, `write_environment_yml`, `write_checksums`), as required by CLAUDE.md step 6. Each bundle now carries a portable replay script (`$CLAWBIO_ROOT` / `$OUTPUT_DIR`), a named conda recipe and SHA-256 digests, instead of a `commands.sh` with machine-specific absolute paths baked in.
- Fix a reproducibility bug that replaying a bundle exposed: `sample-qc-triage`, `crispr-screen-triage` and `marker-dominance-mapper` iterated set-derived column names, so `result.json` key order shifted with `PYTHONHASHSEED` and two identical runs produced different checksums. Sorted the loops and added a byte-stability test to each.
- Update the `## Output Structure` tree in all six SKILL.md files to list the files now written.

This is the first tranche; 30 skills still hand-roll their bundles.

## Skill affected

`sample-qc-triage`, `crispr-screen-triage`, `marker-dominance-mapper`, `analyze-fasta`, `fastreer`, `ukb-navigator` — one commit per skill.

Behaviour notes: `analyze-fasta` keeps its existing `run.json`, and `fastreer` keeps its `environment.txt` Java/pip snapshot alongside the new conda recipe. `ukb-navigator` passes the search query through `shlex.quote`, so a multi-word query survives replay as a single shell argument.

## Checklist

Every change was developed red/green: the bundle-completeness and byte-stability tests were written first and observed failing before the implementation landed.

## Test output

```
$ pytest -q skills/sample-qc-triage/tests skills/crispr-screen-triage/tests \
    skills/marker-dominance-mapper/tests skills/analyze-fasta/tests \
    skills/fastreer/tests skills/ukb-navigator/tests
......................................................................   [100%]
70 passed in 28.50s
```

Full repo suite (`pytest -q tests`) reports `4 failed, 302 passed`. All four failures are in `tests/benchmark/test_finemapping_benchmark.py` and are pre-existing — they need the `jax`/`sushie` extras (`uv sync --extra fine-mapping`) and fail identically on `main` with this branch stashed.

Bundles were also verified by replay: running a generated `reproducibility/commands.sh` from a copied output directory re-runs the skill and reproduces its outputs.


manuelcorpas (changes requested):
### Maintainer review (pr-audit plus domain read)

**Verdict: changes requested, and most of the blocker is on our side.** Moving these six skills onto the shared helpers is the right direction, and one commit per skill made it easy to check. What holds up: no flag that `commands.sh` used to record is lost; the analysis code is untouched apart from `sorted()` in the loaders, which fixes the hash-seed key-order bug you describe; `run.json` and `environment.txt` are unchanged; the bundle is written after all outputs in every skill; `shlex.quote` on the ukb-navigator query is correct; and no test assertion was removed.

**Blocking**

1. **The new recipes cannot replay the run they sit next to.** Before this PR `sample_qc_triage.py` imported only the standard library. It now imports `clawbio.common.reproducibility`, and importing anything under `clawbio.common` runs `clawbio/common/__init__.py`, which loads `scrna_io` (numpy, pandas) and `report`, which loads `audit` (opentelemetry). The recipe declares `pip_deps=[]` and `python_version="3.10"` (`sample_qc_triage.py:157-162`), while `pyproject.toml` requires `>=3.11`. So `conda env create -f reproducibility/environment.yml` followed by `bash reproducibility/commands.sh` stops at import, in all six skills.

   This is mostly not your bug: the helper defaults to 3.10 and sits behind a heavy package `__init__`, and `skills/gwas-prs/repro_bundle.py:40` already documents the trap. Two ways out: declare the real dependency set and Python 3.11 in each recipe, as gwas-prs does, or wait for a helper change that imports without the package `__init__`. Say which you prefer.

2. **SKILL.md now contradicts the code.** "Python 3.10+ standard library only" at `sample-qc-triage/SKILL.md:164`, `crispr-screen-triage/SKILL.md:165` and `marker-dominance-mapper/SKILL.md:163`; "pure stdlib + Biopython" at `analyze-fasta/SKILL.md:255`; and `analyze-fasta/SKILL.md:46` and `:143` and `sample-qc-triage/SKILL.md:88` still list only `commands.sh` (and `run.json`) among the reproducibility outputs.

**Non-blocking**

3. Replay calls whichever `python` is on PATH (`clawbio/common/reproducibility.py:98`). The triage skills previously recorded `sys.executable`, so on a machine without `python` pointing at Python 3 the replay now fails outright.
4. `commands.sh` is written with `write_text_lf`, not atomically (`reproducibility.py:125`), and every skill points `--output` back at `"$OUTPUT_DIR"`. A replay regenerates the script bash is still reading, and bash reads the tail of the new file when the regenerated `CLAWBIO_ROOT` default has a different length. `write_text_lf_atomic` already exists in `clawbio/common/textio.py:43`. Helper side again.
5. The fastreer demo replay records `--input "$OUTPUT_DIR/demo_samples.vcf"` without `--demo` (`fastreer.py:476-488`), so replay takes the strict path and fails when fastreeR or Java is missing, which is exactly when the demo ran. `--window-variants`, and ukb-navigator's `--n-results` and `--db-path`, are not recorded either. These predate the PR.
6. Every bundle test runs `--demo`. A non-demo `--input`, a path with spaces, and an actual replay are untested, and `test_ukb_navigator.py:122` also passes on main's unquoted `--query "{query}"` form.

manuelcorpas:
Thanks, 04717ae to 56f0d04 covers all four helper defects (lazy `clawbio/common/__init__.py`, `"${PYTHON:-python3}"`, atomic `commands.sh`, 3.11 default), and the `test_no_skill_pins_a_python_below_requires_python` check is a good addition. I had started the same helper fix separately and have dropped it. Checked at 56f0d047:

1. **Why `test (3.11)` is red: an obsolete test, not your change.** `skills/gwas-prs/tests/test_repro_bundle.py:210` (`test_numpy_and_pandas_really_are_eager_imports`) asserts the eager import your `__init__` removes, and its docstring says what to do: "If this ever stops being true, drop them from REPLAY_PIP_DEPENDENCIES." Run on their own, as the CI step does, the gwas-prs tests give 1 failed, 166 passed, and this is the failure. Importing `gwas_prs` on this head now loads `opentelemetry` (through `clawbio.common.report`) and `requests`, but not numpy or pandas. So: drop `numpy>=1.24` and `pandas>=2.0` from `REPLAY_PIP_DEPENDENCIES` (`skills/gwas-prs/repro_bundle.py:52`), update the set in `test_environment_declares_every_eager_runtime_dependency` (`:195`), and turn `:210` into a check that they are not loaded. It passes in one combined local run only because other tests have already imported both libraries, so run the gwas-prs tests on their own to check.

2. **A new `commands.sh` is now 0711, not 0755.** `write_text_lf_atomic` writes through `mkstemp`, which creates files 0600, and keeps the old mode only when the file already exists. On this head a fresh bundle from both `write_portable_commands_sh` and `write_commands_sh` (`clawbio/common/reproducibility.py:125` and `:240`) is 0711, so anyone else reading a shared bundle gets permission denied; main gives 0755. `path.touch(exist_ok=True)` before the atomic write fixes it, because the new file then gets the normal umask mode, which the atomic write preserves. A test that fails without it: write a plain file with `write_text` in the same `tmp_path`, and assert the new `commands.sh` mode equals that mode `| 0o111`.

3. **Minor, optional.** `import clawbio.common as c; c.parsers` worked on main because every submodule was imported eagerly; `__getattr__` (`clawbio/common/__init__.py:43`) now raises AttributeError for submodule names. Nothing in the repo uses that form, so only worth it if you want to keep it: fall back to `import_module(f"{__name__}.{name}")`.

Beyond the gwas-prs test, the affected suites on this head (`clawbio/common/tests`, `clawbio/tests`, the six skills, gwas-prs, nutrigx, wgs-prs, cell-detection, proteomics-de) show no failure that main does not already have.

Separately, #454 records the tightened protection and your PyPI Owner role and is waiting on your approval.

camlloyd:
Thanks — all six points are addressed. Taking the blockers in order.

**1. Recipes could not replay.** I went with the helper change rather than declaring the heavy dependency set per skill, because it fixes all six at once and the ~30 skills still to migrate. `clawbio/common/__init__.py` now resolves its exports lazily (PEP 562), so `from clawbio.common.reproducibility import ...` no longer runs the package `__init__` and no longer pulls in numpy, pandas or opentelemetry. Verified against a bare venv with nothing installed:

```
before: ModuleNotFoundError: No module named 'opentelemetry'
after:  Sample QC Triage wrote .../report.md
```

`clawbio/common/tests/test_import_weight.py` asserts it in a subprocess, so a future import added to any of those submodules fails the suite rather than the replay.

`python_version` is 3.11 in all six. While fixing that I folded in the default itself: `write_environment_yml` defaulted to 3.10, so every caller that omitted the argument had the same defect, and `cell-detection` and `proteomics-de` passed `"3.10"` explicitly. Two tests now read `requires-python` from `pyproject.toml` instead of hardcoding — one for the signature default, one scanning every `python_version=` literal under `skills/`. The second is what found those two skills. I kept it in this PR rather than stacking it, since it is the same defect these recipes had.

**2. SKILL.md contradicted the code.** Fixed at every line you listed: the three "Python 3.10+ standard library only" lines, analyze-fasta's "pure stdlib + Biopython", and the output trees and workflow steps at `analyze-fasta:46`, `:143` and `sample-qc-triage:88` that named only `commands.sh` and `run.json`.

**3 and 4.** Both correct, both one line. `commands.sh` now goes through the existing `write_text_lf_atomic` — its docstring describes this exact self-overwrite case — and the recipe invokes `"${PYTHON:-python3}"` rather than a bare `python`.

**5.** fastreer records `--demo` for demo runs instead of the generated input, plus `--window-variants`. ukb-navigator's `--n-results` and `--db-path` are recorded when they differ from the default and omitted when they do not, so the recipe shows what the run chose.

**6.** `sample-qc-triage` now copies a completed run, executes its own `commands.sh` from a directory with a space in the path, and compares checksums. On the ukb-navigator quoting test you were right, and more so than I first thought: my initial strengthening still passed against the unquoted form, because `blood pressure; echo pwned` survives inside double quotes too. It now shlex-parses the command line and asserts the query is a single argument, with a query containing an embedded quote. Verified by reverting `shlex.quote`, which fails the test.

Two notes, neither in scope here:

- `clawbio/common/tests/test_report.py::test_generate_report_header_requires_skill_version` fails on a clean `main` — `skill_version` has a default of `""`, so no `TypeError` is raised. CI's allowlist does not run `clawbio/common/tests/`, so it has been red unnoticed.
- `skills/xena-tcga-gene-query/SKILL.md:46` has the same stale "commands.sh and run.json" description as the ones fixed here.

manuelcorpas:
The red `test (3.13)` here is not yours: all 26 cell-detection failures are `No module named 'tifffile'`, which is in no dependency list, so that skill's tests cannot pass in CI. They only ran because the one-line `python_version` change counts as touching the skill. #455 adds a light `cell-detection` extra and also makes CI run `clawbio/tests/` and `clawbio/common/tests/` in full, so your new helper tests will actually run after merge. Rebase on main once it lands, or leave it and I will re-run this.

manuelcorpas:
Checked at 1acfae92. The helper work is right and all fourteen checks pass. Two things and then I will approve.

**1. This PR still regresses the mode of a new `commands.sh`, from 0755 to 0711.** Ran your code rather than read it:

```
new commands.sh mode: 0o711 | a plain write plus +x would be: 0o755
```

`write_text_lf_atomic` writes through `mkstemp`, which creates 0600, and preserves an existing file's mode only when the file already exists, so the `| 0o111` at `clawbio/common/reproducibility.py:126` and `:241` lands on 0600 rather than on the umask mode. Anyone other than the owner can execute a shared bundle's script but not read it. `path.touch(exist_ok=True)` immediately before the atomic write fixes both call sites, because the new file then gets the usual umask mode and the atomic write keeps it. A test that fails without it: write a plain file with `write_text` in the same `tmp_path`, then assert the new `commands.sh` mode equals that file's mode `| 0o111`. I checked that it goes red if the `touch` is removed.

**2. 7558d00 and 1acfae92 dropped work that #455 has now unblocked.** `test_no_skill_pins_a_python_below_requires_python` was the check that found `cell-detection` and `proteomics-de` pinning 3.10, and both are back at `python_version="3.10"`, below `requires-python`. That revert was only needed because cell-detection's tests could not pass in CI: `tifffile` and `scikit-image` were in no dependency list. #455 merged a light `cell-detection` extra, and on that PR the skill ran 76 passed, 4 skipped in CI. So please restore the test and both pins here. Your branch is 17 commits behind main, so merge main in first.

Nothing else outstanding from either round: the recipes import in a bare interpreter, the docs match the code, the fastreer demo replays as `--demo`, ukb-navigator records its non-default flags, and the shlex-parsed quoting test is stronger than what I asked for.

manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Nicely scoped refactor, and the tests earn their place. Resolving exports lazily in `clawbio/common/__init__.py` and then measuring the claim in a fresh interpreter (`test_import_weight.py`) is the right order of operations, and dropping numpy and pandas at `repro_bundle.py:51` follows from that measurement rather than from hope. Two details I checked and liked: `fastreer.py:468` records `--demo` instead of the generated input, so a replay does not take the strict path that the fallback existed for; and `test_reproducibility_command_quotes_multiword_query` parses with `shlex.split` instead of asserting on quote characters, so it would actually catch a regression. The `sorted()` calls over set iteration are what make the byte stability tests mean anything.

No clinical, variant or population genetics logic is touched, and nothing new leaves the machine. The static scan reported nothing on this PR and I found nothing to contradict it.

### Non-blocking suggestions

For later, not for this PR.

1. `crispr_screen_triage.py:119`, `marker_dominance_mapper.py:138` and `sample_qc_triage.py:146` pass `input_path` straight from argparse. A relative `--input data/x.csv` is not relative to `_PROJECT_ROOT`, so it renders as a bare relative path and the replay resolves it against whatever directory bash happens to start in. `analyze_fasta.py` and `fastreer.py` both call `.resolve()` first; worth matching.
2. `skills/gwas-prs/SKILL.md:182` still says environment.yml declares numpy and pandas because importing `clawbio.common` loads them eagerly. That sentence is no longer true after this change.

---
Approved and merged.

PR #456, opened 2026-09-18
fix: treat GWAS Catalog null associations as empty

krudo-taco:
## Summary

GWAS Catalog sometimes serialises `_embedded.associations` as JSON `null`. `.get("associations", [])` still returns `None` when the key is present, and slicing it raised `TypeError`.

Coerce `null` / non-list payloads to `[]` so a missing association page is an empty result, not a crash.

## Validation

- [x] `python3 -m pytest skills/gwas-lookup/tests/test_gwas_lookup.py`: 18 passed, including the new null-associations regression.
- [x] `python3 -m py_compile` on `gwas_catalog.py`.
- `git diff --check`: passed.

## Checklist

- [x] User-facing problem and affected call path described above.
- [x] Public association objects are unchanged; only the null/non-list guard is new.
- [x] Added a FakeClient regression for `associations: null`.
- [x] No credentials, cached API responses, or generated artifacts committed.


manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Thanks, this is a tidy defensive fix. Swapping `data.get("_embedded", {})` for `data.get("_embedded") or {}` in `gwas_catalog.py:41` is the right shape: a JSON `null` under an existing key defeats the default argument, and the old code would then have raised `TypeError` on the `raw_assocs[:max_hits]` slice at line 45, outside the `try`, so the whole lookup crashed rather than reporting no hits. Zero associations is already a normal, handled outcome downstream (`gwas_lookup_core/normalise.py:52`), so returning `status: ok` with an empty list is the correct semantics here, not a swallowed error. The new test drives `_make_client` through monkeypatch and asserts both `associations` and `total_associations`, which is the right level.

**Non-blocking suggestions** (for later, not required here)

1. `gwas_catalog.py:13-14`: the `isinstance` guard also turns any non-list payload, say a dict, into an empty result, so a genuinely malformed response becomes indistinguishable from "this rsID has no associations". Consider returning the `status: error` branch for the non-list case and keeping the silent empty only for `None` and absent.

2. `tests/test_gwas_lookup.py:29`: the docstring asserts the Catalog "occasionally serialises associations as null". The test proves the code path, not that behaviour of the source. A link to an observed response would make the claim checkable; the handling is harmless either way.

3. Pre-existing, not yours: `total_associations` at `gwas_catalog.py:79` counts the returned page, not the Catalog total, so it under-reports past `max_hits`.

---

---

manuelcorpas (commented):
**Verdict: NEEDS REVIEW** (minor; audited at 9012b43b)

The change is safe in the way that matters. It sits after the failure boundary, so it cannot turn a network or API failure into "no associations": `raise_for_status()` sends timeouts, 5xx and 404 to the `except`, which returns `status: "error"`, and downstream only ingests `status == "ok"`. The new lines govern a 200 with a JSON body only.

The test is not tautological and fails red with the fix reverted, since `None[:max_hits]` raises `TypeError`.

Two asks before merge:

**1. The null claim needs evidence.** The description says the Catalog sometimes serialises `_embedded.associations` as JSON `null`, with no rsID, no captured payload and no documentation cited. It did not reproduce against the live API: known rsIDs returned populated lists, unknown ones returned 404, and Spring Data REST omits `_embedded` for an empty page rather than nulling the inner key. It may well be real for some rsID or some release, but a PR body's claim about a data source is a claim, not evidence. Please name the rsID or attach the body.

**2. The type guard trades a crash for a reassuring empty, and is untested.**

```python
if not isinstance(raw_assocs, list):
    raw_assocs = []
```

A malformed 200 (an API contract change serialising `associations` as a dict or string) now renders as `status: "ok"`, `total_associations: 0`, which the report prints as "No GWAS associations found." under an OK badge. Before this PR it raised an uncaught `TypeError` and crashed the caller. For schema drift the crash is the safer failure. Prefer returning `status: "error"` with the unexpected type in the message for the non-list, non-`None` case, keeping only `None` as empty, and add a test for that branch.

Unrelated observation while checking: the live endpoint ignored `size`, returning the full association set, so the comment about the API applying `max_hits` does not hold. The slice still caps output, so behaviour is correct; only the comment overstates.

krudo-taco:
Thanks for the audit. Addressed in 5630a68 and df9f676, and the description is rewritten.

1. I can't produce evidence for `null` and have withdrawn the claim. Probed today: rs7903146 returns 200 with a list of 352, `?size=5` still returns 352, and rs999999999999 returns 404. `associations: null` stays empty only because you asked to keep None as empty. `_embedded: null` is now an error, since nothing supports treating it as empty.
2. Only a missing `_embedded` or a missing or null `associations` key is empty now. A non-object body, a non-object `_embedded` or a non-list `associations` returns `status: "error"` with `Unexpected response format: <field> is <type>`. The `or` fallbacks are gone too; they let `{}` and `""` through as empty. Tests cover 8 malformed and 3 empty shapes. The red runs and a main-vs-PR table are in the description.
3. On `size` you are right, and so was the maintainer review on #443, which I did not answer before it merged. df9f676 reverts the `size` forwarding and replaces the #443 test, whose fake honoured `size` and so never exercised the slice. The new test fails with `[:max_hits]` removed.

One factual note: `gwas_lookup.py:219-228` catches per-API exceptions, so on main a null payload did not crash the CLI. It showed gwas_catalog as `WARNING: error` with `'NoneType' object is not subscriptable`. Only direct callers of `get_associations()` saw the exception.

manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Thanks, this is a clean fix in the right direction. Distinguishing "no associations" from "payload we do not understand" is exactly right for this skill: a malformed response previously became an empty OK result, which reads downstream as a SNP with no GWAS hits. The shape checks in `gwas_catalog.py:36-46` are narrow and ordered sensibly, treating a missing `_embedded` or a null `associations` as empty while flagging anything else, and `normalise.py:52` already gates on `status == "ok"`, so the new error dict degrades safely. The parametrised tests cover both branches properly.

One point worth confirming, not blocking. The removal of `params={"size": max_hits}` at `gwas_catalog.py:29` rests on the claim in the comment that this endpoint is unpaged. I believe that is correct: it is a Spring Data REST property resource, so it returns the full collection with no `page` block, unlike `/associations` or the `search/findByRsId` endpoints, and `skills/clinical-trial-finder/gwas_bridge.py:57` already calls it without a size. But the new test at `test_gwas_lookup.py:335` asserts this against a fake client, so it cannot actually prove upstream behaviour, and the same was true of the test it replaces.

1. Since a paged response would still be a list and pass the new checks, consider treating a `page` key in the payload as schema drift too. That converts the one residual failure mode, silently returning the first 20 of 60 associations with `total_associations` understated, into the visible WARNING you have just built.
2. A short note in `SKILL.md` recording why no size parameter is sent would stop this being re-added.

---

**Domain review: gwas-lookup**

The schema validation half of this is the right shape. Separating "no associations" from "a payload shape we do not recognise" is exactly the distinction that matters, and routing drift to `status: "error"` really does surface as `WARNING: error` in the Data Sources table (`gwas_lookup_core/report.py:181`), so the stated intent holds. Treating a missing `_embedded` and a null `associations` as an empty OK result is correct, and the parametrised cases cover the shapes well.

One question blocks me, on the other half.

`gwas_catalog.py:42` drops `params={"size": max_hits}`, reversing #423/#443, on the claim that this endpoint is not paged. I cannot settle that from the repository, and the test cannot either: `FakeClient` at `tests/test_gwas_lookup.py:335` returns all 30 rows by construction, so it passes whether or not the live API pages. If the endpoint does page (Spring Data REST pages collection resources at 20 by default), a `max_hits=100` lookup would quietly return 20 associations and report `total_associations: 20`, which reads as a complete answer for the SNP. Could you either attach a live response for a SNP with more than 20 associations showing no `page` block, or keep `size` forwarded? It is harmlessly ignored if you are right, and it prevents truncation if not.

**Non-blocking suggestions**

- The rewritten test no longer asserts anything about request parameters (`tests/test_gwas_lookup.py:346`), so nothing pins the call shape any more. Worth asserting `captured["params"]` explicitly, whichever way this lands.

---

PR #458, opened 2026-09-18
fix(repro): give a freshly written commands.sh mode 0755, not 0711

camlloyd:
## Summary

- A freshly written `reproducibility/commands.sh` landed on mode **0711**, not 0755: `write_text_lf_atomic` writes through `mkstemp` (0600) and only restored a mode when the file already existed, so the blanket `| 0o111` in `reproducibility.py` produced a script that group and other can execute but not read. Raised at the end of #451 and still live on `main`.
- Fixed at the two shared points every caller routes through, rather than per skill: `write_text_lf_atomic` now gives a **new** file the umask-derived mode a plain write would have given it (0644 under the usual `0o022`), and still preserves an existing file's mode; `reproducibility.py` mirrors the read bits (`mode | ((mode & 0o444) >> 2)`) instead of ORing `0o111`, so 0644 → 0755 while a 0600 file written under a restrictive umask becomes 0700 rather than leaking group/other execute.
- Five new tests, written red first: two umask cases in `test_textio.py`, and `TestCommandsShMode` covering both `write_commands_sh` and `write_portable_commands_sh` plus a guard that rewriting keeps an existing script's mode.

## Skill affected

None directly — `clawbio/common/{textio,reproducibility}.py`, so every skill already migrated to the shared helpers (`analyze-fasta`, `crispr-screen-triage`, `fastreer`, `marker-dominance-mapper`, `sample-qc-triage`, `ukb-navigator`) now ships a 0755 `commands.sh`. Verified on a real run:

```
$ python skills/sample-qc-triage/sample_qc_triage.py --demo --output "$TMP"
sample-qc-triage commands.sh mode: 0o755   # was 0o711
```

## Checklist

- [x] SKILL.md is present and complete (YAML frontmatter + methodology) — no SKILL.md changes; library-only fix, no output contract changes
- [x] Demo data included for reviewers to test — existing `--demo` paths; no new fixtures needed

## Test output

New and existing library tests:

```
$ pytest -q clawbio/common/tests/test_textio.py clawbio/common/tests/test_reproducibility.py
62 passed in 0.22s
```

Full library suite plus every skill on the shared helpers, to confirm no bundle regressed:

```
$ pytest -q clawbio/ skills/sample-qc-triage skills/analyze-fasta skills/fastreer \
    skills/crispr-screen-triage skills/marker-dominance-mapper skills/ukb-navigator
608 passed in 12.28s
```

The two new `test_textio.py` cases fail on `main` with `0o600 != 0o644`, and all three `TestCommandsShMode` cases fail with `0o711 != 0o755`.


manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Good fix, and the diagnosis is right: ORing `0o111` onto mkstemp's 0600 gives 0711, an execute bit nobody can pair with a read. Mirroring the read bits in `_make_executable` (clawbio/common/reproducibility.py:9) is the correct shape; 0644 becomes 0755, 0600 stays owner only, and a restrictive umask is never widened. The derivation at clawbio/common/textio.py:155 gives a new file exactly the mode a plain `open()` would have, so an atomically written bundle now matches one written through `write_text_lf`. The tests assert the real stat mode rather than an implementation constant and restore the umask in `finally`, which is the right way to test this.

No clinical or scientific logic is touched and no skill directories change, so there is no pathway here to a wrong result. The static scan reported nothing and I found nothing to add to it.

1. `_umask()` at clawbio/common/textio.py:130 sets 0o022 for an instant, so a concurrent writer in the same process would create its file under that looser mask instead of the user's. You already flag this in the docstring; caching the value once at import would close it if it ever bites. Incidentally the `ponytail:` marker on textio.py:134 appears nowhere else in the repo, so `Note:` would read more clearly.
2. `_make_executable` adds nothing when a file has no read bits (a umask of 0o444 yields 0o222), leaving commands.sh non-executable. Exotic, and `bash commands.sh` still works, but forcing `S_IXUSR` when `S_IWUSR` is set would make the guarantee unconditional.

---
Approved and merged.

PR #459, opened 2026-09-18
fix(rnaseq-de): write the repro bundle with the shared helpers

camlloyd:
## Summary

- Replaces `rnaseq-de`'s hand-rolled `reproducibility/` bundle with `clawbio.common.reproducibility` (`write_commands_sh`, `write_environment_yml`, `write_checksums`), as the contributing workflow requires; the local `_sha256` helper goes with it.
- Fixes a dependency the hardcoded bundle had already lost: `scipy` is imported at module scope for the t-distribution but was missing from `environment.yml`, so an environment rebuilt from the bundle could not run the skill.
- `environment.yml` now pins `python=3.11` and installs `pydeseq2` under `pip:`; `commands.sh` gains a shebang; checksum labels stay anchored to the output dir.

## Skill affected

`rnaseq-de` (no change to inputs, outputs, statistics, or the `## Output Structure` tree — the three bundle files are the same three, written by the shared helpers)

## Checklist

## Test output

Three tests added, written first and run red before the swap — `commands.sh` had no shebang, and `scipy missing from environment.yml`:

```
$ pytest skills/rnaseq-de/tests/ -q
........s....ssss...                                                     [100%]
15 passed, 5 skipped in 4.47s
```

The checksums test passes both before and after by design: it is the regression guard that label anchoring (`tables/de_results.csv`, `figures/volcano.png`) survived the swap.

Reviewers can see the corrected bundle with:

```
python skills/rnaseq-de/rnaseq_de.py --demo --output /tmp/rnaseq_de_demo
cat /tmp/rnaseq_de_demo/reproducibility/environment.yml
```

### Not addressed here

- `ruff` reports E402 on this file (imports after the `sys.path.insert`): 5 before this change, 6 after. The new import follows the existing `clawbio.common.report` import and shares its cause; fixing that pattern is a separate change.
- `tests/benchmark/test_finemapping_benchmark.py` fails on `main` with `ModuleNotFoundError` for sushie (needs `uv sync --extra fine-mapping`). Unrelated and pre-existing.


manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Good change. The repro bundle now goes through the shared helpers, so commands.sh gets a shebang and LF endings, and the checksum labels come out identical to the hand written ones (`tables/...`, `figures/...`). Dropping the local `_sha256` is safe; `hashlib` had no other use in `rnaseq_de.py`. The scipy point in the test is a real gap rather than vocabulary written to fit the code: `scipy.stats` is imported at module scope in `rnaseq_de.py:13` and the old environment.yml never listed it, so the bundle could not rebuild a working environment.

I checked ordering too: `write_repro_files` is called at `rnaseq_de.py:654`, after the tables and figures are written, so the globs are populated. Input paths outside the output dir fall back to the bare filename under `anchor`, which matches the previous labels. The static scan reported nothing on this PR and I found nothing to add.

**Non-blocking suggestions** (for later, nothing here blocks):

- `rnaseq_de.py:654` passes `backend`, not `backend_used`, so an `--backend auto` run records `auto` in commands.sh. Replayed on a host without pydeseq2, that reproduces via the Welch t path instead, with no signal in the bundle. Recording `backend_used` would pin it.
- `write_environment_yml` defaults to `python=3.10` exactly, where the old file said `python>=3.10`. Sensible for reproducibility, just a behaviour change worth knowing; pass `python_version` if you want a different pin.
- `write_checksums` skips missing files silently. Harmless here, since every path is either just written or globbed, but worth remembering if the list ever takes user supplied paths.

---
Approved and merged.

PR #460, opened 2026-09-18
fix(audit): write the provenance a skill_run span already accepts

camlloyd:
## Summary

- `skill_run` accepts `input_checksum`, `input_file` and `output_dir`, and its docstring warns they are written to `~/.clawbio/audit.jsonl` — but the body only set `gen_ai.agent.id` and `gen_ai.agent.version`, so all three were accepted and silently discarded.
- Each is now set as a span attribute when non-empty, so a record can be reconciled against what is on disk. Empty values are omitted, keeping the historical record shape for callers that pass nothing.
- The PII warning in the docstring becomes load-bearing rather than aspirational: callers must scrub paths and sample identifiers, because they now actually reach the log.

## Skill affected

None directly — `clawbio/common/audit.py`, shared by every caller of `skill_run`.

Found while wiring `pubmed-summariser` into the audit module (follow-up branch): a live demo run produced a `skill_run` span with no trace of where its report was written, despite `output_dir` being passed.

## Checklist

- [x] SKILL.md is present and complete (YAML frontmatter + methodology) — n/a, shared library change
- [x] Demo data included for reviewers to test — n/a, covered by unit tests

## Test output

Two tests added, written first. The first ran red with `KeyError: 'input_checksum'`; the second guards the no-argument record shape and passed before and after.

```
$ pytest clawbio/common/tests/test_audit.py -q
................                                                         [100%]
16 passed in 0.31s
```

Wider run, `clawbio/` and `bot/`: `537 passed`. `ruff check clawbio/common/audit.py`: clean.

## Reviewer note

This changes what lands in an append-only log for every `skill_run` caller. Records written before this commit keep their existing shape; new records gain the three keys only when a caller passes them.

Recording the paths is the intended direction: the goal is data lineage, so a run must say what it read and what it produced. This PR is the enabling step — it makes `skill_run` capable of carrying provenance. It is not lineage on its own, because almost nothing passes those arguments yet.


manuelcorpas (approved):
### ClawBio automated audit

**Verdict: SAFE**  

Thanks, this closes a real gap. `skill_run` has accepted `input_checksum`, `input_file` and `output_dir` since it was written, and the docstring at `clawbio/common/audit.py:96` already told callers those values land in `~/.clawbio/audit.jsonl`, but nothing ever set them on the span, so the provenance was silently dropped. The fix is the right shape: attributes go on the span and reach the record through the existing `record.update(dict(span.attributes))` path in `_JsonlExporter.export`, so nothing else had to change.

The tests are honest ones. `test_skill_run_records_provenance_attributes` reads the JSONL the exporter actually wrote rather than asserting against a mock, and `test_skill_run_omits_provenance_keys_when_not_given` pins the old record shape so existing consumers do not break. The `clawbio.*` namespace is the correct call given the OTel GenAI conventions have no equivalent key.

No clinical logic, no data-source vocabulary and no network egress here, so nothing in the classification or pharmacogenomics path is affected. The static scan reported nothing and I found nothing it missed.

1. No production caller passes `input_file` or `output_dir` yet, only the tests, so behaviour is unchanged today. Once a skill does start passing them, the PII warning at `audit.py:96` goes live and the log is set `chflags uappend` on macOS, which makes redaction awkward. Storing a hash of the path rather than the path may age better.
2. `tool_call` coerces attribute values with `str()` at `audit.py:147`. Doing the same here would stop a `Path` argument being dropped by OTel as an invalid attribute type.

---
Approved and merged.

PR #461, opened 2026-09-18
feat(audit): optional OTLP export

camlloyd:
## Summary

- Audit spans went only to `~/.clawbio/audit.jsonl`. This adds optional OTLP export so they can be read in a tracing backend, opt-in via `CLAWBIO_OTLP_ENDPOINT`, with the exporter as an optional extra (`clawbio[otlp]`) imported lazily — the base install and the default local-only behaviour are unchanged.
- Makes the exported spans legible to OpenInference viewers: span kinds (`AGENT` / `TOOL`), a named project, `tool.name`, and `input.value` / `output.value` aliased from attributes the spans already carried.
- Fixes three ways a failing tool call sent an input path to the collector while `OPENINFERENCE_HIDE_INPUTS` was set.

## Skill affected

None — `clawbio/common/audit.py`, shared by every caller of `skill_run`.

## Checklist

- [x] SKILL.md is present and complete (YAML frontmatter + methodology) — n/a, shared library change
- [x] Demo data included for reviewers to test — n/a, covered by unit tests

## Test output

```
$ pytest clawbio/common/tests/ -q
147 passed in 0.77s
```

Every behavioural claim below was also checked against a Phoenix instance running locally, by reading the spans back out of its API rather than trusting the code.

To try it, any OTLP-capable backend will do (Phoenix listens on 6006):

```bash
pip install 'clawbio[otlp]'
CLAWBIO_OTLP_ENDPOINT=http://localhost:6006 python -c "
from clawbio.common.audit import skill_run, tool_call
with skill_run('pharmgx', '0.2.0', input_file='demo_patient.txt', output_dir='reports/demo'):
    with tool_call('cpic_lookup', cmd=['echo', 'CYP2D6']):
        pass"
```

## What the commits do

| commit | |
|---|---|
| `d6c70ec0` … `82cad1b3` | the OTLP export itself, its lock entry, and the docstring warning that export sends span attributes off the machine |
| `1e84cbb3` | `openinference.span.kind` + `tool.name`. Without the kind, Phoenix renders every span as `unknown` with no agent or tool grouping |
| `f0f06bee` | `openinference.project.name`, so traces stop landing in Phoenix's shared `default` project |
| `138f93a5` | `input.value` / `output.value`, which fill the Input/Output panels and which evaluators require |
| `3c4c9daa` | honours `OPENINFERENCE_HIDE_INPUTS` / `HIDE_OUTPUTS`, redacting with the spec's `__REDACTED__` marker so the record survives |
| `5d36a40d` | drops `clawbio.input.file`, `clawbio.output.dir` and `gen_ai.tool.call.arguments` — each was a second copy of a value already on the span, and the copy defeated the hide flags |
| `df1a8904` | the leak fix, below |
| `e562381f` | `CLAWBIO_AUDIT_LOG`, so a caller's test can point the log somewhere other than `$HOME` |
| `6062e94e` | a tool span with no subprocess reports its keyword arguments as `input.value` |

## Reviewer note

The gate is deliberately `CLAWBIO_OTLP_ENDPOINT`, not the standard `OTEL_EXPORTER_OTLP_ENDPOINT`. Spans can carry output paths and sample identifiers, so an org-wide or CI-wide setting of the standard variable would silently ship them off the machine, against Safety Rule 1. A ClawBio-specific name makes enabling export a deliberate act.

The leak in `df1a8904` is the part most worth a close read. With `OPENINFERENCE_HIDE_INPUTS=true`, a failing tool call still sent the input path three ways: captured `stderr` quotes the arguments a tool was given, `CalledProcessError`'s text embeds the whole argv, and the OTel SDK's own exception handling recorded both an exception event and a status description carrying that text verbatim. The first two are now redacted by either flag, and the spans opt out of `record_exception` / `set_status_on_exception` so the SDK stops overwriting what was redacted. It was found by running a real `bcftools` against a path that does not exist — a synthetic command that always succeeds never reaches that code.

One limit worth stating plainly: these flags cover the values the spans carry, not a caller that puts an identifier in a skill name or an attrs key. They are not a substitute for scrubbing, nor for leaving the endpoint unset.

Not included, deliberately: no loopback/remote endpoint check, no header or path passthrough, and no allowlist filtering at the export boundary. Each is worth adding when a specific backend or deployment calls for it rather than up front.


manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Thanks, this is careful work. Refusing to read `OTEL_EXPORTER_OTLP_ENDPOINT` and using a private variable instead (audit.py:109) is the right instinct, and the reasoning is recorded next to it. The `provider.shutdown()` in the `finally` (audit.py:135) is a real fix; a short run would otherwise exit before the first batch tick. The test exercises both branches honestly: the deferred import at audit.py:113 means the monkeypatch actually takes effect.

One thing to settle before this lands.

**The export carries the fields this module itself labels as PII.** `tool_call` children share the provider, so `gen_ai.tool.call.arguments` (audit.py:162), the truncated `stderr` (audit.py:171) and arbitrary `**attrs` such as `rsid=` all reach the collector alongside the root span. In practice that is input VCF paths, output directories and sample IDs. The docstring at audit.py:104 still says the destination is `~/.clawbio/audit.jsonl`, so an operator who sets the variable has been told the wrong thing. Please either allowlist attributes on the OTLP processor, or update both PII warnings to name network export.

**Non-blocking suggestions**

- A dead endpoint will stall every skill run at shutdown until the batch timeout expires. A short explicit timeout would stop an unreachable collector looking like a hung analysis.
- Worth noting in the extra's comment that this is plain HTTP, so it wants a localhost collector or a TLS URL.

No scientific or clinical logic is touched, and the static scan flagged nothing I disagree with.

---

camlloyd:
Thanks — all three addressed in the latest commit.

**Blocking: docstrings named the wrong destination.** Took the "update both PII warnings" option rather than an attribute allowlist. Both docstrings now say that when `CLAWBIO_OTLP_ENDPOINT` is set the same values also go over the network to the collector, and `tool_call`'s warning names `stderr` explicitly alongside `cmd` and `**attrs`, since the children share the provider.

An allowlist on the OTLP processor is the stronger fix, but it would mean maintaining a second, quieter notion of what an attribute is for, and the values it would filter are ones callers are already required to scrub before they reach the JSONL. Happy to build it if you'd rather have the belt as well as the braces.

**Timeout:** `OTLPSpanExporter(..., timeout=5)`, with a comment on why.

**Plain HTTP:** noted in the extra's comment in `pyproject.toml` — localhost collector or an `https://` URL.

```
$ pytest clawbio/common/tests/ -q
132 passed in 0.53s
```

camlloyd:
Correction to my last comment: I'd written the warnings as though export always sends spans off the machine. It doesn't — pointing `CLAWBIO_OTLP_ENDPOINT` at a local Phoenix or a localhost collector keeps everything on this machine, which is the expected setup.

Reworded: the docstrings now say the attributes go to the configured collector, and that a local one keeps them here while a remote endpoint does not. Same for the comment on the env var choice — the risk with `OTEL_EXPORTER_OTLP_ENDPOINT` is that an org-wide value points at a collector that may not be local, not that export is off-host by nature.

132 passed.

camlloyd:
Please **squash and merge** this one when it goes in. The history is an iterative trail — a change, its verification against a running Phoenix, then a simplification — which was useful while the work was in progress but reads better on `main` as a single commit.

manuelcorpas (commented):
**Verdict: NEEDS REVIEW** — one doc change blocks (audited at 6062e94e)

The design is sound and the leak fix is real. Export is strictly opt-in: there is no default endpoint, `if endpoint:` guards the only exporter construction, the exporter package is not even imported when unset, and the `TracerProvider` is per-`skill_run` and never global, so auto-instrumentation and `OTEL_EXPORTER_OTLP_ENDPOINT` cannot attach to it. With the hide flags set, `input.value` on both span types, the `error` attribute, the status description and `stderr` are all redacted, and `record_exception=False` stops the SDK re-attaching the raw exception. The test reads the `ReadableSpan` objects the exporter is handed, which is the right layer.

**Blocking: `docs/data-handling.md` still says this cannot happen.**

> The `clawbio` package (CLI, runner, shared helpers) makes no network calls. There is no telemetry, no update check, no analytics. The audit layer in `clawbio/common/audit.py` writes a local JSONL file and nothing else.

That sentence is false the moment `CLAWBIO_OTLP_ENDPOINT` is set, and `SECURITY.md` puts exactly this in scope. The docstrings were fixed after the automated audit commented; the institutional page was not. Note that `tests/test_data_handling_doc.py` scans only `skills/`, so it structurally cannot see a new egress path in `clawbio/common/` — the gate that should have caught this cannot reach it. Please add a paragraph to "The short version" naming the variable, what a span carries (skill name, version, paths, argv, stderr excerpt, error text, checksum), the `OPENINFERENCE_HIDE_*` flags and their limit, and that plain HTTP is used.

**WARNING, one unredacted path.** When `cmd` is given, every `**attrs` kwarg is written raw regardless of either hide flag:

```python
for k, v in attrs.items():
    span.set_attribute(k, str(v))
```

A caller writing `tool_call("vep", cmd=[...], input_vcf=path)` ships the path to the collector with both flags set. No skill does this today, so it is latent rather than live, but it should be closed or named in the doc.

**Rebase needed.** The conflict is not in `audit.py`; it is `pyproject.toml` and `uv.lock`, both moved by the dependabot commit on 21 September. Regenerate the lockfile rather than hand-merging, since `lockfile-check` gates.

Two INFO items: `f"{endpoint.rstrip('/')}/v1/traces"` means anyone following OTel habit and setting the full URL gets `/v1/traces/v1/traces`; and the exporter still reads `OTEL_EXPORTER_OTLP_HEADERS` from the environment, so "no header passthrough" is not quite accurate. The version pinning is right: the exporter pins `opentelemetry-sdk~=1.44.0`, which is the alignment the past `_ON_EMIT_RECURSION_COUNT_KEY` failures needed.

camlloyd:
Rebased on `main` and all four points addressed.

**Blocking — `docs/data-handling.md`.** "The short version" now has a second bullet for the export: the variable name, that there is no default endpoint and the exporter package is not imported when it is unset, what a span carries (skill name and version, input path, output dir, checksum, tool/phase names, command line, keyword arguments, exit code, 500-char `stderr`, error text), the `OPENINFERENCE_HIDE_*` flags and their limit (values only — span names, tool names and attribute keys always go), and that the transport is plain OTLP/HTTP so it wants a localhost collector or an `https://` URL. The first bullet no longer claims audit.py writes a local file "and nothing else".

You are right that `tests/test_data_handling_doc.py` structurally cannot see this: it walks `skills/` only. Rather than widen the scanner over the whole package, there is now one test in that file pinning this specific path — if `audit.py` still constructs an `OTLPSpanExporter`, the page must name `CLAWBIO_OTLP_ENDPOINT`. It fails against the page as it stood before this commit.

**WARNING — unredacted `**attrs` beside `cmd`.** Closed rather than documented. Those kwargs are arguments to the command whose `input.value` is already masked, so `OPENINFERENCE_HIDE_INPUTS` now covers them too:

```python
for k, v in attrs.items():
    span.set_attribute(k, _hide(str(v), "OPENINFERENCE_HIDE_INPUTS"))
```

The keys are still written as given, so the docstring keeps the instruction to scrub. Test: `test_hide_inputs_redacts_attrs_passed_alongside_cmd`, which fails on the previous code.

**Rebase.** Done on `main` including the dependabot commit of 21 September. `uv.lock` was regenerated from `main`'s copy with `uv lock`, not hand-merged; `uv lock --check` is clean.

**INFO — `/v1/traces/v1/traces`.** Fixed: the path is appended only when it is not already there, so both `http://localhost:4318` and `http://localhost:4318/v1/traces` work. Test covers both spellings.

**INFO — `OTEL_EXPORTER_OTLP_HEADERS`.** Correct, and I should not have said otherwise: the exporter reads it from the environment like any other OTLP exporter. It is the exporter's own behaviour and I have left it, since suppressing it would take a header argument that overrides what an operator deliberately set. The endpoint remains the one thing not taken from the `OTEL_*` namespace.

```
$ uv run pytest clawbio/common/tests/ tests/test_data_handling_doc.py -q
155 passed in 2.37s
```

Still asking for **squash and merge**.

manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Good change, and the care is in the right places.

Choosing `CLAWBIO_OTLP_ENDPOINT` over honouring `OTEL_EXPORTER_OTLP_ENDPOINT` is the correct call, and the comment at audit.py says why: an org-wide OTel variable would otherwise have started shipping VCF paths to whatever collector the org runs. The OpenInference vocabulary checks out against the spec rather than against the implementation; `openinference.span.kind` AGENT/TOOL, `tool.name`, `input.value`/`output.value` and the `__REDACTED__` marker are strings Phoenix really consumes. `test_failure_does_not_leak_through_otel_exception_recording` is the test I would have asked for: reading the spans handed to the exporter, not the JSONL, is what catches the status description. Amending the "makes no network calls" line in docs/data-handling.md and pinning it with `test_core_egress_paths_are_documented` stops that claim drifting.

I checked the renamed keys. Nothing outside audit.py reads `clawbio.input.file`, `clawbio.output.dir` or `gen_ai.tool.call.arguments`, and no skill calls `tool_call`, so the reshape breaks no consumer.

- `clawbio/common/report.py:27` defaults `log_path` to the constant at `report.py:13`, so it always passes a path and `_resolve_log` never sees `CLAWBIO_AUDIT_LOG`. Set the variable and spans move while `write_audit_log` records stay in `~/.clawbio/audit.jsonl`; the trail splits. Passing `None` there would fix it.
- With the endpoint set but the `otlp` extra not installed, the `from opentelemetry.exporter...` import in `skill_run` (audit.py:163) raises a bare `ModuleNotFoundError` and takes the run down. Loud, so no wrong answer, but a message naming `clawbio[otlp]` would save someone a puzzle.

---

---

manuelcorpas:
### ClawBio automated audit

**Verdict: NEEDS_REVIEW**  

Read the whole diff. No skill changed; this is core infrastructure (`clawbio/common/audit.py`), and the privacy reasoning in it is the strongest part.

What is right: the export is genuinely opt-in, with no default endpoint and the exporter package not imported at all until `CLAWBIO_OTLP_ENDPOINT` is set (audit.py:158). Declining to honour `OTEL_EXPORTER_OTLP_ENDPOINT` (audit.py:160) is the correct call; an org-wide setting would otherwise route VCF paths off the machine. `record_exception=False, set_status_on_exception=False` (audit.py:179) closes a real leak, and its test reads the spans handed to the exporter rather than only the JSONL, so it proves what it claims. tests/test_data_handling_doc.py:570 pins the doc to the code, which is the right shape of guard.

No clinical or scientific logic is touched: no ACMG, ClinVar, gnomAD or CPIC vocabulary, no scores, no citations. The renames drop `clawbio.input.file` and `gen_ai.tool.call.arguments`; nothing in the tree reads those keys back, so no consumer breaks.

1. `clawbio/common/report.py:27` passes an explicit `log_path`, so `CLAWBIO_AUDIT_LOG` never redirects `write_audit_log`. The variable is only half honoured today.
2. Enabling egress is silent. One line on stderr when the endpoint is honoured (audit.py:158) would help, since an env var can come from a wrapper or container the user did not write.
3. docs/data-handling.md reads as though the hide flags cover every value listed; `clawbio.input.checksum` (audit.py:191) is never redacted. The docstring at audit.py:148 states this correctly.

---

---

PR #465, opened 2026-09-20
refactor(repro): write reproducibility bundles via shared helpers (tranche 2)

camlloyd:
## Summary

- Second tranche of the reproducibility-bundle migration: four API-bridge skills stop hand-rolling `reproducibility/` and write it with `clawbio.common.reproducibility` instead.
- Two of them (`galaxy-bridge`, `flow-bio`) were shipping an `environment.yml` that was not a conda environment at all — a dump of `galaxy_url` / `tool_id` / date, so `conda env create -f` could never recreate the run. `flow-bio` wrote no checksums on any path, and its demo recipe hard-coded `--output /tmp/flow_demo` regardless of where the run actually wrote.
- Every `commands.sh` is now portable (`CLAWBIO_ROOT` / `OUTPUT_DIR` anchored, `set -euo pipefail`, mode 0755) and records the run's own flags rather than raw `sys.argv` or a string that only resolves from one directory. User-supplied values go through `shlex.quote`.

## Skill affected

`labstep`, `protocols-io`, `galaxy-bridge`, `flow-bio`

Per-skill detail:

| Skill | Before | After |
|---|---|---|
| `labstep` | raw `sys.argv` recipe with no anchoring, no checksums, `python>=3.10` against a repo floor of 3.11 | portable recipe carrying the run's flags, conda env, checksums over `report.md` + `result.json` |
| `protocols-io` | `commands.sh` had no shebang and baked in the absolute output path | portable recipe; local `_sha256` deleted in favour of the shared hasher |
| `galaxy-bridge` | fake `environment.yml`; `python galaxy_bridge.py ...` only resolves from inside the skill dir; demo wrote a lone `commands.sh` | real conda env (bioblend), portable recipe, demo gets the same bundle as a live run |
| `flow-bio` | fake `environment.yml`, no checksums anywhere, 15 call sites each passing a command string | `write_reproducibility` takes the run's flags; all 15 call sites get the full bundle |

SKILL.md output trees were updated where they described the old files (`galaxy-bridge`, `flow-bio`) or omitted the bundle entirely (`labstep`).

Follows the same pattern as #451 (tranche 1: sample-qc-triage, crispr-screen-triage, marker-dominance-mapper, analyze-fasta, fastreer, ukb-navigator).

## Checklist

## Test output

Written test-first (red → green) per skill; the assertions pin the bundle contract: portable `commands.sh`, a named conda environment, and checksum labels that resolve inside the output directory.

```
$ pytest -q skills/labstep/tests skills/protocols-io/tests skills/galaxy-bridge/tests skills/flow-bio/tests
185 passed in 0.69s

$ pytest -q tests/ clawbio/
1 failed, 58 passed in 3.14s
```

The single failure is `tests/benchmark/test_finemapping_benchmark.py::TestMethodRunners::test_susie_returns_pips`, which fails identically on `main` (missing `sushie` extra) and is untouched here.

Manual replay check on the labstep demo bundle:

```
$ python skills/labstep/labstep.py --demo --output /tmp/labstep
$ bash /tmp/labstep/reproducibility/commands.sh     # re-runs, exit 0
$ cd /tmp/labstep && shasum -a 256 -c reproducibility/checksums.sha256
report.md: OK
result.json: OK
```

## Note for the reviewer

`skills/galaxy-bridge/demo/reproducibility/` is generated demo output that is checked into the repo, and `--demo` with no `--output` writes into it. I reverted what my test run touched and left it otherwise alone, so it is now stale relative to this code. It looks like it should be deleted and gitignored — happy to do that in a separate PR.


manuelcorpas (approved):
**Verdict: SAFE** (audited at 9488ee9d)

The refactor is behaviour-preserving where it matters. The hash algorithm is unchanged: the deleted per-skill `_sha256` is the same chunked `hashlib.sha256` as the shared helper, and the digest format is identical. No skill loses a bundle file, and two gain one they lacked: labstep and flow-bio now write `checksums.sha256`, and the ordering change in labstep (`result.json` before the repro bundle) is what makes that checksum meaningful.

Two files are written differently and both are improvements. `commands.sh` gains a shebang, `set -euo pipefail` and `$CLAWBIO_ROOT`/`$OUTPUT_DIR` in place of absolute paths, with quoting via `shlex.quote`. galaxy-bridge's checksum labels become relative to the output directory, so `sha256sum -c` now resolves from there. The fake `environment.yml` in galaxy-bridge (which carried `galaxy_url:`/`tool_id:` keys) becomes a real conda env with `bioblend`.

Tests assert content rather than existence: shebang, `CLAWBIO_ROOT`, repo-relative script path, absence of `str(tmp_path)`, env name, pip dep, 64-hex digests, and label containment.

Three follow-ups, none blocking:

1. `galaxy_bridge.py` computes `"repo_root" if input_path.is_relative_to(_PROJECT_ROOT) else "auto"` on an unresolved `Path(args.input)`. A relative `--input reads.fq` renders verbatim into a script that claims portability and does not `cd`, so replay from another directory fails. `Path(args.input).resolve()` fixes it. Pre-existing, but the new bundle advertises portability.
2. `skills/galaxy-bridge/demo/reproducibility/commands.sh` is checked in and now stale, and `--demo` with no `--output` defaults to `DEMO_DIR`, so a demo run dirties the tree. Your offered follow-up is the right call; there is no `reproducibility` rule in `.gitignore`.
3. `"--run", tool_id` is the one user-supplied value not passed through `shlex.quote`. Galaxy tool ids are safe in bash, so this is tidiness.

Worth knowing: the run timestamp no longer appears anywhere in the labstep and flow-bio bundles, and protocols-io now hard-imports `clawbio.common`, so it is no longer runnable as a standalone folder. Both match the tranche 1 precedent.

